PR #47, opened 2022-04-22
EASI-1988/plan list by collaborator

StevenWadeOddball:
# EASI-1988

## Changes and Description

- Plan Collaborators used to get model plans
- Initializer of model becomes the model Lead Collaborator
- Unable to delete the last model lead from the plan collaborator.
- Add basic unit test for collaborator.

## How to test this change
1. `run scripts/dev test:go` 
     a. Note that the test pass
     b. I had to disable deleting model plan from the plan milestones unit test, as it would violate a foreign key constraint

2. `Run  scripts/dev db:clean && scripts/dev db:seed`
    a. Note that there are records for the seeded model plans
    
3. Run the following GQL scripts to note new functionality

```gql
mutation createModelPlan {
  createModelPlan(input: {modelName: "My Custom Model", id: "9b5db64b-8d42-4b91-a601-d6f9732b5298"
status: PLAN_DRAFT}) {
    id
    createdBy
    collaborators {
      id
      euaUserID
      teamRole
      fullName
    }
  }
}
```

attempt to delete the plan collaborator that was created when the model plan was created. Update the id to be from the query above

```gql
mutation deletePlanCollaborator {
  deletePlanCollaborator(input: {
        id: "",
        modelPlanID: "9b5db64b-8d42-4b91-a601-d6f9732b5298",
        euaUserID: "MINT",
        teamRole: MODEL_LEAD,
        fullName: "mint Doe",

  }) {
      id
      euaUserID
      teamRole
      fullName
  }
}
```

Note that that should give you an error

add another model team lead with this query

```gql
mutation createPlanCollaborator {
  createPlanCollaborator(input: {
    id: "a76dc7b7-331d-4876-8ce2-926fbd06249c"
    modelPlanID: "9b5db64b-8d42-4b91-a601-d6f9732b5298"
    euaUserID: "WJZZ"
    fullName:"Steven Wade"

    teamRole: MODEL_LEAD
  }) {
    id
    modelPlanID
    euaUserID
    fullName

    teamRole
  }
}
```

Try to delete the initial collaborator again. It should now be allowed since there is another Model Lead Listed

Now try to update the newest collaborator to a role that isn't model lead. It won't be allowed since there is now only 1 model_lead

```gql
mutation updatePlanCollaborator {
  updatePlanCollaborator(input: {
    id: "a76dc7b7-331d-4876-8ce2-926fbd06249c"
    modelPlanID: "9b5db64b-8d42-4b91-a601-d6f9732b5298"
    euaUserID: "WJZZ"
    fullName:"Steven Wade"

    teamRole: LEADERSHIP
  }) {
    id
    modelPlanID
    euaUserID
    fullName

    teamRole
  }
}
```
## PR Author Review Checklist

## PR Reviewer Guidelines


pkg/graph/schema.resolvers.go
```diff
@@ -53,12 +53,20 @@ func (r *modelPlanResolver) Collaborators(ctx context.Context, obj *models.Model
 func (r *mutationResolver) CreateModelPlan(ctx context.Context, input model.ModelPlanInput) (*models.ModelPlan, error) {
 	logger := appcontext.ZLogger(ctx)
 	principal := appcontext.Principal(ctx).ID()
+	principalInfo, err := r.service.FetchUserInfo(ctx, principal)
+	if err != nil { //if can't get user info, use EUAID as commonName
```
StevenWadeOddball, line 57: Note this logic is likely to go away as we move away from storing user name in the database. This code address if there is a failed call to cedar to get the users first name here. If first name is kept in the DB, this should be revisited

scripts/dev
```diff
@@ -352,17 +352,16 @@ namespace :db do
...
+    plan_basics,
+    plan_collaborator,
+    plan_document,
+    plan_milestones,
+
+    model_plan"
+
     puts "Cleaning database..."
-    `echo "#{sql}" | psql`
+    `echo "TRUNCATE #{tableList}" | psql`
```
StevenWadeOddball, line 364: TRUNCATE allows delete of data without triggering `BEFORE DELETE` trigger used for plan collaborator.

pkg/graph/resolvers/model_planR.go
```diff
@@ -9,12 +9,22 @@ import (
...
 
 // ModelPlanCreate implements resolver logic to create a model plan
-func ModelPlanCreate(logger *zap.Logger, plan *models.ModelPlan, store *storage.Store) (*models.ModelPlan, error) {
+func ModelPlanCreate(logger *zap.Logger, plan *models.ModelPlan, store *storage.Store, principalInfo *models.UserInfo) (*models.ModelPlan, error) {
 
 	createdPlan, err := store.ModelPlanCreate(logger, plan)
 	if err != nil {
 		return nil, err
 	}
+	///TODO get user full name, need to acces CEDAR for this
```
OddTomBrooks, line 18: Is this still relevant?

pkg/graph/resolvers/model_planR.go
```diff
@@ -9,12 +9,22 @@ import (
...
+	colab := models.PlanCollaborator{
+		EUAUserID:   *createdPlan.CreatedBy,
+		ModelPlanID: createdPlan.ID,
+		TeamRole:    models.TeamRoleModelLead,
+		CMSCenter:   models.CMSCMMI,
+		FullName:    principalInfo.CommonName, //TOOD get this information from CEDAR
+	}
```
OddTomBrooks, line 25: Can be extracted to method

pkg/graph/resolvers/plan_collaborator_test.go
```diff
@@ -0,0 +1,67 @@
...
+	colab := models.PlanCollaborator{
+		FullName:    "Fake name",
+		EUAUserID:   "FAKE",
+		CreatedBy:   models.StringPointer("FAKE"),
+		ModifiedBy:  models.StringPointer("FAKE"),
+		ID:          uuid.MustParse("f85c15cd-f920-4080-bdeb-e79b3d8f056a"),
+		ModelPlanID: uuid.MustParse("85b3ff03-1be2-4870-b02f-55c764500e48"),
+		TeamRole:    models.TeamRoleLearning,
+		CMSCenter:   models.CMSCMMI,
+	}
```
OddTomBrooks, line 24: Can be extracted to method

pkg/graph/resolvers/plan_collaborator_test.go
```diff
@@ -0,0 +1,67 @@
...
+	colab := models.PlanCollaborator{
+		FullName:    "Fake name",
+		EUAUserID:   "FAKE",
+		CreatedBy:   models.StringPointer("FAKE"),
+		ModifiedBy:  models.StringPointer("FAKE"),
+		ID:          uuid.MustParse("f85c15cd-f920-4080-bdeb-e79b3d8f056a"),
+		ModelPlanID: uuid.MustParse("85b3ff03-1be2-4870-b02f-55c764500e48"),
+		TeamRole:    models.TeamRoleLearning,
+		CMSCenter:   models.CMSCenterForMedicare,
+	}
```
OddTomBrooks, line 42: Can be extracted to method

pkg/graph/resolvers/plan_collaborator_test.go
```diff
@@ -0,0 +1,67 @@
...
+	modelPlanID := uuid.MustParse("85b3ff03-1be2-4870-b02f-55c764500e48")
+
+	result, err := FetchCollaboratorsByModelPlanID(tc.Logger, tc.Principal, modelPlanID, tc.Store)
+	assert.NoError(t, err)
+
+	assert.NotNil(t, result)
+
+}
+
+//TODO add test to make sure you can't delete the last model lead when logic is added
```
OddTomBrooks, line 63: Is this still relevant?

pkg/graph/resolvers/planmilestones_test.go
```diff
@@ -41,12 +45,13 @@ func TestCreatePlanMilestonesResolver(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Equal(t, 1, int(rowsAffected))
 
-	sqlResult, err = store.ModelPlanDeleteByID(logger, plan.ID)
-	assert.NoError(t, err)
+	//TODO buisness logic not implemented to delete model plan when collaborator etc are on the object.
+	// sqlResult, err = store.ModelPlanDeleteByID(logger, plan.ID)
+	// assert.NoError(t, err)
```
OddTomBrooks, line 50: Is this still relevant?

pkg/graph/resolvers/planmilestones_test.go
```diff
@@ -78,10 +86,11 @@ func TestFetchPlanMilestonesByID(t *testing.T) {
...
 
-	rowsAffected, err = sqlResult.RowsAffected()
-	assert.NoError(t, err)
-	assert.Equal(t, 1, int(rowsAffected))
+	// rowsAffected, err = sqlResult.RowsAffected()
+	// assert.NoError(t, err)
+	// assert.Equal(t, 1, int(rowsAffected))
```
OddTomBrooks, line 95: Is this still relevant?

OddTomBrooks (commented):
Mostly superficial checks and cleanup, well done!

pkg/graph/resolvers/model_planR.go
StevenWadeOddball: it could, but it's not being reused at this point. I don't think it needs to be here, I think this is probably more readable in this instance.

StevenWadeOddball:
Note, this PR likely will need to be updated after https://github.com/CMSgov/mint-app/pull/55 is merged into main.

pkg/graph/resolvers/model_plan.go
```diff
@@ -9,12 +9,21 @@ import (
 )
 
 // ModelPlanCreate implements resolver logic to create a model plan
-func ModelPlanCreate(logger *zap.Logger, plan *models.ModelPlan, store *storage.Store) (*models.ModelPlan, error) {
+func ModelPlanCreate(logger *zap.Logger, plan *models.ModelPlan, store *storage.Store, principalInfo *models.UserInfo) (*models.ModelPlan, error) {
 
 	createdPlan, err := store.ModelPlanCreate(logger, plan)
```
ClayBenson94, line 14: I wonder if it makes sense to have a `store` method that could wrap
1. Create Model Plan
2. Create Plan Collaborator

in a [sqlx transaction](https://jmoiron.github.io/sqlx/#transactions) so that they're done atomically 🤔 

This would have the benefit of making the error handling in this function simpler, as you wouldn't have to handle the `err` of `ModelPlanCreate` and the `err` of `CreatePlanCollaborator` individually.

Might not be worth the lift right now, but a thought to consider any time we have multiple unique SQL statements to make as part of a single resolver.

migrations/V9__Plan_Collaborator_Lead_Trigger.sql
```diff
@@ -0,0 +1,29 @@
...
+    RETURN NEW;
+    END IF;
+END
+$role_check$ LANGUAGE plpgsql;
+
+
+CREATE TRIGGER collaborator_lead_req_update
+    BEFORE UPDATE ON plan_collaborator
+	FOR EACH ROW
+    WHEN ((OLD.team_role = 'MODEL_LEAD') AND (NEW.team_role != 'MODEL_LEAD'))
```
ClayBenson94, line 21: My knowledge of SQL triggers is pretty sparse - should this be an `OR` instead of an `AND`?

migrations/V9__Plan_Collaborator_Lead_Trigger.sql
```diff
@@ -0,0 +1,29 @@
...
+ IF (SELECT count(*)
+ FROM plan_collaborator
+ WHERE  team_role = 'MODEL_LEAD' and model_plan_id = OLD.model_plan_id 
+ )<2
+ THEN
+    RAISE EXCEPTION 'There must be at least one MODEL_LEAD assigned to each model plan';
+     END IF;
+     IF (TG_OP = 'DELETE') THEN RETURN OLD;
+     ELSE
+    RETURN NEW;
```
ClayBenson94, line 12: Can you explain this `if delete, return old, else return new` block? I don't think I quite understand why we always wouldn't return `OLD`, since the operation always is failed in this `IF` block I think

ClayBenson94 (changes requested):

migrations/V9__Plan_Collaborator_Lead_Trigger.sql
StevenWadeOddball: Yeah, so this is looking at the operation that happened. This trigger happens on delete or update statements.  This is basically saying if it was a delete statement, return the row that we just deleted, but if it's an update statement,  return the brand new updated row. Otherwise, we are not returning the new data.

migrations/V9__Plan_Collaborator_Lead_Trigger.sql
StevenWadeOddball: I don't think so for this situation, I only want to trigger this statement if the row that I'm updating was already a MODEL_LEAD and is changing to something that isn't a model lead.  Or will trigger this statement a lot more than is needed in this situation.

ClayBenson94: Derp - I misread `!=` as `=`.

pkg/graph/resolvers/model_plan.go
StevenWadeOddball: Hmm  🤔, not a bad thought

The only thing I don't like about that is you would have a store file for model_plan, that would also be creating database records for plan collaborator. That could be fine.

I guess the other question is, would we want to not create the model plan if the plan collaborator fails to be inserted? If so, the sqlx transaction would make sense so we could roll it back

ClayBenson94: That was what I was thinking (roll it back if we can't make the collaborator) - because if you fail to make the collaborator, you're left in a spot where nobody can access the plan since there's 0 collaborators.

I don't think we need to solution this right now, but can you leave a `//TODO` comment in there clarifiying it's something we should be aware of?

ClayBenson94: Also - I agree with you on that - the nice part about Go packages is we can have a file for methods like this while keeping it in the same package

```go
package store
```

in a file like `pkg/store/model_plan_collab.go` - though I'm not sold on this right now lol

StevenWadeOddball: that's a good thought, and it isn't even necessarily a problem to have the action happen in both stores. Another option which I kind of like is to roll it into one SQL statement. I don't know if there is ever a situation where we want to make a model plan without a collaborator. The main thing I don't like about that is you have multiple entry points to write to the table. It can just start to get messy, thought it would be the most efficient method

pkg/graph/resolvers/plan_collaborator_test.go
StevenWadeOddball: thanks, added a unit test for this.

pkg/graph/resolvers/model_plan.go
StevenWadeOddball: added a comment to address this for now

migrations/V9__Plan_Collaborator_Lead_Trigger.sql
```diff
@@ -0,0 +1,29 @@
+CREATE FUNCTION collaborator_role_check_trigger() RETURNS TRIGGER AS $role_check$
+BEGIN
+ IF (SELECT count(*)
+ FROM plan_collaborator
+ WHERE  team_role = 'MODEL_LEAD' and model_plan_id = OLD.model_plan_id 
+ )<2
```
ClayBenson94, line 6: Why is this `<2` instead of `<1`?

ClayBenson94 (changes requested):

migrations/V9__Plan_Collaborator_Lead_Trigger.sql
StevenWadeOddball: less than 2 captures that there is only 1 `MODEL_LEAD` selected. <1 means there aren't any `MODEL_LEAD`s. 

Otherwise we could represent this as <= 1. 

This is important because this function is used as a `BEFORE` trigger, so the check is happening before we make a change instead, of checking after a change and rolling back.

ClayBenson94: Ahh - the face that it's a `BEFORE` helps a ton. Would you mind just chucking a quick comment on this for anyone looking in the future?

PR #56, opened 2022-04-27
[NOREF] Added SQLFluff linter

ClayBenson94:
# NOREF

## Changes and Description

- Adds and runs `sqlfluff` linter on all `.sql` files

It's worthy of noting that `sqlfluff` is in beta, and has trouble parsing some statements (e.g. `UPDATE...RETURNING` statements).

It should still do a decent job catching most things in most of our SQL, though.

## How to test this change

```bash
echo 'I should also look over the modified .sql files to make sure nothing broke!'
echo 'I can also modify some .sql files with things that should break a linter, and see if it breaks!'
scripts/dev lint
```

image

## PR Author Review Checklist

## PR Reviewer Guidelines


StevenWadeOddball (commented):
I like this @ClayBenson94 , I'm curious, do you know if we can configure more of the rules? I like the capitalization enforcement, i wish it would capitalize data types instead of enforce lower case eg forcing `TEXT` to `text`. That's just a preference of mine, at the end of the day uniformity is more imporant

ClayBenson94:
> I like this @ClayBenson94 , I'm curious, do you know if we can configure more of the rules? I like the capitalization enforcement, i wish it would capitalize data types instead of enforce lower case eg forcing `TEXT` to `text`. That's just a preference of mine, at the end of the day uniformity is more imporant

Yes! 00f4bd40dca1be870745730e4d54c073681026a4 adds a few things
- Comments in the `.sqlfluff` file linking to the rules page
- Comments in the `.sqlfluff` file linking / describing each rule
- Additions of rules L040 and L063, which handle capitalizing TRUE/FALSE/NULL, and datatypes like `TEXT` or `EUA_ID`

StevenWadeOddball (approved):
Thanks Clay! I like it, I think this looks good to me! If we want to change any formatting configurations later, I think we will be in a good place from here. Regarding table name aliases, I think you are right for now that we can explicitly deny aliases. If we start doing more than 4 joins consistently, we might want to change it, but I can't imagine the use case for our application at the moment.

PR #59, opened 2022-04-29
Easi 1930/plan documents

OddTomBrooks:
# EASI-1930

## Changes and Description

CRUD statements

- GQL backend
    - Create
        - Ability to upload documents
    - Read
        - Ability to retrieve the documents

### Create
```
mutation createModelPlan {
  createModelPlan(input: {modelName: "My Custom Model"}) {
    id
  }
}
```

### Read
```
mutation createPlanDocument ($input: PlanDocumentInput!) {
    createPlanDocument (input: $input) {
        document {
            id
            modelPlanID
            fileType
            bucket
            fileKey
            virusScanned
            virusClean
            fileName
            fileSize
            documentType
            otherType
            deletedAt
            createdBy
            createdDts
            modifiedBy
            modifiedDts
        }
        presignedURL
    }
}
```
```
{
  "input": {
    "id": null,
    "modelPlanID": "4d51fe5e-73ee-4f06-8dbb-440c1dec761a",
    "documentParameters": {
      "fileName": "test_file.pdf",
      "fileSize": 512512,
      "fileType": "application/pdf",
      "documentType": "OTHER",
      "otherTypeDescription": "A fake document"
    },
    "url": "http://localhost:9000/easi-test-bucket/e9eb4a4f-9100-416f-be5b-f141bb436cfa.pdf?X-Amz-Algorithm=AWS4-HMAC-SHA256&"
  }
}
```

## PR Author Review Checklist

## PR Reviewer Guidelines


pkg/graph/resolvers/plan_document.go
```diff
@@ -0,0 +1,80 @@
...
+	payload := model.PlanDocumentPayload{
+		Document:     document,
+		PresignedURL: presignedURL,
+	}
+
+	return &payload, nil
+}
+
+// PlanDocumentCreate implements resolver logic to create a plan document object
+func PlanDocumentCreate(logger *zap.Logger, input *model.PlanDocumentInput, principal *string, store *storage.Store, s3Client *upload.S3Client) (*model.PlanDocumentPayload, error) {
```
StevenWadeOddball, line 28: I think for consistency, this shouldn't take the generated `model.PlanDocumentInput`, but instead take the `models.PlanDocument`. The other resolvers convert the input to the hand rolled object in schema.resolvers.go.

pkg/graph/resolvers/plan_document.go
```diff
@@ -0,0 +1,80 @@
...
+	document, err := store.PlanDocumentCreate(logger, principal, input, s3Client)
+	if err != nil {
+		return nil, genericmodel.HandleModelUpdateError(logger, err, models.PlanDocument{ID: *input.ID})
+	}
+
+	return createDocumentPayload(s3Client, document)
+}
+
+// PlanDocumentRead implements resolver logic to fetch a plan document object by ID
+func PlanDocumentRead(logger *zap.Logger, id uuid.UUID, store *storage.Store) (*models.PlanDocument, error) {
```
StevenWadeOddball, line 38: I think `get` would be more consistent in this spot as well.

pkg/graph/resolvers/plan_document.go
```diff
@@ -0,0 +1,80 @@
...
+	document, err := store.PlanDocumentRead(logger, id)
+	if err != nil {
+		return nil, err
+	}
+
+	return document, nil
+}
+
+// PlanDocumentsReadByModelPlanID implements resolver logic to fetch a plan document object by model plan ID
+func PlanDocumentsReadByModelPlanID(logger *zap.Logger, id uuid.UUID, store *storage.Store, s3Client *upload.S3Client) ([]*models.PlanDocument, error) {
```
StevenWadeOddball, line 48: same as above, I suggest naming this `get`

pkg/graph/schema.graphql
```diff
@@ -1,292 +1,354 @@
...
+}
+
+"""
+PlanDocumentCreateParameters represents the specific data required to create or modify a document on a plan
+"""
+input PlanDocumentParameters {
+  fileName: String
+  fileSize: Int!
+  fileType: String
+  documentType: String
```
StevenWadeOddball, line 132: this could be a type in GQL so front end knows accepted values

pkg/graph/schema.resolvers_helpers.go
```diff
@@ -5,7 +5,7 @@ import (
 	"github.com/cmsgov/mint-app/pkg/models"
 )
 
-// ConvertToModelPlan takes an auto-generated model plan input and converts it to a hand-written one
```
StevenWadeOddball, line 8: I don't think that these comments, variable name changes / or anything not directly concerning a `Document` object should be touched in this branch. It increases the chance of conflicts, and is not the work scoped for this branch. If the change should be made, I think we should make a separate branch, or at least a ticket to do it in the future.

pkg/shared/utilitySQL/query_construction.go
```diff
@@ -1,16 +1,19 @@
...
+
+// CreateIDQueryMap creates a map with key "id" which maps to the passed in id
+func CreateIDQueryMap(id uuid.UUID) map[string]interface{} {
+	return map[string]interface{}{
+		"id": id,
+	}
+}
+
+// CreateModelPlanIDQueryMap creates a map with key "model_plan_id" which maps to the passed in id
+func CreateModelPlanIDQueryMap(id uuid.UUID) map[string]interface{} {
```
StevenWadeOddball, line 15: This is fine, I think leaving the content of the function inline might be clearer code, since the function isn't doing very much.

pkg/storage/SQL/plan_document_create.sql
```diff
@@ -0,0 +1,33 @@
...
+        :virus_clean,
+        :file_name,
+        :file_size,
+        :document_type,
+        :other_type,
+        :deleted_at,
+        :created_by,
+        :modified_by
+    )
+    RETURNING *;
```
StevenWadeOddball, line 33: new linter doesn't allow *

pkg/storage/plan_documentStore.go
```diff
@@ -0,0 +1,229 @@
...
+var planDocumentGetByModelPlanIDSQL string
+
+//go:embed SQL/plan_document_delete_by_id.sql
+var planDocumentDeleteByIDSQL string
+
+// PlanDocumentCreate creates a plan document
+func (s *Store) PlanDocumentCreate(
+	logger *zap.Logger,
+	principal *string,
+	input *model.PlanDocumentInput,
```
StevenWadeOddball, line 42: I think we should ideally use the models.PlanDocument here

StevenWadeOddball (changes requested):
Good work! I focused mostly on paradigm consistency eg, where do we convert `input` to `hand rolled models`. Also I don't know how hard it would be to provide a mock document for seeded data, but I think that would be useful, or at the very least seed a data row.

ClayBenson94 (changes requested):
I think the main item to be addressed is answering the question:

> How, as a FE caller of these GQL endpoints, do I get a pre-signed URL to use as a `url` in `createPlanDocument`?

EASi exposes a GQL query to do this, and I think we should do the same!

Also - looks like there's some files that might have been touched by an editor, and some files that don't quite lint properly. Might be worth checking those files out from `main` if the intent wasn't to touch them!

OddTomBrooks:
@ClayBenson94 Yes, there were a few unpublished changes. They are pushed up now.

pkg/models/plan_document.go
```diff
@@ -0,0 +1,49 @@
...
+	VirusScanned bool `json:"virusScanned" db:"virus_scanned"`
+	VirusClean   bool `json:"virusClean" db:"virus_clean"`
+
+	FileName             *string    `json:"fileName" db:"file_name"`
+	FileSize             int        `json:"fileSize" db:"file_size"`
+	DocumentType         *string    `json:"documentType" db:"document_type"`
+	OtherTypeDescription *string    `json:"otherType" db:"other_type"`
+	DeletedAt            *time.Time `json:"deletedAt" db:"deleted_at"`
+
+	CreatedBy   *string    `json:"createdBy" db:"created_by"`
```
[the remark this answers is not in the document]
OddTomBrooks: Discussed with @ClayBenson94 and decided to resolve this in a dedicated PR to resolve all such cases

pkg/graph/resolvers/plan_document.go
OddTomBrooks: @ClayBenson94 Let's have a discussion on naming conventions. I went with Read as it matches the CRUD acronym, but Steven raises a good point and we should align one way or another.

ClayBenson94 (approved):
Let's get er in!

PR #63, opened 2022-05-03
[EASI-1997] Add 'NOT NULL' constraints to all created and modified by fields

ClayBenson94:
# EASI-1997

## Changes and Description

- Added `NOT NULL` constraint to all `created_by` and `modified_by` fields in the DB.
- Normalize resolver functions to set `CreatedBy` and `ModifiedBy` in the `resolvers` package functions

## How to test this change

Pretty much running through the GQL queries should validate these changes. Additionally, running `scripts/dev db:seed` should also seed the DB and show that all required fields are being populated.

## PR Author Review Checklist

## PR Reviewer Guidelines


pkg/graph/schema.resolvers.go
```diff
@@ -81,8 +81,6 @@ func (r *mutationResolver) CreateModelPlan(ctx context.Context, input model.Mode
 
 	plan := ConvertToModelPlan(&input)
 
-	plan.CreatedBy = &principal
```
StevenWadeOddball, line 84: thanks for making this consistent! I think this is the paradigm that we should follow.

StevenWadeOddball (approved):
Looks good to me! Thanks Clay! I tested creating a few objects, and I ran `scripts/dev test:go` successfully

pkg/graph/schema.resolvers.go
ClayBenson94: Agree. I think we may even introduce some generic "always modify created by, modified by" logic somewhere that's shared, as I think this is readable, but probably easy to forget to add to a resolver!

PR #68, opened 2022-05-04
[Easi-1932] Draft Model Plan - Basics

garyjzhao:
# EASI-1932

## Changes and Description

- Create each form page to capture required information as it pertains to Basic task list
- Form saving (GQL mutation) must happen after each page, on 'Save and Exit', as well as a 3 sec debounce for text input
- Display alerts and error handling as illustrated in Figma
- Dynamic form components must behave as illustrated by Figma
- Breadcrumbs
- Page numbers
- All static text must use translations
- Form validations for every step
- Cypress Testing
- Unit testing

## PR Author Review Checklist

## PR Reviewer Guidelines


src/views/ModelPlan/TaskList/Basics/Milestones/index.scss
```diff
@@ -0,0 +1,7 @@
+.milestone-form {
+  .fieldGroup__wrapper {
+    display: flex;
+    grid-gap: units(3);
+    align-items: flex-end;
+  }
```
garyjzhao, line 6: Using util classes had weird alignment issues on the FE when there is an error. See images below
|   |   |
|:--:|---|
|with util classes|Screen Shot 2022-05-04 at 4 24 50 PM|
|with sass styles|Screen Shot 2022-05-04 at 4 25 09 PM|

src/views/ModelPlan/TaskList/index.tsx
```diff
@@ -68,7 +70,18 @@ const TaskList = () => {
...
+        if (basics === null) {
+          return 'READY';
+        }
+        if (
+          basics?.status === 'COMPLETE' &&
+          milestones?.status === 'COMPLETE'
+        ) {
+          return 'COMPLETE';
+        }
+        return 'IN_PROGRESS';
```
garyjzhao, line 82: On second thought, this logic probably shouldn't be here. 

@ClayBenson94 @StevenWadeOddball 
Will there be a field in the BE that I can use to determine when the section is `READY`, `IN PROGRESS` or `COMPLETE`

cc @patrickseguraoddball

src/views/ModelPlan/TaskList/Basics/index.tsx
```diff
@@ -1,29 +1,440 @@
...
+
+  const [update] = useMutation<UpdateModelPlanType>(UpdateModelPlan);
+
+  const handleFormSubmit = (
+    formikValues: PlanBasicModelPlanFormType,
+    isAutoSave: boolean = false,
+    continuesToNextStep: boolean = true
+  ) => {
+    update({
+      variables: {
```
patrickseguraoddball, line 83: Per Clay's recent PR, the input for `UpdateModelPlan` need to be in the format: 

`{
        id: modelID,
        changes: {
                ...fields
        }
}`

patrickseguraoddball (changes requested):

src/views/ModelPlan/TaskList/Basics/index.tsx
```diff
@@ -1,29 +1,439 @@
...
+} from 'utils/modelPlan';
+import planBasicsSchema from 'validations/planBasics';
+import { NotFoundPartial } from 'views/NotFound';
+
+import Milestones from './Milestones';
+import Overview from './Overview';
+
+type PlanBasicModelPlanFormType = {
+  modelName: string;
+  modelCategory: string;
```
patrickseguraoddball, line 48: `modelCategory` should be `string` or `null`

src/views/ModelPlan/TaskList/Basics/index.tsx
```diff
@@ -1,29 +1,439 @@
...
+    formikValues: PlanBasicModelPlanFormType,
+    isAutoSave: boolean = false,
+    continuesToNextStep: boolean = true
+  ) => {
+    update({
+      variables: {
+        id: modelId,
+        changes: {
+          modelName: formikValues.modelName,
+          modelCategory:
```
patrickseguraoddball, line 87: With change `modelCategory` to be possibly null, you can have the input value purely be `formikValues.modelCategory`

src/views/ModelPlan/TaskList/Basics/index.tsx
```diff
@@ -1,29 +1,439 @@
...
+        }
+      })
+      .catch(errors => {
+        formikRef?.current?.setErrors(errors);
+      });
+  };
+
+  const initialValues: PlanBasicModelPlanFormType = {
+    modelName: modelName as string,
+    modelCategory: modelCategory ?? '',
```
patrickseguraoddball, line 113: Here `modeCategory` default should be `null` instead of ''

src/views/ModelPlan/TaskList/Basics/index.tsx
```diff
@@ -1,29 +1,439 @@
...
+                      className="margin-top-4"
+                    >
+                      <Label htmlFor="plan-basics-model-category">
+                        {t('modelCategory')}
+                      </Label>
+                      <FieldErrorMsg>{flatErrors.modelCategory}</FieldErrorMsg>
+                      <Field
+                        as={Dropdown}
+                        id="plan-basics-model-category"
+                        name="role"
```
patrickseguraoddball, line 233: `name` should be `modelCategory`

src/views/ModelPlan/TaskList/Basics/index.tsx
```diff
@@ -1,29 +1,439 @@
...
+                        value={values.modelCategory}
+                        onChange={(e: any) => {
+                          setFieldValue('modelCategory', e.target.value);
+                        }}
+                      >
+                        <option value="" key="default-select" disabled>
```
patrickseguraoddball, line 239: Here you can remove the value prop of the default option and instead add the prop `selected` to have the null value be the default while still displaying the `-Select-` string

src/views/ModelPlan/TaskList/Basics/Overview/index.tsx
```diff
@@ -0,0 +1,333 @@
...
+    modelType: modelType ?? '',
+    problem: problem ?? '',
+    goal: goal ?? '',
+    testInterventions: testInventions ?? '',
+    note: note ?? ''
+  };
+
+  const handleFormSubmit = (
+    formikValues: PlanBasicsOverviewTypes,
+    isAutoSave: boolean = false,
```
patrickseguraoddball, line 78: I think a rework of how we handle redirects on all types of form submissions is needed.  Instead of passing multple props to the submission, I suggest a single prop like `redirect`.  It can be optional and it would look something like this - 

`redirect?: 'next' | 'back' | 'task-list'`

Where the absence of the prop would signal an autosave, and other submissions pass in one of those string props to tell where to redirect after successful mutation.

src/views/ModelPlan/TaskList/Basics/Overview/index.tsx
```diff
@@ -0,0 +1,333 @@
...
+          problem: formikValues.problem,
+          goal: formikValues.goal,
+          testInventions: formikValues.testInterventions,
+          note: formikValues.note
+        }
+      }
+    })
+      .then(response => {
+        if (!response?.errors) {
+          if (continuesToNextStep) {
```
patrickseguraoddball, line 96: On a successful response the redirect prop would trigger a conditional like this

```      
          if (redirect === 'next') {
            history.push('/models/${modelId}/task-list/basics/milestones');
          } else if (redirect === 'back') {
            history.push('/models/${modelId}/task-list/basics');
          } else if (redirect === 'task-list') {
            history.push('/models/${modelId}/task-list/');
          }
```

src/views/ModelPlan/TaskList/Basics/Overview/index.tsx
```diff
@@ -0,0 +1,333 @@
...
+  );
+
+  const { modelName } = data?.modelPlan || {};
+  const { id, modelType, problem, goal, testInventions, note } =
+    data?.modelPlan?.basics || {};
+
+  const [update] = useMutation<UpdatePlanBasicsType>(UpdatePlanBasics);
+
+  const initialValues = {
+    modelType: modelType ?? '',
```
patrickseguraoddball, line 69: `modelType` default should be null instead of ''

src/views/ModelPlan/TaskList/Basics/Milestones/index.tsx
```diff
@@ -0,0 +1,647 @@
...
+    performancePeriodEnds: milestones?.performancePeriodEnds ?? null,
+    wrapUpEnds: milestones?.wrapUpEnds ?? null,
+    highLevelNote: milestones?.highLevelNote ?? '',
+    phasedIn: milestones?.phasedIn ?? undefined,
+    phasedInNote: milestones?.phasedInNote ?? ''
+  };
+
+  const handleFormSubmit = (
+    formikValues: PlanBasicsOverviewTypes,
+    isAutoSave: boolean = false
```
patrickseguraoddball, line 97: Same solution as mentioned in `Overview` could be used here.  And each place you call the submission sohuld be updated to reflect the props

src/validations/planBasics.ts
```diff
@@ -0,0 +1,45 @@
+import * as Yup from 'yup';
+
+const datePickerSchema = Yup.date()
+  .nullable()
+  .min(new Date(), 'Date cannot be in the past');
+
+const planBasicsSchema = {
+  pageOneSchema: Yup.object().shape({
+    modelName: Yup.string().trim().required('Enter the Model Name'),
+    modelCategory: Yup.string().required('Enter the Model Category'),
```
patrickseguraoddball, line 10: This should be - `modelCategory: Yup.string().nullable().required('Enter the Model Category')`

src/validations/planBasics.ts
```diff
@@ -0,0 +1,45 @@
...
+    cmsCenters: Yup.array()
+      .min(1, 'Select a CMS Component')
+      .required('Select a CMS Component')
+  }),
+
+  pageOneSchemaWithCmmiGroup: Yup.object().shape({
+    modelName: Yup.string().trim().required('Enter the Model Name'),
+    modelCategory: Yup.string().required('Enter the Model Category'),
```
patrickseguraoddball, line 18: This should be - `modelCategory: Yup.string().nullable().required('Enter the Model Category')`

src/validations/planBasics.ts
```diff
@@ -0,0 +1,45 @@
...
+    cmsCenters: Yup.array()
+      .min(1, 'Select a CMS Component')
+      .required('Select a CMS Component'),
+    cmmiGroup: Yup.array()
+      .min(1, 'Select a CMMI Group')
+      .required('Select a CMMI Group')
+  }),
+
+  pageTwoSchema: Yup.object().shape({
+    modelType: Yup.string().required('A model type is required')
```
patrickseguraoddball, line 28: This should be - `modelType: Yup.string().nullable().required('A model type is required')`

patrickseguraoddball (changes requested):
A few additional notes:

- The checkboxes actually extend the whole page and select the option even if clicking very far from the checkbox

- You should be able to remove a date from the milestones, but it throws a validation error when trying to remove a date


Just want to say great work Gary!  This was a lot of work

src/views/ModelPlan/TaskList/Basics/index.tsx
```diff
@@ -1,29 +1,439 @@
...
+      }
+    }
+  );
+
+  const { modelName, modelCategory, cmsCenters, cmmiGroups, cmsOther } =
+    data?.modelPlan || {};
+
+  const [update] = useMutation<UpdateModelPlanType>(UpdateModelPlan);
+
+  const handleFormSubmit = (
```
patrickseguraoddball, line 77: We need to ensure a user can't erase a model name.  Because autosave and `save and exit` don't use the form validation to progress, we need another check to ensure those two saves can't happen if the model name is empty.  Currently I can erase the model name, autosave, and it is permanently erased.  This may need to happen in the submit handlers to ensure no saves can happen in that field is empty.

src/validations/planBasics.ts
```diff
@@ -0,0 +1,45 @@
...
+
+const planBasicsSchema = {
+  pageOneSchema: Yup.object().shape({
+    modelName: Yup.string().trim().required('Enter the Model Name'),
+    modelCategory: Yup.string().nullable().required('Enter the Model Category'),
+    cmsCenters: Yup.array()
+      .min(1, 'Select a CMS Component')
+      .required('Select a CMS Component')
+  }),
+
```
patrickseguraoddball, line 15: I think we need a validation for if a user selects `other` - a specified name is required to progress?  May need to check with UX on this to be sure

src/views/ModelPlan/TaskList/Basics/Milestones/index.tsx
```diff
@@ -0,0 +1,657 @@
...
+            if (redirect === 'task-list') {
+              history.push(`/models/${modelId}/task-list`);
+            }
+          }
+        })
+        .catch(errors => {
+          formikRef?.current?.setErrors(errors);
+        });
+    } else {
+      create({
```
patrickseguraoddball, line 135: I've gotten this error message in milestones quite a few times.  I think it's related to trying to recreate the milestones object.  This may be resolved by Clay's PR, but currently is throwing errors.  I think the autosave is submitting multiple create methods.

`pq: duplicate key value violates unique constraint "plan_milestones_model_plan_id_key"
createPlanMilestones`

patrickseguraoddball (changes requested):
Thanks for the changes Gary!  I think we could use some unit tests for each the basics, overview, and milestones as well.

src/views/ModelPlan/TaskList/Basics/Milestones/index.tsx
ClayBenson94: If this is happening infrequently, it might just be worth not burning too much time on it. As @patrickseguraoddball mentioned, https://github.com/CMSgov/mint-app/pull/85 sets out to eliminate the need for this `if/else` block entirely.

src/views/ModelPlan/TaskList/Basics/index.tsx
ClayBenson94: It might be worth considering what pages get auto-saved, if possible. Maybe certain ones (like this whole page, since it has no long-form text input) don't even have auto-save, just to simplify things?

Given that the inputs on this page are
- Model Name (do we want that to be auto-saved?)
- Category (Dropdown)
- Centers (Multi-select/dropdown)
- CMMI Groups (Multi-select/dropdown)
- "CMSOther" (A small text field)

maybe we just say "eh, this page doesn't do auto-save"

Is that taking it too far? Thoughts?

patrickseguraoddball: I think the `modelName` is the only field on the entire form that we don't want to autosave.  Everything else, even if required by validation is fine, as they won't be able to progress without filling it out.  I can also see the side of just removing autosave if it isn't providing value on forms with ~2 input

src/views/ModelPlan/TaskList/Basics/index.tsx
```diff
@@ -1,29 +1,480 @@
...
+                    <div className="margin-top-6 margin-bottom-3">
+                      <Button
+                        type="submit"
+                        disabled={!(dirty || isValid)}
+                        onClick={() => setErrors({})}
+                      >
+                        {h('next')}
+                      </Button>
+                    </div>
+                    <Button
```
patrickseguraoddball, line 404: I think these two save methods are too complex for what they need to be.  Both can be the normal submit functionality, except for the autosave.  I think it makes most sense to not autosave at all of the modelName is empty.  If** we do want to autosave, we can handle that in the submit handler function.

src/views/ModelPlan/TaskList/Basics/index.tsx
```diff
@@ -1,29 +1,480 @@
...
+                    >
+                      {/* TODO:
+                        2. Ensure all steps of form works
+                        3. milestone errors
+                      */}
+                      <IconArrowBack className="margin-right-1" aria-hidden />
+                      {h('saveAndReturn')}
+                    </Button>
+                  </Form>
+                  <AutoSave
```
patrickseguraoddball, line 423: This autosave submithandler can behave as normal, except a conditional for if there is a modelName

```
<AutoSave
                    values={values}
                    onSave={() => {
                      if (formikRef.current!.values.modelName)
                        handleFormSubmit(formikRef.current!.values);
                    }}
                    debounceDelay={3000}
```

src/views/ModelPlan/TaskList/Basics/index.tsx
```diff
@@ -1,29 +1,480 @@
...
+  const { modelName, modelCategory, cmsCenters, cmmiGroups, cmsOther } =
+    data?.modelPlan || {};
+
+  const [update] = useMutation<UpdateModelPlanType>(UpdateModelPlan);
+
+  const handleFormSubmit = (
+    formikValues: PlanBasicModelPlanFormType,
+    redirect?: 'next' | 'back'
+  ) => {
+    update({
```
patrickseguraoddball, line 82: Before the update method, we can now validate for modelName that gets submitted

```
if (!formikValues.modelName) {
      formikRef?.current?.setFieldError('modelName', 'Enter the Model name');
      return;
    }

This will only trigger on the Save&Exit submit handler, as the autosave will not get called, and the normal form submit uses a different validation trigger.  
```

src/views/ModelPlan/TaskList/Basics/index.tsx
```diff
@@ -1,29 +1,480 @@
...
+import { NotFoundPartial } from 'views/NotFound';
+
+import Milestones from './Milestones';
+import Overview from './Overview';
+
+type PlanBasicModelPlanFormType = {
+  modelName: string;
+  modelCategory: string | null;
+  cmsCenters: string[];
+  cmmiGroup: string[];
```
patrickseguraoddball, line 50: I think this should be `cmmiGroups` rather than `cmmiGroup` to reflect the gql field and make types clearer.  All the references in this file should reflect this.

src/views/ModelPlan/TaskList/Basics/Overview/index.tsx
```diff
@@ -0,0 +1,337 @@
...
+                          label={t('Notes')}
+                        />
+                      </FieldGroup>
+                    )}
+
+                    <div className="margin-top-6 margin-bottom-3">
+                      <Button
+                        type="button"
+                        className="usa-button usa-button--outline margin-bottom-1"
+                        onClick={() => history.goBack()}
```
patrickseguraoddball, line 289: According to UX, we rarely want to use the browser's back functionality and instead want to always route to the previous form page.  In the previous comment, there was another option for 'back' as a route prop for form submission that can be used to explicitly set the previous route.

src/views/ModelPlan/TaskList/Basics/Milestones/index.tsx
```diff
@@ -0,0 +1,629 @@
...
+                          label={t('Notes')}
+                        />
+                      </FieldGroup>
+                    )}
+
+                    <div className="margin-top-6 margin-bottom-3">
+                      <Button
+                        type="button"
+                        className="usa-button usa-button--outline margin-bottom-1"
+                        onClick={() => history.goBack()}
```
patrickseguraoddball, line 581: Same as basics - According to UX, we rarely want to use the browser's back functionality and instead want to always route to the previous form page

src/views/ModelPlan/TaskList/Basics/Overview/index.tsx
```diff
@@ -0,0 +1,337 @@
...
+        }
+      })
+      .catch(errors => {
+        formikRef?.current?.setErrors(errors);
+      });
+  };
+
+  return (
+    <MainContent className="margin-bottom-5">
+      <div className="grid-container">
```
patrickseguraoddball, line 110: We can/should use Truss `Grid` components for setting up grids

src/views/ModelPlan/TaskList/Basics/Milestones/index.tsx
```diff
@@ -0,0 +1,629 @@
...
+        }
+      })
+      .catch(errors => {
+        formikRef?.current?.setErrors(errors);
+      });
+  };
+
+  return (
+    <MainContent className="margin-bottom-5">
+      <div className="grid-container">
```
patrickseguraoddball, line 129: We can/should use Truss `Grid` components for setting up grids

src/views/ModelPlan/TaskList/Basics/Milestones/index.tsx
```diff
@@ -0,0 +1,629 @@
...
+                values
+              } = formikProps;
+              const flatErrors = flattenErrors(errors);
+              const handleOnchange = (e: any, field: string) => {
+                if (e === '') {
+                  return;
+                }
+                try {
+                  setErrors({});
+                  new Date(e).toISOString();
```
patrickseguraoddball, line 193: Not sure why this line is here

patrickseguraoddball (changes requested):

src/views/ModelPlan/TaskList/Basics/Milestones/index.tsx
garyjzhao: The try/catch statement is an error handler. It tries `new Date(e).toISOString();` to see if it can actually convert `e` to a Date. If it is fine, it continues through the function (setting the field value and validation the field). If it errors, it will go to the `catch`  section.

patrickseguraoddball: Wouldn't the catch case trigger without that when trying to set the date in the state setter method?

src/views/ModelPlan/TaskList/Basics/Milestones/index.tsx
```diff
@@ -0,0 +1,634 @@
...
+
+                    <div className="margin-top-6 margin-bottom-3">
+                      <Button
+                        type="button"
+                        className="usa-button usa-button--outline margin-bottom-1"
+                        onClick={() =>
+                          history.push(
+                            `/models/${modelId}/task-list/basics/overview`
+                          )
+                        }
```
patrickseguraoddball, line 586: I think for continuity the redirection should happen in the submit handler.  `Back` on milestones is handled differently that in overview.  `Back` should still save the form, so this should mimic the back functionality of overview

src/views/ModelPlan/TaskList/Basics/Milestones/index.tsx
```diff
@@ -0,0 +1,634 @@
...
+              const handleOnchange = (e: any, field: string) => {
+                if (e === '') {
+                  return;
+                }
+                try {
+                  setErrors({});
+                  setFieldValue(field, new Date(e).toISOString());
+                  validateField(field);
+                } catch (error) {
+                  setFieldError(field, 'Please use a valid date format');
```
patrickseguraoddball, line 198: This should use the translation library

patrickseguraoddball, line 198: Honestly I think this should existing in a validation schema, and on not onchange/every key stroke.  The validation should happen once on submit.  I'm able to submit the form with errors if I type something incorrectly, then type something incorrectly in another box, the first validation error gets cleared.  I think autosave might prove to be tricky here, so feel free to leave out autosave on this page due to its complexity with dates.

patrickseguraoddball (changes requested):

patrickseguraoddball (approved):

PR #70, opened 2022-05-05
Easi 1998/archive model plan

patrickseguraoddball:
# EASI-1998

## Changes and Description

- Added ability to archive a model plan
- Added cypress test for archiving model plan
- Refactored some unit tests

## Notes

Passing in queried model plan as prop from task list instead of just the modelId from the url param.  The additional fields of the model are needed for archiving.

## How to test this change

Seed database using MINT (if testing cypress test)
Select/create a model plan and navigate to the task list and select `Remove your Model Plan`, then confirm
Should redirect to home page with model plan archived and no longer returned from query

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/cypress/integration/modelPlan.spec.js
+++ b/cypress/integration/modelPlan.spec.js
@@ -34,4 +34,30 @@ describe('The Model Plan Form', () => {
 
     cy.contains('h1', 'Model Plan task list');
   });
+
+  it('archives a model plan', () => {
+    cy.visit('/models/f11eb129-2c80-4080-9440-439cbe1a286f/task-list');
+
+    cy.wait(1000);
+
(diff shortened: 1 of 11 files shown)


ClayBenson94 (dismissed):
It seems like when I archive a plan, it sets a bunch of fields in the DB to `NULL`. This might be the exact use case for action-specific mutations, because I think the [query that archives the plan](https://github.com/CMSgov/mint-app/pull/70/files?show-viewed-files=true&file-filters%5B%5D=#diff-1e2ada076cee5368c5d960bafd6fb7a22180cd25291db7085de925859f28bb9aR40-R44) doesn't pass all the fields (like name, for example)

garyjzhao (approved):
Archiving worked 🎉

PR #88, opened 2022-05-18
Easi 2038/modified fields nullable

OddTomBrooks:
# EASI-2038

## Changes and Description

- Made ModifiedBy nullable
  - Updated systems accordingly

## PR Author Review Checklist

## PR Reviewer Guidelines


pkg/graph/resolvers/plan_basics.go
```diff
@@ -20,6 +20,7 @@ func UpdatePlanBasics(logger *zap.Logger, id uuid.UUID, changes map[string]inter
 	if err != nil {
 		return nil, err
 	}
+
```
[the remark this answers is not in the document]
OddTomBrooks: Yeah, I don't remember exactly what it's all about but something to do with Linux (Unix?) vs Windows EOF. As I recall the best thing to do is just let it keep the EOF newline. Let's talk about it later today

pkg/storage/model_planStore.go
```diff
@@ -43,17 +43,20 @@ func (s *Store) ModelPlanCreate(logger *zap.Logger, plan *models.ModelPlan) (*mo
...
 			fmt.Sprintf("Failed to create model plan with error %s", err),
-			zap.String("user", models.ValueOrEmpty(plan.ModifiedBy)),
+			zap.String("user", models.ValueOrEmpty(plan.CreatedBy)),
 		)
 		return nil, err
 	}
 	retPlan := models.ModelPlan{}
 
+	plan.ModifiedBy = nil
+	plan.ModifiedDts = nil
```
ClayBenson94, line 53: I wonder if it'd be more clear/reliable to just have the create SQL statement not even take `modifiedBy` and `modifiedDts` as params. That way, even if you somehow passed a `plan` object with those fields, they'd still get created `NULL`.

It's a tradeoff I think - makes it so that you're less flexible in creating objects exactly as you state them, but also means that you'll never accidentally write code that populates that field and sets it. 🤷  I could go either way - thoughts?

ClayBenson94 (dismissed):
One minor comment - can go either way on it so just feel free to dismiss if you disagree 😁

pkg/storage/model_planStore.go
OddTomBrooks: I have been back and forth on this one SO MUCH, glad you brought it up. Truth is I'm just not sure, like you said either way is a pro/con. For now I think let's just go with this and have a call to change it if we want.

ClayBenson94 (approved):

PR #93, opened 2022-05-24
[NOREF] Only start MINT if migrations complete in docker-compose

ClayBenson94:
# No Ticket

## Changes and Description

- Modify the `depends_on` statement in our docker-compose file to only start the service if the migrations complete successfully.

## How to test this change

- Modify a migration to be invalid (bad SQL, same version as another migration, etc.)
- `scripts/dev up`
- Expect an error!

## PR Author Review Checklist

- [x] Tested user-facing changes with voice-over and the [rotor menu](https://support.apple.com/guide/voiceover/with-the-voiceover-rotor-mchlp2719/mac)

## PR Reviewer Guidelines


StevenWadeOddball (approved):
Great Idea, this works when I test it locally. It will be convenient to have the error message in this spot instead of needing to hunt down logs in Docker 
image

PR #95, opened 2022-05-25
[EASI-2051] SQL error handling

ClayBenson94:
# EASI-2051

## Changes and Description

- Removed unneeded error type checking when using SQLX's `Select` (which doesn't actually return an error when no results are found)
- Removed checking of `sql.ErrNoRows` when using SQLX's `Get` (so that we now get an error when querying a single entity and don't find it)

## Open Items
- We should still take some time to figure out how we want to do contextually appropriate error logging, if that's a route we want to take (over just forwarding the errors up the call stack)

## How to test this change

1. `scripts/dev up:backend`
2. Try and get an entity by ID that doesn't exist:
```gql
query getModelPlan {
  modelPlan(id: "eeeeeeee-d693-4daf-87ef-02c97701fc46") {
    id
    basics {
      id
    }
  }
}
```
3. Expect an error

## PR Author Review Checklist

- [x] Tested user-facing changes with voice-over and the [rotor menu](https://support.apple.com/guide/voiceover/with-the-voiceover-rotor-mchlp2719/mac)

## PR Reviewer Guidelines

### DIFF
--- a/pkg/graph/resolvers/plan_collaborator_test.go
+++ b/pkg/graph/resolvers/plan_collaborator_test.go
@@ -85,11 +85,9 @@ func (suite *ResolverSuite) TestDeletePlanCollaborator() {
 	suite.NoError(err)
 	suite.EqualValues(deletedCollaborator, collaborator)
 
-	// Ensure we get nil when we fetch it
-	// TODO: FetchByID methods should probably error if they don't find what they're looking for,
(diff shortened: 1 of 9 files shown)


StevenWadeOddball (approved):
Looks good to me! I tested this locally, running test scripts and trying to retrieve an object that doesn't exist. I do agree that continuing to think about if there are other scenarios that require specific behavior would be good

PR #99, opened 2022-05-31
[EASI-2048] Generically calculate task statuses

ClayBenson94:
# EASI-2048

## Changes and Description

- Added a helper method (and included tests) for generically calculating the "status" of an item based on struct tags

## Known Limitations

- This doesn't work on some fields that it probably should (basically any one that's a `pq.StringArray`). For now, though, I think it gets us close enough.

## How to test this change

The test suite should generally cover these, but you can use GQL mutations to update something simple like `Plan Basics` to have statuses properly calculated.

## PR Author Review Checklist

## PR Reviewer Guidelines


OddTomBrooks (approved):
Clever use of tags!

PR #100, opened 2022-06-01
Task List Section Locking Subscription

OddTomBrooks:
# EASI-1995

## Changes and Description

- Added a PubSub service
- Added GQL subscriptions for task list section locking

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/pkg/graph/resolvers/plan_task_list_section_locks.go
+++ b/pkg/graph/resolvers/plan_task_list_section_locks.go
@@ -1,0 +1,99 @@
+package resolvers
+
+import (
+	"sync"
+
+	"github.com/google/uuid"
+
+	"github.com/cmsgov/mint-app/pkg/graph/model"
+	"github.com/cmsgov/mint-app/pkg/graph/model/subscribers"
+	"github.com/cmsgov/mint-app/pkg/models/pubsubevents"
+	"github.com/cmsgov/mint-app/pkg/shared/pubsub"
+)
+
+var (
+	taskListSectionLocks     map[uuid.UUID]map[string]model.TaskListSectionLockStatus
+	taskListSectionLockMutex sync.Mutex
+)
+
+func init() {
+	taskListSectionLocks = make(map[uuid.UUID]map[string]model.TaskListSectionLockStatus)
+	taskListSectionLockMutex = sync.Mutex{}
+}
+
+// GetTaskListSectionLocks returns the list of locked task list sections. Any sections not included should be considered as unlocked.
+func GetTaskListSectionLocks(modelPlanID uuid.UUID) []*model.TaskListSectionLockStatus {
+	sectionsLockedMap, found := taskListSectionLocks[modelPlanID]
+	if !found {
+		return nil
+	}
+
+	var sectionsLocked []*model.TaskListSectionLockStatus
+	for key := range sectionsLockedMap {
+		status := sectionsLockedMap[key]
+		sectionsLocked = append(sectionsLocked, &status)
+	}
+
+	return sectionsLocked
+}
+
+// SubscribeTaskListSectionLockChanges creates a Subscriber and registers it for the pubsubevents.TaskListSectionLocksChanged event
+func SubscribeTaskListSectionLockChanges(pubsub *pubsub.PubSub, modelPlanID uuid.UUID, principal string, onDisconnect <-chan struct{}) (<-chan *model.TaskListSectionLockStatusChanged, error) {
+	subscriber := subscribers.NewTaskListSectionLockChangedSubscriber(principal)
+
+	pubsub.Subscribe(modelPlanID, pubsubevents.TaskListSectionLocksChanged, subscriber, onDisconnect)
+
+	return subscriber.GetChannel(), nil
+}
+
+// LockTaskListSection will lock the provided task list section on the provided model
+func LockTaskListSection(pubsub *pubsub.PubSub, modelPlanID uuid.UUID, section string, principal string) {
+	if isSectionLocked(modelPlanID, section) {
+		return
+	}
+
+	status := model.TaskListSectionLockStatus{
+		Section:  section,
+		LockedBy: principal,
+	}
+
+	taskListSectionLockMutex.Lock()
+	taskListSectionLocks[modelPlanID][section] = status
+	taskListSectionLockMutex.Unlock()
+
+	pubsub.Publish(modelPlanID, pubsubevents.TaskListSectionLocksChanged, model.TaskListSectionLockStatusChanged{
+		ChangeType: model.ChangeTypeRemoved,
+		LockStatus: &status,
+	})
+}
+
+// UnlockTaskListSection will unlock the provided task list section on the provided model
+//	This method will fail if the provided principal is not the person who locked the task list section
+//	TODO: Allow authorized personnel to unlock task list section which were locked by someone else
+func UnlockTaskListSection(pubsub *pubsub.PubSub, modelPlanID uuid.UUID, section string, principal string) {
+	if !isSectionLocked(modelPlanID, section) {
+		return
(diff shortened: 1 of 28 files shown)


pkg/shared/pubsub/event.go
```diff
@@ -0,0 +1,4 @@
+package pubsub
+
+// Event provides the identifying information necessary for a subscription model
+type Event string
```
StevenWadeOddball, line 4: Maybe EventType would be a clearer name here?

pkg/graph/schema.resolvers.go
```diff
@@ -237,6 +237,27 @@ func (r *mutationResolver) DeleteDiscussionReply(ctx context.Context, id uuid.UU
...
+
+	return true, nil
+}
+
+func (r *mutationResolver) UnlockTaskListSection(ctx context.Context, modelPlanID uuid.UUID, section string) (bool, error) {
+	principal := appcontext.Principal(ctx).ID()
+
+	resolvers.UnlockTaskListSection(r.pubsub, modelPlanID, section, principal)
+
+	return true, nil
```
StevenWadeOddball, line 253: I think this is potentially very confusing to always return true. 

I locked the task as USER MINT, tried to unlock as user COOL. I could see that it didn't unlock, but I got a response that suggested that it did unlock the task list section. I think maybe there should be a different return type for this struct? This also is swallowing the error
image

I think the paradigm for the other resolvers is to have the resolver return the return object and error, instead of building it here in the schema resolvers file.

pkg/graph/schema.graphql
```diff
@@ -606,6 +617,26 @@ updateDiscussionReply(id: UUID!, changes: DiscussionReplyChanges!): DiscussionRe
 
 deleteDiscussionReply(id: UUID!): DiscussionReply!
 @hasRole(role: MINT_BASE_USER)
+
+lockTaskListSection(modelPlanID: UUID!, section: String!): Boolean!
```
StevenWadeOddball, line 621: I think we should have a task list section field here. In your example, we could lock any string we wanted. 

I don't think there is any point to allow users to send a request for a section that doesn't exist.

pkg/graph/resolvers/plan_task_list_section_locks.go
```diff
@@ -0,0 +1,165 @@
...
+		return false
+	}
+
+	_, found = session[section]
+	return found
+}
+
+// GetTaskListSectionLocks is a convenience relay method to call the corresponding method on a resolver implementation
+func GetTaskListSectionLocks(modelPlanID uuid.UUID) []*model.TaskListSectionLockStatus {
+	return NewPlanTaskListSectionLocksResolverImplementation().GetTaskListSectionLocks(modelPlanID)
```
StevenWadeOddball, line 144: Can you speak to why you chose to use the implementation method here? I feel like it might be clearer to just to duplicate making the object instance here. It's nice that this can be called in one line, but I think it would be clearer to me without the function call. (It feels a lot more complex)

StevenWadeOddball (changes requested):
A lot of great stuff here! I think that this subscription system will open a lot of doors for us.  I still need to review more, but I have a little bit of feedback (see the comments below). I wasn't able to test the subscription portion yet, could you provide a bit more data about what the URL you send the web socket init and subscribe methods to?

I would also really like to see an ADR and some sort of documentation for this. From our conversations about subscriptions, I know you have put a lot of excellent thought into why you went with this technology, and considerations  we should have for the future. I think it is important to document it here while it is still fresh. 

Can you also speak to how you envision the task list locking to go? It seems that this really is just a method for the front end to call, but we don't enforce it anywhere on the backend.  I think it could be worth a bit of extra time on this portion. 

Another possible approach could be to notify the front end when a user is on the front end task list section, and fire an event to the backend. The backend can then just note that the front end sent a user editing `___` task list section or left task list section and update accordingly.  This would add control flow logic to the backend, instead of requiring the front end to handle locking / unlocking imperatively.

I also think that we should enforce a task list locked section on the backend, eg handle an event where a user tries to update a section that was locked by someone else. I didn't see any logic that would prevent that currently.

pkg/graph/resolvers/plan_task_list_section_locks.go
OddTomBrooks: There are three main things to address here:

- Why have I made these wrapper functions which simply create and use an object?
-- In order to provide a similar retroactive structure compliant with our other resolvers
- Ok sure, but why create an object? Why not just do the same as we have been doing?
-- Mostly to facilitate the generation of a mock resolver, see pkg/graph/resolvers/mockplantasklistresolver/resolver.go
- What value does the mock resolver provide?
-- As it is, not too much. Currently we can use the mock resolver to satisfy system level testing. However, by building out a system where the server is provided with a mapping of resolvers we greatly improve the flexibility and testability of our servers. That is outside the scope of this PR, having enough side complexity as-is.

TL;DR:
Upon the development of system where GraphQL can access provided resolver instances, these convenience/wrapper functions will be obsolete and should be removed.

OddTomBrooks:
> A lot of great stuff here! I think that this subscription system will open a lot of doors for us. I still need to review more, but I have a little bit of feedback (see the comments below). I wasn't able to test the subscription portion yet, could you provide a bit more data about what the URL you send the web socket init and subscribe methods to?
> 
> I would also really like to see an ADR and some sort of documentation for this. From our conversations about subscriptions, I know you have put a lot of excellent thought into why you went with this technology, and considerations we should have for the future. I think it is important to document it here while it is still fresh.
> 
> Can you also speak to how you envision the task list locking to go? It seems that this really is just a method for the front end to call, but we don't enforce it anywhere on the backend. I think it could be worth a bit of extra time on this portion.
> 
> Another possible approach could be to notify the front end when a user is on the front end task list section, and fire an event to the backend. The backend can then just note that the front end sent a user editing `___` task list section or left task list section and update accordingly. This would add control flow logic to the backend, instead of requiring the front end to handle locking / unlocking imperatively.
> 
> I also think that we should enforce a task list locked section on the backend, eg handle an event where a user tries to update a section that was locked by someone else. I didn't see any logic that would prevent that currently.

**URL to test subscriptions:**
ws://localhost:8085/api/graph/query

**ADR:**
Yep, agreed. I considered the ADR to be part of the Jira ticket but not PR. If we'd rather have an ADR before the PR is merged, no problem.

**Task List / Flow Control Responsibility:**
The API design on this is intended to be simple and straightforward. Except for constraining the sections to an enum rather than raw strings. The reasoning is to expect a degree of design iteration on this feature. It's far from bulletproof, but as long as an admin can go in and clear the locks it shouldn't create blockers for users through MVP.

One of the key things we'll need to properly support task section locking and unlocking in a secure way is to make it so:
- Users cannot lock more than one section at a time
- There will be an OnDisconnect event called when a user disconnects
  - Remove all user section locks
  
As for the server doing the locking on the client's behalf, I'm not sure how practical that is in a stateless environment. I suspect it is the ideal solution in a vacuum, but with everything else in context I'm not sure that our tech stack is well-positioned to make assertions about users page locality. Maybe it's possible to build a middle layer that checks page requests but that's presumptive and non-trivial.

pkg/graph/schema.resolvers.go
OddTomBrooks: Great point. This is updated for Lock and Unlock along with descriptive errors for the cases they can fail. There is also improved feedback from the admin endpoint UnlockAllTaskListSections to return the list of section locks removed.

pkg/graph/schema.graphql
OddTomBrooks: Yep, agreed. Done.

ClayBenson94:
Closing for now, as this was a Spike ticket and helped us understand some new angles/approaches!

PR #108, opened 2022-06-08
[NOREF] Fixing data dump by only dumping data (and not dumping flyway schema)

ClayBenson94:
# No Ticket

## Changes and Description

- Modify `scripts/dev db:dump` to only dump data (with `--data-only`)
- Modify `scripts/dev db:dump` to not dump flyway migration data (with `-T` flag)
  - This is because the flyway table is NOT touched by `scripts/dev db:clean`, and would lead to conflicts when trying to re-insert that data.

## How to test this change

- Run `scripts/dev db:dump` and see what gets output!
- Run `scripts/dev db:seed` multiple times, and see that you get errors (as there's primary key conflicts with the data, this is expected!)
- Run `scripts/dev db:clean && scripts/dev db:seed` and expect no errors.
  - Test to see if the data is actually seeded and works!

## PR Author Review Checklist

## PR Reviewer Guidelines


StevenWadeOddball (approved):
Thanks Clay, this looks good to me. I merged in main, and also excluded the new existing_models table from the dump command, since it's data is provided in a migration.

PR #115, opened 2022-06-13
[EASI-1940] - General Characteristics FE

patrickseguraoddball:
# EASI-1940

## Changes and Description

- Task List section for General Characteristics
- Query for existing models
- Imported MultiSelect component from EASI
- Custom component for AddNote
- Cypress and unit testing

## Notes

There are no validations on the form as none are required as far as I can tell.  This section is completely free form, no dates to validate, etc.  
Ashley's PR on the MultiSelect is still in review.  I wanted to get this PR started, and have implemented the component from EASI with a few adjustments.  I may come back to this pending Ashley's PR doesn't pass review.

## How to test this change

- Create or seed a model plan
- Select `General Characteristics` from the task list
- Try to break it on each page of the form

MultiSelect combobox should be a combination of both current MINT models and existing models imported from external source into db.  Sorted alphabetically.

## PR Author Review Checklist

## PR Reviewer Guidelines


StevenWadeOddball:
I also noticed that the notes don't start expanded if there is already data in that field. When you click add note, you do see the note field, but the data is not shown on page load. 

I'd expect for this to be expanded if there is already content in this field
image
When clicked
image

StevenWadeOddball:
I noticed when is care coordination is changed from yes, to no, the saved How So? field is cleared out.  Could you explain why? Is that so we don't have a no answer with a record showing why care is involved?

StevenWadeOddball:
> > I noticed when is care coordination is changed from yes, to no, the saved How So? field is cleared out. Could you explain why? Is that so we don't have a no answer with a record showing why care is involved?
> 
> I thought about this a bit, and not sure how to handle it. I think UX should weigh in on dependent answers that revert from yes to no, and if that consequent 'yes' information should be maintained.
> 
> I think I did it for the first one, and meant to ask UX.

That makes sense to me, I can see it both ways really. It makes sense not to store the data in the database, but also it feels a little inconvenient if I'm going back and forth on an answer. I think having UX weigh in is a good call. Anyways, this is a valid choice in how to handle it I think

StevenWadeOddball (approved):
Looks good to me! I tested it locally. My above comments are non-blocking

patrickseguraoddball:
> Something I noticed on the multi-select, If have a box selected, then I press enter, it progresses to the next page instead of checking the box. Space bar selects the check though. Do you expect it to work like this? I was expecting the box to be checked when I pressed enter, it could just be an issue with my expectations however.
> 

That's a good catch.  The Mutiselect is actually still in draft on EASI.  I can pass this on to Ashley as a part of her work for review.  Thanks so much Steven!

StevenWadeOddball:
> > Something I noticed on the multi-select, If have a box selected, then I press enter, it progresses to the next page instead of checking the box. Space bar selects the check though. Do you expect it to work like this? I was expecting the box to be checked when I pressed enter, it could just be an issue with my expectations however.
> 
> That's a good catch. The Mutiselect is actually still in draft on EASI. I can pass this on to Ashley as a part of her work for review. Thanks so much Steven!

My pleasure! Thanks Patrick!

PR #116, opened 2022-06-13
implement common function for deserializing pq.string arrays in resolver

StevenWadeOddball:
This change implements common functionality for casting a PQ string array to a specific enum type. At present, you still have to implement the resolver for each pq.StringArray, but it reduces the lines of code considerably. 

Note, credit goes to @DylanSpOddball for implementation of ConvertEnums  helper function. Thanks Dylan!

I also corrected a type in the updatePlanBeneficiaries mutation. No functional change made there.
## How to test this change

`scripts/dev up:backend:debug && scripts/dev db:seed`

Run this GQL script that should include most / all of the properties that are an array of enums
```gql
query ModelPlan{
  
  modelPlan(id:"6e224030-09d5-46f7-ad04-4bb851b36eab"){
    id
    modelName
    modelCategory
    cmsCenters
    cmmiGroups
    generalCharacteristics {
        resemblesExistingModelWhich
        alternativePaymentModelTypes
        keyCharacteristics
        geographiesTargetedTypes
        geographiesTargetedAppliedTo
        agreementTypes
        authorityAllowances
        waiversRequiredTypes
    }
    providersAndParticipants {
        participants
        selectionMethod
        communicationMethod
        participantsIds
        providerAddMethod
        providerLeaveMethod

    }
    beneficiaries {
        beneficiaries
        beneficiarySelectionMethod

    }
    

  }
  
  
}
```

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/pkg/graph/schema.resolvers.go
+++ b/pkg/graph/schema.resolvers.go
@@ -5,6 +5,7 @@ package graph
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/google/uuid"
 
@@ -17,24 +18,12 @@ import (
 )
 
 func (r *modelPlanResolver) CmsCenters(ctx context.Context, obj *models.ModelPlan) ([]models.CMSCenter, error) {
-	// TODO: We should probably have a better way to handle enum arrays
-	var cmsCenters []models.CMSCenter
-
-	for _, item := range obj.CMSCenters {
-		cmsCenters = append(cmsCenters, models.CMSCenter(item))
-	}
-
+	cmsCenters := models.ConvertEnums[models.CMSCenter](obj.CMSCenters)
 	return cmsCenters, nil
 }
 
 func (r *modelPlanResolver) CmmiGroups(ctx context.Context, obj *models.ModelPlan) ([]model.CMMIGroup, error) {
-	// TODO: We should probably have a better way to handle enum arrays
-	var cmmiGroups []model.CMMIGroup
-
-	for _, item := range obj.CMMIGroups {
-		cmmiGroups = append(cmmiGroups, model.CMMIGroup(item))
-	}
-
+	cmmiGroups := models.ConvertEnums[model.CMMIGroup](obj.CMMIGroups)
 	return cmmiGroups, nil
 }
 
@@ -159,10 +148,8 @@ func (r *mutationResolver) UpdatePlanGeneralCharacteristics(ctx context.Context,
 	return resolvers.UpdatePlanGeneralCharacteristics(logger, id, changes, principal, r.store)
 }
 
-func (r *mutationResolver) UpdatePlanBeneficiares(ctx context.Context, id uuid.UUID, changes map[string]interface{}) (*models.PlanBeneficiaries, error) {
-	principal := appcontext.Principal(ctx).ID()
-	logger := appcontext.ZLogger(ctx)
-	return resolvers.PlanBeneficiariesUpdate(logger, id, changes, principal, r.store)
+func (r *mutationResolver) UpdatePlanBeneficiaries(ctx context.Context, id uuid.UUID, changes map[string]interface{}) (*models.PlanBeneficiaries, error) {
+	panic(fmt.Errorf("not implemented"))
 }
 
 func (r *mutationResolver) UpdatePlanParticipantsAndProviders(ctx context.Context, id uuid.UUID, changes map[string]interface{}) (*models.PlanParticipantsAndProviders, error) {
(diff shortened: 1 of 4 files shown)


pkg/graph/schema.resolvers.go
```diff
@@ -537,3 +451,15 @@ type planGeneralCharacteristicsResolver struct{ *Resolver }
...
+// The code below was going to be deleted when updating resolvers. It has been copied here so you have
+// one last chance to move it out of harms way if you want. There are two reasons this happens:
+//  - When renaming or deleting a resolver the old code will be put in here. You can safely delete
+//    it when you're done.
+//  - You have helper methods in this file. Move them out to keep these resolver files clean.
+func (r *mutationResolver) UpdatePlanBeneficiares(ctx context.Context, id uuid.UUID, changes map[string]interface{}) (*models.PlanBeneficiaries, error) {
+	principal := appcontext.Principal(ctx).ID()
+	logger := appcontext.ZLogger(ctx)
+	return resolvers.PlanBeneficiariesUpdate(logger, id, changes, principal, r.store)
+}
```
ClayBenson94, line 465: Can you move this code into the newly generated `UpdatePlanBeneficiaries` function and delete this block?

pkg/models/model_helpers.go
```diff
@@ -22,3 +24,12 @@ func ValueOrEmpty(st *string) string {
...
 }
+
+//ConvertEnums converts a pq.StringArray to specific, castable type
+func ConvertEnums[EnumType ~string](pqGroups pq.StringArray) []EnumType {
+	enumValues := []EnumType{}
+	for _, item := range pqGroups {
+		enumValues = append(enumValues, EnumType(item))
+	}
+	return enumValues
+}
```
ClayBenson94, line 35: I spent some time playing with this function in the Go playground. I really wanted to find a way to validate that the enum array generated is valid (as auto generated enums do provide a `.IsValid` function that can be used to check if the value is in the allowed list.

I couldn't figure out a way to get this to work from a generics perspective (tried to make the type an Interface, but then had trouble casting), so I just ended up giving up. Not sure if you thought about this at all, but it was an interesting learning experience for me in Go generics and slice types!

For what it's worth, this is probably a moot point because the GQLGen generated resolver code (under the hood) calls `IsValid` before returning the GQL response, and we also have these types strictly bound in our DB, so I can't practically think of a situation where this would actually happen.

ClayBenson94 (changes requested):

pkg/models/model_helpers.go
DylanSpOddball: > auto generated enums do provide a .IsValid function that can be used to check if the value is in the allowed list.

That's interesting, I didn't know of that; are you referring to [this function](https://pkg.go.dev/reflect#Value.IsValid) from the `reflect` package in the stdlib, or something else?

pkg/graph/schema.resolvers.go
```diff
@@ -17,24 +18,12 @@ import (
...
 
 func (r *modelPlanResolver) CmsCenters(ctx context.Context, obj *models.ModelPlan) ([]models.CMSCenter, error) {
-	// TODO: We should probably have a better way to handle enum arrays
-	var cmsCenters []models.CMSCenter
-
-	for _, item := range obj.CMSCenters {
-		cmsCenters = append(cmsCenters, models.CMSCenter(item))
-	}
-
+	cmsCenters := models.ConvertEnums[models.CMSCenter](obj.CMSCenters)
```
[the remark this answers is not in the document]
StevenWadeOddball: I thought about it a bit, but I think it would be better to handle in another PR. I'm not envisioning this method as the final solution for enum arrays, but a step in the right direction for now. I think this model is one of the few that is in an an array and exists in the `models` package vs the  auto generated `model` package. 
 
 Ideally, I'd like to tackle the final approach in [EASI-1982](https://jiraent.cms.gov/browse/EASI-1982) . I started to do some work towards that here, but I think it would make sense to do another take on that ticket, and if it isn't needed still, remove them and rely on auto-generated. I think the main reason is to not have any hand rolled types reference any auto-generated model types.

pkg/models/model_helpers.go
ClayBenson94: @DylanSpOddball https://github.com/CMSgov/mint-app/blob/b8d4482de3b70aab3a1a1e85663be5c7c5028496/pkg/graph/model/models_gen.go#L82-L123

DylanSpOddball: Oh, I didn't catch that you were referring to generated code from GQL enums, that makes more sense. Thanks.

ClayBenson94 (approved):

PR #117, opened 2022-06-14
[EASI-2025] Remove NULLIF in SQL statements

ClayBenson94:
# EASI-2025

## Changes and Description

This PR removes some unneeded empty string handling in our SQL, as it's practically not needed anymore. Technically, a struct with `""` could still manually be created and inserted into the database, but practically (via GQL update statements) it cannot happen, as the following block of code (which handles updates for our models) handles this case already: https://github.com/CMSgov/mint-app/blob/16add5f857920af6bb2d64c74722552f30f7506c/pkg/graph/resolvers/apply_changes.go#L20-L29

There's some more research we could do in the long term around using `zero.String` in our models to entirely prevent this, but I think the AC of this ticket are satisfied by this change.

- Remove `NULLIF()` statements in SQL commands for creating/updating tables.

## How to test this change

- `scripts/dev up:backend` to start the service
- `scripts/dev test:go` to run the backend test suite
- `scripts/dev db:seed` to seed the DB

```graphql
mutation updatePlanBasics {
  updatePlanBasics(id:"4132e334-4a10-4e8e-9ef1-a72fe910dc82", changes:{
#    goal:""
  }) {
    id
    modelPlanID
    modelType
    problem
    goal
    testInventions
    note
    createdBy
    createdDts
    modifiedBy
    modifiedDts
    status
  }
}
```
See that `goal` is set to a valid string.

```graphql
mutation updatePlanBasics {
  updatePlanBasics(id:"4132e334-4a10-4e8e-9ef1-a72fe910dc82", changes:{
    goal:""
  }) {
    id
    modelPlanID
    modelType
    problem
    goal
    testInventions
    note
    createdBy
    createdDts
    modifiedBy
    modifiedDts
    status
  }
}
```

See that `goal` is now set to `null`, even though we passed an empty string.

## PR Author Review Checklist

## PR Reviewer Guidelines


migrations/V2__Add_Domains.sql
```diff
@@ -2,3 +2,5 @@ CREATE DOMAIN EUA_ID AS TEXT
 CONSTRAINT CHECK_VALID_EUA_ID CHECK (
     VALUE ~ '^[A-Z0-9]{4}$'
 );
+
+CREATE DOMAIN ZERO_STRING AS TEXT CHECK (length(value) > 0);
```
StevenWadeOddball, line 6: To point it out,  right now, this checks for '',  but it doesn't check for ' ' or '  '. This is ok, but we aren't handling situations where users are just inserting spaces.

StevenWadeOddball (approved):
This looks good to me! I think that this is a good way to ensure we don't have empty strings instead of null in the database. 

I verified the domain worked correctly 

```sql
update model_plan
SET cms_other = ''

WHERE id = 'ce3405a0-3399-4e3a-88d7-3cfc613d2905'
```
returned `ERROR:  value for domain zero_string violates check constraint "zero_string_check"
SQL state: 23514`


I do like your idea of checking for extra blank spaces in the string coming from GQL as an extra data sanitization step. That change would likely be more appropriate in a separate PR.

migrations/V2__Add_Domains.sql
ClayBenson94: I think this is fine, since the zero value for a string is `""` and handling that case alone I think is sufficient for now 👍

PR #123, opened 2022-06-21
Easi 1982/generic type array serialization

StevenWadeOddball:
# EASI-1982

## Changes and Description

- Proof of concept generic scan and value functions
- Place holder enumArray GQL directive

This branch is a proof of concept for how we can have explicit types on GO models, and also share generic functionality to scan and value.  This functionality was largely adopted from PQ.stringArray scan and value functions,  but adapted to use GO 1.18 generics.
- [GenericScan](https://github.com/CMSgov/mint-app/pull/123/files#diff-5767031ae723e7829be9ca1e85a874abed8b30d61f21d920c15fecb41e119182R40-R58)
- [GenericValue](https://github.com/CMSgov/mint-app/pull/123/files#diff-5767031ae723e7829be9ca1e85a874abed8b30d61f21d920c15fecb41e119182R60-R83)

GO does not allow UnNamed types to have a receiver function defined for it. To use the scan and value functions, you need to make a type alias for the typed array, and then implement the Scan and Value interface.
- [CMSCentersG](https://github.com/CMSgov/mint-app/pull/123/files#diff-4d084287bb8a7e16fba216370cecb9a2089414aa0704977c440d4837b35be904R52-R60)
- [CMMIGroupG](https://github.com/CMSgov/mint-app/pull/123/files#diff-4d084287bb8a7e16fba216370cecb9a2089414aa0704977c440d4837b35be904R62-R70)

Unfortunately, you can not use a type alias in GraphQL, as such you still need to implement the resolver
 - [CMMIGroups](https://github.com/CMSgov/mint-app/pull/123/files#diff-67372475881f223c5f20bf1514c5035f46db7372effcf91c6fc249c55c4f24e1R25-R29)

It looks like the way to get around this would be to define a plugin for GQL to handle these cases explicitly. You can see some introductory work done towards this effort [here](https://github.com/CMSgov/mint-app/pull/123/files#diff-166a0d682d447944991a09dab5e4b52919964900e077025895db9931ec364c99R60-R67), but more would need to be done. These GQL config changes are unnecessary at the current implementation level, but I left them to show possibilities for development in that direction. Note specifically, the skip_runtime decorator specifies that this is used in the code generation phase.

I started a plugin for GQL gen to see if we could bypass the requirement. It seems like a fairly big lift, and we would also have to support the plug in the future. I think it would be reasonable to simply implement the resolvers for the unnamed types, to type aliases. You can run the GQL endpoint with the example plugin injected by running this command `go run cmd/gqlgen/main.go` The example is largely taken from gqlgens documentation [here](https://gqlgen.com/reference/plugins/#using-a-plugin).  The main example file is [here](https://github.com/99designs/gqlgen/tree/master/plugin/modelgen)
If we want to go forward with this, we would need to either A. Remove this plugin, B. fully implement it.

Another thing to note, when you have the type defined in the go object, you need to explicitly define the enum in GO, GQL, and GRAPHQL. You can't use the autogenerated types from the model package, because it is a circular dependency.  

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/cmd/gqlgen/main.go
+++ b/cmd/gqlgen/main.go
@@ -1,0 +1,29 @@
+//go:build ignore
+// +build ignore
+
+package main
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/99designs/gqlgen/api"
+	"github.com/99designs/gqlgen/codegen/config"
+	enumArray "github.com/cmsgov/mint-app/cmd/gqlgen/plugin"
+)
+
+func main() {
+	cfg, err := config.LoadConfigFromDefaultLocations()
+	if err != nil {
+		fmt.Fprintln(os.Stderr, "failed to load config", err.Error())
+		os.Exit(2)
+	}
+
+	err = api.Generate(cfg,
+		api.AddPlugin(enumArray.New()),
+	)
+	if err != nil {
+		fmt.Fprintln(os.Stderr, err.Error())
+		os.Exit(3)
+	}
+}
--- a/cmd/gqlgen/plugin/enum_array_plugin.go
+++ b/cmd/gqlgen/plugin/enum_array_plugin.go
@@ -1,0 +1,130 @@
+package enumArray
+
+import (
+	// "fmt"
+	"fmt"
+	"go/types"
+
+	// "sort"
+	"strings"
+
+	// "github.com/99designs/gqlgen/codegen/config"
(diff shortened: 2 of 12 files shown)


StevenWadeOddball:
Closing pull request in favor of implmentation https://github.com/CMSgov/mint-app/pull/134

PR #125, opened 2022-06-22
PubSub Service

OddTomBrooks:
# NOREF

## Changes and Description

- Introduced a pubsub utility so we can manage internal event dispatching pending a more comprehensive solution

## PR Author Review Checklist

## PR Reviewer Guidelines


ClayBenson94:
Changes themselves look solid - though I think something might be going on with the `mock` and `golang-ci-lint` 🤔 

https://github.com/CMSgov/mint-app/runs/6998844186?check_suite_focus=true#step:10:96

OddTomBrooks:
> > Also thinking that we can try and merge this PubSub client, then when we go to implement task list locking directly, we can try and use this _and_ an open source pubsub client, and see which implementation we prefer!
> > Thoughts, @OddTomBrooks & @StevenWadeOddball?
> 
> I think that sounds like a good approach! I think the content looks good. @OddTomBrooks, I noticed you have all the different types in individual files. It seems like a lot of them are are just a type alias, eg session_map and subscriber_map, without other additional logic. Did you consider putting them in a single file?
> 
> Was also wondering if you considered returning a struct with fields for the function [setupPubSubTest](https://github.com/CMSgov/mint-app/pull/125/files#diff-a172e516c638eb4be7964532dd9b0f1d4191c4c1549f9cc2c1fd1903ae9c2c4bR23-R33) ?
> 
> Those questions aren't blockers, I'm just curious.

It's commonly established clean coding practices to prefer more small files with single responsibility than few large files with multiple responsibility. If the team prefers merging the files I don't have a big problem with it, but I do feel it makes the code slightly less clean.

StevenWadeOddball:
Thanks Tom! As is is good with me! I think it's just another way to do it, so I wanted to learn why you chose to do it this way.

pkg/shared/pubsub/testpubsub/pubsub_implementation_test.go
```diff
@@ -0,0 +1,33 @@
...
+	ps, subscriber, eventType, payload, modelPlanID, _, disconnectChannel := setupPubSubTest(t)
+
+	subscriber.EXPECT().GetID().Return("MOCK_SUBSCRIBER").Times(2)
+	subscriber.EXPECT().Notify(payload)
+
+	ps.Subscribe(modelPlanID, eventType, subscriber, disconnectChannel)
+	ps.Publish(modelPlanID, eventType, payload)
+}
+
+func setupPubSubTest(t *testing.T) (*pubsub.ServicePubSub, *mockpubsub.MockSubscriber, pubsub.EventType, string, uuid.UUID, string, chan struct{}) {
```
ClayBenson94, line 23: For what it's worth (and I know there's been some discussion around `setupPubSubTest()` here already), I've taken an approach of using Testify to setup tests (and using their [SetupTest()](https://pkg.go.dev/github.com/stretchr/testify/suite?utm_source=godoc#SetupTestSuite) function - [example in MINT](https://github.com/CMSgov/mint-app/blob/3eb0efae59847b8776b413b42fe9945f84b60622/pkg/graph/resolvers/resolver_test.go#L20-L23))

This lets you tack on items to the `suite` object itself, making them globally available to each test, which is especially nice if you have test-specific dependencies (like a pubsub instance, subscriber instance, etc.) that you want populated for each test.

Might be overkill here, since it's just the 1 test using it, but I think it could be nice if we started doing this across our tests!

ClayBenson94 (changes requested):
(Don't feel like you have to make the change I suggested here, but just throwing it out!)

pkg/shared/pubsub/testpubsub/pubsub_implementation_test.go
OddTomBrooks: Took a stab, will implement in a second pass

ClayBenson94 (dismissed):
Approving - looks like there's just a `go.mod` conflict. Should probs be resolved via the CLI rather than manually editing `go.mod`, but otherwise we're good! Good job on this!!

OddTomBrooks:
Interesting. I tried go mod tidy as a pass to fix this but had no luck.
Let’s have a call about it.

ClayBenson94 (approved):

PR #138, opened 2022-07-06
Easi 2150 - tech debt gql refactor

patrickseguraoddball:
# EASI-2150

## Changes and Description

- Refactored GQL queries/mutations in to folders based on task list
- Implemented stronger type safety by removing type assertions and implementing page specific queries
- Added simple error handling for each page (Should revisit later)
- Updated react-autosuggest
- Fixed error and validation for an uncontrolled input on Basics 
- Fixed warning for Trans component, replaced with a simpler hook
- Created a cypress run order (ex:  archive should be tested last if we are using same model)

## Notes

- The previous fix to the Trans component was throwing warnings.  The use of the component rather than the hook was unnecessary in the instances related to modelName.  Replaced with the hook.
- React-autosuggest had a deprecated dependency that was throwing warnings in RTL tests
- Made an effort to remove as much type assertions as necessary, resulting in individualized queries/per page queries.

I know a lot of files were touched in this PR.  The scope of the cleanup/reorg was more expansive than initially thought.

## How to test this change

Run all unit and cypress tests
Create a model plan, and run through task list items and actions

## PR Author Review Checklist

## PR Reviewer Guidelines


cypress.json
```diff
@@ -1,5 +1,16 @@
...
+  "testFiles": [
+    "login.spec.js",
+    "modelPlan.spec.js",
+    "collaborator.spec.js",
+    "discussions.spec.js",
+    "documents.spec.js",
+    "characteristics.spec.js",
+    "participantsAndProviders.spec.js",
+    "opsEvalAndLearning.spec.js",
+    "archive.spec.js"
```
ClayBenson94, line 14: Does this achieve what we're looking for? I ask because we're running `scripts/dev db:seed` on each test file, so we _should_ have a fresh database for each test!

https://github.com/CMSgov/mint-app/blob/0b35902a1f707477716706b5d7bb68428b26b4eb/cypress/support/global-hooks.js#L1-L3

cypress/integration/characteristics.spec.js
```diff
@@ -84,6 +87,9 @@ describe('The Model Plan General Characteristics Form', () => {
     cy.contains('button', 'Next').click();
 
     // Page - /characteristics/involvements
+
+    cy.wait(500);
```
ClayBenson94, line 91: Was this test failing without the `wait()`s? I thought Cypress was usually pretty good about waiting for selectors to appear, and 500ms seems like it'd fall in the range of the default timeout (4 seconds, [src](https://docs.cypress.io/guides/references/configuration#Timeouts))

ClayBenson94 (changes requested):

cypress/integration/characteristics.spec.js
patrickseguraoddball: Yeah the tests were failing on each page after the GQL refactor, which I assumed could possibly be because those queries were no longer cached on the first page.  I need to look into why it was failing to find the selectors without the wait.

cypress.json
patrickseguraoddball: Hmmm.  The two test files following the spec where model plan was archived - they were consistently failing to find the 'Empty Plan'.  I'll look into this further as well as if it is indeed seeding on each test file

ClayBenson94: https://github.com/CMSgov/mint-app/blob/1349485a8367358ba15469f458245c32ed6341e9/scripts/data/local-db.sql#L24

Hmm - it does appear to be in the data dump 🤔 perhaps because `db:seed` isn't wiping the database, any changes to that model plan (by ID) will not properly get reflected?

@patrickseguraoddball maybe try adding `cy.exec('scripts/dev db:clean')` to that `global-hooks` (before the seed) as well?

cypress/integration/characteristics.spec.js
patrickseguraoddball: Going to work on troubleshooting cypress next sprint 

https://jiraent.cms.gov/browse/EASI-2166

patrickseguraoddball:
Added the cy.wait back in and will be addressed in another ticket

garyjzhao (approved):

PR #142, opened 2022-07-12
Add ADR for data exporting approach

downeyn-cms:
# EASI-2061

## Changes and Description

- Add ADR for data exporting approach/spike

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/docs/adr/0002-adr-data-export.md
+++ b/docs/adr/0002-adr-data-export.md
@@ -1,0 +1,43 @@
+# Data Exporting Approach
+
+[EASI - 2061](https://jiraent.cms.gov/browse/EASI-2061)
+
+MINT has a need to support the exporting of data from the application for reporting and other use cases.
+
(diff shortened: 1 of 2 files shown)


StevenWadeOddball:
This seems like it could be a good approach to me! As you noted, the use cases of exporting all the data, vs one model might complicate this a bit. Currently, every task list section you query results in it's own SQL Query.

 If we try to export lots of models all at once we might want to consider approaches to maximize performance. Utilizing [dataloaders](https://gqlgen.com/reference/dataloaders/) was suggested in the past as a way of optimizing performance and avoiding an N+1 problem.  This might buy us the performance that a custom SQL query would provide, as well as being flexible.

downeyn-cms:
Thank you for input Steven, I think i'll add the point about improving data loaders into the "Follow Up Decisions and Actions" section.

StevenWadeOddball (approved):
This sounds like a good direction to go, thanks Nick!

PR #154, opened 2022-07-21
Easi 2156/db seeding audit 

StevenWadeOddball:
# EASI-2156

## Changes and Description

- Minor dev script enhancements for current db-seeding and db-dumping paradigm
    - This will likely go away with future refactoring around db:seeding
    -  This allows you to pass a file name to the db seed command eg `scripts/dev db:seed[filename]`
         - EX scripts/dev db:seed[pmButler] 
    -  This allows you to pass a file name to the db dump command eg `scripts/dev db:dump[filename]`
         - EX scripts/dev db:dump[pmButler] 
- ADR for database seeding paradigm
    -  This will inform a future user story to refactor db seeding to use the agreed upon method.
    -   The future ticket is [EASI-2187](https://jiraent.cms.gov/browse/EASI-2187)
   

## PR Author Review Checklist

## PR Reviewer Guidelines


scripts/dev
```diff
@@ -342,14 +342,33 @@ namespace :db do
...
-    # Seed DB
-    `psql < scripts/data/local-db.sql > /dev/null`
+  task :seed, [:file] do |t, args|
+    file = args.to_a[0]
+
+    if file == nil
+      seed()    
+    else
+      if file.start_with?("/")
+      else
```
ClayBenson94, line 367: What does this `if` with an immediate `else` do?

Also, I assume the `file = ` line falls under the `else`? (If so, indentation might need a touch up to make it clear)

scripts/dev
```diff
@@ -342,14 +342,33 @@ namespace :db do
   end
 
   desc "Dump DB to file"
-  task :dump do
-    sh "pg_dump -T *.*flyway* -T *.*existing_model --data-only postgres > scripts/data/local-db.sql"
+  task :dump, [:file] do |t, args|
```
ClayBenson94, line 345: These changes could use [documentation here](https://github.com/CMSgov/mint-app/blob/0b35902a1f707477716706b5d7bb68428b26b4eb/docs/dev_script_usage.md#L16-L17)

Ruby does allow stuff like `scripts/dev --tasks` to get some "self documenting" things, but this doc covers a bit more of the uses for folks unfamiliar with the syntax (like myself a few minutes ago) - I struggled to figure out how to actually pass the argument to this dump/seed command for a bit 😓

docs/adr/0004-adr-db-seeding.md
```diff
@@ -0,0 +1,66 @@
...
+## Considered Alternatives
+
+* Direct interaction with the database store in Go
+* Continue using db dump file, but let migrations update the dump file
+* Continue using db dump file, but write a Go program to update the seed file programmatically
+* Implement a Go seed program that utilizes the schema resolver methods
+
+## Decision Outcome
+
+* Chosen Alternative: **Implement a Go seed program that utilizes the schema resolver methods**
```
ClayBenson94, line 18: Can you link the ticket that was created to implement this in this PR? Just in the comments is fine (or in this section of the ADR, probably wouldn't hurt for visibility!)

ClayBenson94 (changes requested):

scripts/dev
StevenWadeOddball: Good call about the indentation! I added it. 

The if immediate else was meant to ignore the situation where someone passes an explicit path to the function (something like a temp demo file you keep on your desktop, but isn't needed for the repo eg `/Users/cloak/Desktop/local-db.sql`) .

 But looking at it, I think it is a poor way to structure that logic.  I'm changing it to this, which I think is clearer
 
 ```ruby
 
      if !file.start_with?("/") #files that start with / are fully qualified path names
           file = "scripts/data/"+file+".sql"
      end
 ```

scripts/dev
StevenWadeOddball: that's a good shout out, thanks Clay! I'll add to the documentation file. It is the only task I know of that currently takes arguments, so it definitely is out of the ordinary here.

docs/adr/0004-adr-db-seeding.md
StevenWadeOddball: I started adding in the main message yesterday, but I definitely didn't add any description haha, that would be helpful. I think you are right, it would be clear to have it in the actual ADR as well, so I will add there too.

The future ticket is [EASI-2187](https://jiraent.cms.gov/browse/EASI-2187)

ClayBenson94 (approved):

PR #158, opened 2022-07-22
Easi 2134/task list last updated

patrickseguraoddball:
# EASI-2134

## Changes and Description

- Added `Last Updated` to task list sections
- modifiedDts field should be rendered mm/dd/yyyy for all statuses that are not `READY`
- `Basics` is the only outlier as it has 3 different statuses and 3 different modifiedDts
- Reworked how task list renders statuses to better accommodate displaying relevant task list info

## Notes

- Some seeded data won't accurately display the realistic nature of last modified dates, as there are status if `IN_PROGRESS` with no modifiedDts, which would never occur
- There are some fields in the forms that won't trigger `IN_PROGRESS` status change, which is dependent on the `Last Modified` rendering.  *may need to consult with BE if we want to change or audit this*

## How to test this change

- Create a model plan and fill out fields
- Navigate back to task list and see status and `Last Modified` rendered

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/src/i18n/en-US/modelPlanTaskList.ts
+++ b/src/i18n/en-US/modelPlanTaskList.ts
@@ -34,37 +34,44 @@ const modelPlanTaskList = {
     basics: {
       heading: 'Model basics',
       copy:
-        'Start filling out as much of the basic model information as you know and reach out to the Model Assessment Team if you need help.'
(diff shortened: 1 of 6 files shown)


garyjzhao (approved):
This looks great!

PR #173, opened 2022-08-08
[NOREF] Archive Model Plan Query

garyjzhao:
# NOREF

## Changes and Description

- This PR updates the query called in attempting to archive a Model Plan

## How to test this change

1. Create new model plan
2. Click "Remove your Model Plan" 
3. Ensure model plan is removed and no errors
## PR Author Review Checklist

## PR Reviewer Guidelines


src/queries/ArchiveModelPlan.ts
```diff
@@ -0,0 +1,9 @@
+import { gql } from '@apollo/client';
+
+export default gql`
+  mutation ArchiveModelPlan($id: UUID!, $changes: ModelPlanChanges!) {
```
ClayBenson94, line 4: Would it make sense to have the parameter for this just be a `Bool!`? Right now you could technically call `ArchiveModelPlan` with a whole bunch of changes - seems like the naming might be a bit confusing

Either that, or just rename the query to `UpdateModelPlan` and the other one (that currently has that name) to something like `UpdateModelPlanAndBasics`?

ClayBenson94 (changes requested):

ClayBenson94 (approved):

PR #174, opened 2022-08-09
[NOREF] Update pre-commit hooks

ClayBenson94:
# No Ticket

## Changes and Description

- This is almost an exact clone of https://github.com/CMSgov/easi-app/pull/1727

## How to test this change

- Pull this branch down
- Make changes that you think should break the linting process
- Run pre-commit run --all-files

## PR Author Review Checklist

## PR Reviewer Guidelines


docs/adr/index.md
```diff
@@ -1,14 +0,0 @@
-# Architectural Decision Log
```
StevenWadeOddball, line 1: Did a pre-commit hook delete this? We probably want this file added back again

StevenWadeOddball (changes requested):
Functionally this seems good to me! I wasn't able to break the pre-commit. The only thing I think we'd need is to re-add the index.md for the adrs.

docs/adr/index.md
ClayBenson94: I mentioned this in the original [EASi PR](https://github.com/CMSgov/easi-app/pull/1727#issuecomment-1209610374), but I can echo the main points here:

- The git hook to actually update this automatically (turns out it was supposed to be automatic) was broken.
- I am not sure the index was actually that useful, since it kinda just echoed the filesystem, and the files are named clearly enough as-is

So I ended up just deleting it on purpose 😂

StevenWadeOddball: thanks Clay! Makes sense, I think what you did for EASI, linking directly the folder directory instead of the index will work great :).

StevenWadeOddball (approved):
Looks good to me! thanks Clay!

PR #175, opened 2022-08-10
[NOREF] Rename job codes and related auth/principal function/property names

ClayBenson94:
# NOREF

## Changes and Description

- Rename hard-coded job codes to better reflect what we'll create with the IDM team
- Rename GQL roles to match job code names (this doesn't have to happen, but is somewhat nice IMO)

## How to test this change

1. Use `scripts/dev up` to start up MINT.
2. Login (local auth or the EZOD account, for that matter).
3. Navigate to http://localhost:8085/user-diagnostics to see information about the user.
4. Try different GQL queries/mutations that require different roles.

## What's left to do / open questions
- Should the frontend allow access to Secure routes if `isBasicUser` is `false`? It seems there were some assumptions around "successful Okta auth" being the same as "you're a basic user", but the presence of a job code would make me think that's not a good thing to assume - CC @patrickseguraoddball @StevenWadeOddball 

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/docs/graphql_playground.md
+++ b/docs/graphql_playground.md
@@ -17,16 +17,16 @@ The GraphQL endpoints require a valid `Authorization` header to use, which can b
 Job codes can be included in an array for querying endpoints such as `systemIntake` that require them:
 
 ```
-{ "Authorization":"Local {\"favorLocalAuth\":true, \"jobCodes\":[\"MINT_P_USER\"]}"}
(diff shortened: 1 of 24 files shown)


StevenWadeOddball (approved):
Looks good to me! I ran the tests locally and also signed in with local auth. I checked the new paradigm by only providing the assessment job code without base user. It didn't allow the route as expected. Users explicitly need the user role.
image

PR #176, opened 2022-08-12
Easi 1980/collaborator access control

StevenWadeOddball:
# EASI-1980

## Changes and Description

This change focuses on disallowing making changes to items that are part of model plan that you aren't a collaborator on. Assessment users are excluded from this and are able to make changes to every item.  Normal users that try to update sections that they aren't a collaborator on will get an error message that says they don't have permission to do that.

- Resolvers pass principal struct instead of string.
- Add SQL to check if user is a collaborator by ModelPlanID
- Add SQL to check if user is a collaborator by PlanDiscussionIOD
- Add Check in pre-update for base -struct
    - Manual addition to structs not able to use base struct update currently
        - Plan Document
        - Plan Collaborator
        - Plan Discussion Reply 
- Add interface for model plan relations to enforce permissions before each update
- Add interface for model plan relations to enforce permissions before delete and create for relations that are manually created or deleted
    - EG, collaborator, document, discussion, and discussion reply.
- Add Embedded struct for most use cases that implements the model plan relation
    - Manually implement the functionality for structs that can't use the embedded struct ( EG Model Plan)
  

## How to test this change

1. Run the app
2. Try to update data on any form section where you aren't a collaborator
   a. I suggest using the front end UI. You can  navigate by URL to a model plan you don't have permission to edit. Try to edit, delete, and add to that model plan
   b. I suggest repeating the activity as a user with assessment as well as the basic user role. Note that you can take actions on all model plans. 

##Other things to review
1. Do you agree with the error logging? 
2. Do you agree with the interface naming and uses?
3. Is this in the most useful format for the front end?

## PR Author Review Checklist

## PR Reviewer Guidelines


ClayBenson94 (changes requested):
Looks like there's a few changes that are improperly getting re-introduced from `main`.

Changes to `.pre-commit-config.yaml`, `README.md`, stuff in the `docs/adr/` folder, `package.json`, `yarn.lock`, `scripts/`, and most (all?) of the changes in `src/` all seem to be things we can revert.

pkg/models/model_plan_relation.go
```diff
@@ -0,0 +1,28 @@
...
+
+//CheckAccess fulfills the check accessinterface
+func (m ModelPlanRelation) CheckAccess() {
+
+}
+
+//ICheckAccess is an interface meant to demonstrate that this model should have an access check
+type ICheckAccess interface {
+	CheckAccess()
+}
```
ClayBenson94, line 28: I can't seem to find an implementation of `CheckAccess()` on any of our models. Is the presence of this interface acting as a "flag" to set on models that should have access control checks implemented for them?

pkg/graph/resolvers/base_struct.go
```diff
@@ -1,20 +1,46 @@
...
+	"go.uber.org/zap"
+
+	"github.com/cmsgov/mint-app/pkg/authentication"
 	"github.com/cmsgov/mint-app/pkg/models"
+	"github.com/cmsgov/mint-app/pkg/storage"
 )
 
 // BaseStructPreUpdate applies incoming changes from to a TaskList Section, and validates it's status
-func BaseStructPreUpdate(bs models.IBaseStruct, changes map[string]interface{}, principal string) error {
+func BaseStructPreUpdate(logger *zap.Logger, bs models.IBaseStruct, changes map[string]interface{}, principal authentication.Principal, store *storage.Store, applyChanges bool) error {
```
ClayBenson94, line 12: Is there a reason this would be on `BaseStructPreUpdate` and take an `IBaseStruct`? I would assume that we'd want to pass something that specifically has a ModelPlanID 🤔

ClayBenson94 (changes requested):

pkg/models/model_plan_relation.go
StevenWadeOddball: That's correct, I'm essentially using it like a flag. I'm not strongly attached to the implementation, it's definitely not the main use for interfaces. 

Another possible implementation for the same effect would be to add the CheckAccess() method to IBaseStruct which would return a bool, and we could override the function on classes that need the access check, eg BaseTaskListSection etc

ClayBenson94: Yeah - that would mean the function would probably need to take a `store`, but conceptually that might be a bit cleaner, because then you would just be able to call `CheckAccess` in the access control file, rather than calling the store method directly after doing type assertion

pkg/graph/resolvers/base_struct.go
StevenWadeOddball: The reason I put the check there is because all objects should be going through BaseStructPreUpdate, since every object should be a BaseStruct. 

There are edge cases where you want to check something that doesn't have a ModelPlanID directly, so you need to handle it by other parameters (eg discussion id). I tried an implementation where I just wrote that logic directly in the separate resolver, but I didn't like it as much as this. A structure like this means that as long as it is a base struct, access will be checked.

```mermaid
flowchart LR
a1(PreUpdate)
a1--> a2(Check If Struct Requires Access Check)
a2 -->a3(Yes) 
a3-->a10(AssessmentUser?)
a10--True-->a5
a10-->a11(False)
a11--ModelPlanRelation--> a6(CheckByModelPlanID)
a11--DiscussionRelation-->a7(CheckByDiscussionID)
a6-->a8(Collaborator?)
a7-->a8
a8--True-->a5
a8--false-->a9(ReturnError)
a2 --> a4(No)
a4-->a5(ContinueWithAction)


```

pkg/models/model_plan_relation.go
StevenWadeOddball: Hmm, I wasn't thinking about it like that. I was just meaning, you'd implement the function to say that the struct requires an access check. So that would mean instead of seeing if ICheckAccess is implemented, we check true or false from the CheckAccess method.

patrickseguraoddball:
Nice work Steven!  Everything works great as a MINT_USER - I can edit if I am a collaborator on a model plan, and not authorized to mutate any data on a plan that I am not a part of.  

When logged in as MINT_ASSESSMENT I can not edit any part of the model plan (either as a collaborator or not).  This is true for discussions, documents, collaborators, and model plan.

As assessment team member:
Screen Shot 2022-08-19 at 9 25 31 AM

`{"errors":[{"message":"not authorized","path":["updatePlanGeneralCharacteristics"]}],"data":null}`

The error structure, naming, and messages all make sense and are easy to ingest from the front end.

StevenWadeOddball:
> Nice work Steven! Everything works great as a MINT_USER - I can edit if I am a collaborator on a model plan, and not authorized to mutate any data on a plan that I am not a part of.
> 
> When logged in as MINT_ASSESSMENT I can not edit any part of the model plan (either as a collaborator or not). This is true for discussions, documents, collaborators, and model plan.
> 
> As assessment team member: Screen Shot 2022-08-19 at 9 25 31 AM
> 
> `{"errors":[{"message":"not authorized","path":["updatePlanGeneralCharacteristics"]}],"data":null}`
> 
> The error structure, naming, and messages all make sense and are easy to ingest from the front end.

Thanks @patrickseguraoddball! With regards to not being able to access things as an Assessment user, did the user also have the basic user role? With the previous ticket to update role names, you now need to explicitly have Base User, as well as Assessment to be a valid Assessment user.

patrickseguraoddball:
Ah, gotcha, thanks for that clarification!

patrickseguraoddball (dismissed):
From the FE this works exactly as described 👍

StevenWadeOddball:
@ClayBenson94  and I brainstormed this a little more. I made some changes based on our session.
1. Access Control is in it's own package.
2. ICheckAccess Interface is removed.
    a. Instead, BaseStructPreupdate takes a checkAccess parameter. If it has that parameter, an error is returned if the user isn't a collaborator.

pkg/models/plan_discussion.go
```diff
@@ -29,3 +30,18 @@ const (
...
+}
+
+//DiscussionRelation is an embedded struct meant to satisify the IDiscussionRelation interface
+type DiscussionRelation struct {
+	DiscussionID uuid.UUID `json:"discussionID" db:"discussion_id"`
+}
+
+//GetDiscussionID returns DiscussionID
+func (d DiscussionRelation) GetDiscussionID() uuid.UUID {
+	return d.DiscussionID
```
ClayBenson94, line 46: DiscussionRelation is bundled with `plan_discussion.go`, but the ModelPlanRelationhas its own file. Perhaps we should break this into a file as well, for consistency?

pkg/models/plan_document.go
```diff
@@ -2,14 +2,13 @@ package models
...
-
-	"github.com/google/uuid"
 )
 
 // PlanDocument represents a document attached to the plan
 type PlanDocument struct {
 	BaseStruct
-	ModelPlanID uuid.UUID `json:"modelPlanID" db:"model_plan_id"`
+	ModelPlanRelation
+	// ModelPlanID uuid.UUID `json:"modelPlanID" db:"model_plan_id"`
```
ClayBenson94, line 11: Perhaps we can remove this comment?

pkg/server/routes.go
```diff
@@ -157,16 +157,18 @@ func (s *Server) routes(
...
-		if err != nil {
-			return nil, err
-		}
-		if !hasRole {
-			return nil, errors.New("not authorized")
-		}
-		return next(ctx)
-	}}
+	gqlDirectives := generated.DirectiveRoot{
+		HasRole: func(ctx context.Context, obj interface{}, next graphql.Resolver, role model.Role) (res interface{}, err error) {
```
ClayBenson94, line 161: Having trouble telling what this change is - is this just formatting?

pkg/models/plan_collaborator.go
```diff
@@ -1,16 +1,13 @@
...
-
 // PlanCollaborator represents a plan collaborator
 type PlanCollaborator struct {
 	BaseStruct
-	ModelPlanID uuid.UUID `json:"modelPlanID" db:"model_plan_id"`
-	EUAUserID   string    `json:"euaUserId" db:"eua_user_id"`
-	FullName    string    `json:"fullName" db:"full_name"`
-	TeamRole    TeamRole  `json:"teamRole" db:"team_role"`
+	ModelPlanRelation
+	// ModelPlanID uuid.UUID `json:"modelPlanID" db:"model_plan_id"`
```
ClayBenson94, line 7: Can probably get rid of this comment!

pkg/storage/SQL/check_if_collaborator.sql
```diff
@@ -0,0 +1,6 @@
+SELECT EXISTS(
+        SELECT 1
+        FROM plan_collaborator
+        WHERE model_plan_id = :model_plan_id
+            AND eua_user_id = :eua_user_id
+    ) AS isCollaborator
```
ClayBenson94, line 6: Is the `AS isCollaborator` needed?

pkg/storage/SQL/check_if_collaborator_discussion_id.sql
```diff
@@ -0,0 +1,7 @@
+SELECT EXISTS(
+    SELECT 1
+    FROM PLAN_COLLABORATOR
+    INNER JOIN PLAN_DISCUSSION ON PLAN_COLLABORATOR.model_plan_id = PLAN_DISCUSSION.model_plan_id
+    WHERE PLAN_DISCUSSION.id = :discussion_id
+        AND PLAN_COLLABORATOR.eua_user_id = :eua_user_id
+) AS isCollaborator
```
ClayBenson94, line 7: Same here (Is `AS isCollaborator` needed?)

ClayBenson94 (changes requested):

pkg/models/plan_discussion.go
StevenWadeOddball: I think that's a good call! Thanks!

pkg/storage/SQL/check_if_collaborator.sql
StevenWadeOddball: I tested it, It isn't actually needed, but I do like the clarity it provides in the code. I'd prefer to leave it in, but don't feel too strongly. Thoughts?

pkg/server/routes.go
StevenWadeOddball: Thanks Clay! That's right, I meant to just remove the custom directive, but it looks like I saved a space in here. There is no functional change.

StevenWadeOddball:
Thanks @ClayBenson94 I made the changes you suggested and merged in main. I left isCollaborator in the SQL statements, but I can remove it if you prefer.

pkg/storage/SQL/check_if_collaborator.sql
ClayBenson94: I'm fine with leaving it! Just wanted to ask as I might assume changing it would break something, but from the looks of it since it's just 1 return value the name might just be for visibility in the SQL statement.

pkg/graph/resolvers/model_plan.go
```diff
@@ -20,6 +21,11 @@ func ModelPlanCreate(logger *zap.Logger, modelName string, store *storage.Store,
 		},
 	}
 
+	err := BaseStructPreCreate(logger, plan, principal, store, false)
```
ClayBenson94, line 24: I know this it nit-picky, but can you leave a comment here explaining why access control is set to false? That way it's clear to anyone reading that it's disabled here because its the initial model plan, so of _course_ (obvious to us) you can't check for an existing collaborator without an error.

ClayBenson94 (changes requested):
One final (tiny) change, but this looks good! (I promise I can approve after this 😂)

pkg/graph/resolvers/model_plan.go
StevenWadeOddball: Thanks! Sounds good @ClayBenson94, change made and pushed.

ClayBenson94 (approved):

PR #181, opened 2022-08-16
Running server tests in CI using scripts/dev

ClayBenson94:
# EASI-000

## Changes and Description

- Change 1
- Change 2

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -112,6 +112,21 @@ jobs:
         run: |
           ./scripts/github-action-announce-broken-branch
 
+  server_test_temp:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
+      - name: Run backend server tests
+        run: |
+          scripts/dev up:backend
+          scripts/dev test:go
+          scripts/dev down
+      - name: Announce failure
+        if: ${{ failure() }}
+        run: |
+          ./scripts/github-action-announce-broken-branch
+
   e2e_tests:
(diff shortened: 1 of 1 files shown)


.github/workflows/run_tests.yml
```diff
@@ -57,56 +56,71 @@ jobs:
...
+      #     aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+      #     aws-region: us-west-2
+      - name: Set env vars with direnv
+        uses: HatsuneMiku3939/direnv-action@v1
+      - name: Get docker images
+        run: docker-compose pull
+      - name: Docker Layer Caching
+        uses: satackey/action-docker-layer-caching@v0.0.11
       - name: Run backend server tests
-        env:
```
StevenWadeOddball, line 70: This looks pretty great @ClayBenson94! I'm curious about the environment variables being removed. Are they just no longer needed because of the method we are running them? Or were they redundant etc?

ClayBenson94: Yeah, we used to hard-code all these env vars, but since we can run all our tests locally I decided to just fall back on using [direnv-action](https://github.com/CMSgov/mint-app/pull/181/files/b34f20e703d7696ded38d848f4cdfd129f450336#diff-c2e9f7c1bfaf65da79bbd29ed3ce389b2acc8b4099a257914a7292a66e6b9bc9R65-R66) to pull env vars

StevenWadeOddball (approved):
Looks good to me! Thanks for explaining about using the envrc file for testing in CI instead of explicitly setting variables.

PR #184, opened 2022-08-18
Fix: The same EUAID will only receive subscription updates on the most recent subscription

OddTomBrooks:
# NOREF/ws_pubsub_subscriberid_fix

## Changes and Description

- Websocket pubsub subscription now generates a UUID rather than using the EUAID as a Subscriber ID

## How to test this change

- Start two subscriptions to onLockTaskListSectionContext.
```
{
    "payload": {
        "operationName": "onLockTaskListSectionContext",
        "query": "subscription onLockTaskListSectionContext($modelPlanID: UUID!) { onLockTaskListSectionContext(modelPlanID: $modelPlanID) { changeType lockStatus { modelPlanID section lockedBy } actionType } }",
        "variables": {
            "modelPlanID": "f11eb129-2c80-4080-9440-439cbe1a286f"
        }
    },
    "type": "start"
}
```

- Lock a section on the same model ID.
```
mutation lockTaskListSection($modelPlanID: UUID!, $section: TaskListSection!)
{
    lockTaskListSection(modelPlanID: $modelPlanID, section: $section)
}
```
```
{
  "modelPlanID": "f11eb129-2c80-4080-9440-439cbe1a286f",
  "section": "IT_TOOLS"
}
```

-- Observe that both subscriptions are updated with the lock added status
- Close the second subscription opened
-- Observe that the remaining subscription is updated with the lock removed status

## PR Author Review Checklist

## PR Reviewer Guidelines


cypress/support/graphql-test-utils.js
```diff
@@ -2,6 +2,7 @@
 export const hasOperationName = (req, operationName) => {
   const { body } = req;
   return (
+    // eslint-disable-next-line no-prototype-builtins
```
StevenWadeOddball, line 5: Can you clarify why this needed to be added? Is it just something that makes GoLand complain?

StevenWadeOddball (changes requested):
This is looking good @OddTomBrooks ! I pulled it down locally and tested. I ran through your test script though and got a different result.

I was able to make multiple subscriptions in post man and see them in the various tabs, so the main issue seems to be fixed.

```
Close the second subscription opened
-- Observe that the remaining subscription is updated with the lock removed status
```
However, when I disconnected from one ws, I didn't see any message sent to the other socket. Also, when I queried the socket, I still see the section being locked. When I've disconnected both sockets, I still see the section being locked. this could make sense to me, but it doesn't seem to match what your test case suggests. Thoughts?

 (note, I did this a few ways, subscribing as a different user than made the lock, subscribing as the same user that made the lock etc, all in POSTman)

cypress/support/graphql-test-utils.js
OddTomBrooks: This is per Clay's suggestion to suppress an unrelated linting error

OddTomBrooks:
@StevenWadeOddball Strange, I tested that and it was working. I'll run through it again.

StevenWadeOddball:
> @StevenWadeOddball Strange, I tested that and it was working. I'll run through it again.

@OddTomBrooks, sounds good! Let me know if you can't recreate it, I could be doing something wrong, or maybe Postman is giving unexpected results?

OddTomBrooks:
> > @StevenWadeOddball Strange, I tested that and it was working. I'll run through it again.
> 
> @OddTomBrooks, sounds good! Let me know if you can't recreate it, I could be doing something wrong, or maybe Postman is giving unexpected results?

Gave it another run and it worked, updated the PR with statements. Try with these @StevenWadeOddball

StevenWadeOddball (approved):
Thanks for the update @OddTomBrooks. The issue was because I wasn't testing with the context subscription. This looks good to me now after testing! 

I have had a strange issue that I can't reproduce consistently. Occasionally, the lock removed message doesn't get sent to the remaining websocket when you disconnect the other. It seemed to happen mostly when the app just started. However, I used debugging to try and reproduce and did not have the effect again. This might just be a strange one off edge case, but I want to document it in case it occurs again. I think we are good to move forward with the changes.

This image shows the time stamps of disconnecting. Note that the other connection didn't receive the Lock removed message. Querying the locks does reveal that it has been removed, it's just that the notification wasn't received.
image

PR #185, opened 2022-08-18
Easi 2258/nda acceptance

StevenWadeOddball:
# EASI-2258

## Changes and Description

- New nda_agreeement Table
- GQL query to get if The NDA was accepted and when
- GQL mutation to accept / decline the NDA

## How to test this change

1. Scripts/dev up:backend
2. Check if you have an NDA
```gql
query NDAInfo{
  
  ndaInfo(){
      agreed
      agreedDts
    
  }
  
  
}
```
3.  Run the mutation to accept the NDA, you can change accept back and forth from `true` and `false`, not that acceptedDTS gets updated if you change from false to true.
```gql
mutation AcceptNDA 
{
    agreeToNDA(agree: true) {
      agreed
      agreedDts
    }
}
```

NOTE, you can also use this mutation without providing agree, as the default value is `true`
```gql
mutation AgreeToNDA 
{
    agreeToNDA{
      agreed
      agreedDts
    }
}
```
## PR Author Review Checklist

## PR Reviewer Guidelines


pkg/graph/schema.graphql
```diff
@@ -1303,6 +1303,15 @@ input PlanOpsEvalAndLearningChanges @goModel(model: "map[string]interface{}") {
 
     status: TaskStatusInput
 }
+
+"""
+NDAAccepted represents whether a user has accepted an NDA or not. If accepted previously, there will be a datestamp visible
+"""
+type NDAAccepted {
```
ClayBenson94, line 1310: Pedantic, I know, but I think we should try and follow nouns for our types over verbs/adjectives. Something like `NDAAcceptance` might read a little more naturally? Or even `NDAAcceptanceInfo`, or maybe even just `NDAInfo`?

pkg/graph/schema.graphql
```diff
@@ -1398,6 +1408,9 @@ unlockAllTaskListSections(modelPlanID: UUID!): [TaskListSectionLockStatus!]!
 
 updatePlanPayments(id: UUID!, changes: PlanPaymentsChanges!): PlanPayments!
 @hasRole(role: MINT_USER)
+
+acceptNDAAgreement(accept: Boolean!): NDAAccepted!
```
ClayBenson94, line 1412: Again, probably pedantic, but since NDA stands for non-disclosure agreement, I don't think we need "Agreement" at the end here, can probably just make this `acceptNDA`

pkg/graph/schema.graphql
ClayBenson94: Actually, I just scrolled down a bit and I think you already solved this! The table calls it `nda_agreement`, so maybe we can just call it `NDAAgreement` here?

ClayBenson94: Probably worth a quick naming audit of `accept` vs. `agree` too, while we're talking about naming!

ClayBenson94 (changes requested):
Some minor comments around naming. The bulk of this looks good from the code itself - I'll give this a test locally as well, but wanted to get naming comments out anyways!

pkg/graph/schema.graphql
StevenWadeOddball: Thanks @ClayBenson94! How do you feel about NDAInfo? I was avoiding calling it NDAAgreement, because this isn't the GO model with the same name (which was the original approach). 

And good call about agree v accept. It looks like in the vernacular of model forms, most places use agreement information etc. It probably makes sense to change accepted to agreed etc.

pkg/graph/schema.graphql
StevenWadeOddball: Good call, I agree.

pkg/graph/schema.graphql
ClayBenson94: Ah, that's true, since we're not returning the raw type that maps to the DB! Yeah, I'm good with `NDAInfo`!

StevenWadeOddball: @ClayBenson94, thanks! Changes made and the test cases are updated

pkg/graph/resolvers/nda_agreement.go
```diff
@@ -0,0 +1,77 @@
...
+)
+
+//NDAAgreementGetByEUA returns an EUA agreement by eua
+func NDAAgreementGetByEUA(logger *zap.Logger, principal authentication.Principal, store *storage.Store) (*model.NDAInfo, error) {
+	nda, err := store.NDAAgreementGetByEUA(logger, principal.ID())
+	accepted := model.NDAInfo{}
+	if err != nil {
+		return nil, err
+	}
+	if nda == nil {
```
ClayBenson94, line 22: Is it worth defensive programming here, and doing a check to see if `agreed` is `true`, but `agreedDts` is `NULL`, and throwing an error?

I can't really imagine a case (right now) that would cause this, but always try and think about cases like this!

pkg/graph/resolvers/nda_agreement.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/cmsgov/mint-app/pkg/authentication"
+	"github.com/cmsgov/mint-app/pkg/graph/model"
+	"github.com/cmsgov/mint-app/pkg/models"
+	"github.com/cmsgov/mint-app/pkg/storage"
+)
+
+//NDAAgreementGetByEUA returns an EUA agreement by eua
+func NDAAgreementGetByEUA(logger *zap.Logger, principal authentication.Principal, store *storage.Store) (*model.NDAInfo, error) {
+	nda, err := store.NDAAgreementGetByEUA(logger, principal.ID())
+	accepted := model.NDAInfo{}
```
ClayBenson94, line 18: Perhaps we should call this variable something like `info`? `accepted` seems like it would be a boolean when reading.

pkg/graph/resolvers/nda_agreement.go
```diff
@@ -0,0 +1,77 @@
...
+		accepted.Agreed = nda.Agreed
+		accepted.AgreedDts = nda.AgreedDts
+	}
+
+	return &accepted, err
+
+}
+
+//NDAAgreementUpdateOrCreate either writes an entry to the nda table, or updates an existing one
+func NDAAgreementUpdateOrCreate(logger *zap.Logger, accept bool, principal authentication.Principal, store *storage.Store) (*model.NDAInfo, error) {
```
ClayBenson94, line 34: Should we also catch instances of `accept` and convert them to `agree` here?

pkg/graph/schema.graphql
```diff
@@ -1398,6 +1408,9 @@ unlockAllTaskListSections(modelPlanID: UUID!): [TaskListSectionLockStatus!]!
 
 updatePlanPayments(id: UUID!, changes: PlanPaymentsChanges!): PlanPayments!
 @hasRole(role: MINT_USER)
+
+acceptNDA(accept: Boolean!): NDAInfo!
```
ClayBenson94, line 1412: Should we also catch instances of `accept` and convert them to `agree` here?

pkg/graph/resolvers/nda_agreement.go
ClayBenson94: (not requesting changes here, specifically, this one's more food for thought)

ClayBenson94 (changes requested):

pkg/graph/resolvers/nda_agreement_test.go
```diff
@@ -0,0 +1,41 @@
+package resolvers
```
ClayBenson94, line 1: Can you add a test case to this file to test the case of agreeing to an already agreed-to NDA, and ensuring it doesn't update it? (Looks like it shouldn't update the `agreed_dts`, and maybe it shouldn't even update `modified_dts`?)

ClayBenson94 (changes requested):

pkg/graph/resolvers/nda_agreement_test.go
StevenWadeOddball: @ClayBenson94, sounds good, I will add the test.

I'm inclined to not handle not updating the `modified_dts` to be in line with the rest of the app. (When we apply changes, we don't do anything to check if the supplied values are equivalent to what's already in the database, we just take the input and attempt to update, capturing the users intent to update). I think the fact that agreed_dts won't be updated in this case should be sufficient. Thoughts?

pkg/graph/schema.graphql
StevenWadeOddball: hmmm I'm not sure for this one here? agreeNDA, doesn't sound right to me, and if we call it acceptNDA, I think the parameter should say accept as well. Thoughts? I think the other sections do make sense to say agree

pkg/graph/resolvers/nda_agreement_test.go
ClayBenson94: I think that's fine. Also in-line with the rest of the app, but a test case around not updating the `agreed_dts` still sounds valuable!

pkg/graph/schema.graphql
ClayBenson94: Or `agreeToNDA`? Then `agree: true` sorta makes sense?

Maybe the bigger question lies un-asked: Do we need `accept: Boolean!` at all? I would assume if you're calling this mutation you're always accepting it, perhaps we can free two birds with one key?

StevenWadeOddball: Hmm, that is a good question, I was approaching it so that you could later remove your agreement by passing false. BUT to your point, would we ever give the end user the option to revoke their agreement? ( I don't think so from the FIGMA).

ClayBenson94: Same. Maybe we just hard-code the value to true and simplify the resolver logic to match? We can always add this later if it is a need, but right now means that we'd probably want to test all the logic/behvaior for passing `false` when we might not have a need for that.

StevenWadeOddball: True, I actually already created Unit tests for passing false, so I have the the test cases, covered.  

I think it might just make sense to leave the agree param, but set a default GQL parameter value to true.
`agreeToNDA(agree: Boolean! = true): NDAInfo!`

It would still simplify the action for the frontend, but leave the logic in the backend in case it's useful later (since the test cases etc are already built out).

StevenWadeOddball: I made the change with the default value, does this approach work for you @ClayBenson94 ?
```gql
mutation AgreeToNDA 
{
    agreeToNDA{
      agreed
      agreedDts
    }
}
```

pkg/graph/resolvers/nda_agreement_test.go
StevenWadeOddball: I added permutations of the use cases to Unit tests.

ClayBenson94 (approved):
Thanks for making these changes, Steven! (And sorry for the nit-picks!)

StevenWadeOddball:
> Thanks for making these changes, Steven! (And sorry for the nit-picks!)

My pleasure! And thanks for the review! It was good feedback.

PR #190, opened 2022-08-25
Easi 2256/follow model plan

StevenWadeOddball:
# EASI-2256

## Changes and Description

- Add Plan_Favorite table
     - GQL queries and mutations to access the favorite information directly
- Update Model Plan Object, GQL and SQL queries to return isFavorite whenever you get a model plan directly
    - Earlier implementation required a separate SQL call to check, vs checking at time of access

Note:
 - If the approach of accessing directly from the model is agreed upon, I think we can remove the planFavoriteCollection query from GQL, and have it only be accessible from the model
 - If desired we could look at updating the favorite table as part of the model plan update mutation, but having separate endpoints seems cleaner.

## How to test this change

1. `scripts/dev up:backend:debug && scripts/dev db:seed`
2. Create a plan Favorite
```gql
mutation addPlanFavorite 
{
    addPlanFavorite(modelPlanID:"6e224030-09d5-46f7-ad04-4bb851b36eab"){
        modelPlanID
        userID
    }
}
```

3. Get ModelPlan and see that it shows it is favorited
```gql
query ModelPlan{
  
  modelPlan(id:"6e224030-09d5-46f7-ad04-4bb851b36eab"){
    id
    modelName
    isFavorite
  }
  
  
}
```
4. Delete the plan Favorite Record
```gql
mutation DeletePlanFavorite 
{
    deletePlanFavorite(modelPlanID:"6e224030-09d5-46f7-ad04-4bb851b36eab"){
        modelPlanID
        userID
    }
}
```
## PR Author Review Checklist

## PR Reviewer Guidelines


StevenWadeOddball:
After a good pairing session, @ClayBenson94 and I decided it makes more sense to not always return the isFavorite field from SQL. Instead, we decided to keep the separate query, we will look into data loaders in the future if needed to to avoid an N+1 issue.

migrations/V18__Add_Plan_Favorite.sql
```diff
@@ -0,0 +1,19 @@
...
+
+);
+ALTER TABLE plan_favorite
+ADD CONSTRAINT fk_favorite_plan FOREIGN KEY (model_plan_id)
+REFERENCES public.model_plan (id) MATCH SIMPLE
+ON UPDATE NO ACTION
+ON DELETE NO ACTION;
+
+CREATE UNIQUE INDEX idx_plan_favorite_user
+ON plan_favorite(model_plan_id, user_id);
```
ClayBenson94, line 19: Postgres will automatically have an index on just `id` right? Since it's a primary key?

pkg/graph/resolvers/plan_favorite.go
```diff
@@ -0,0 +1,66 @@
...
+
+//PlanFavoriteCollection returns a list of favorites
+func PlanFavoriteCollection(logger *zap.Logger, principal authentication.Principal, store *storage.Store) ([]*models.PlanFavorite, error) {
+	favorites, err := store.PlanFavoriteCollectionByUser(logger, principal.ID(), false)
+	if err != nil {
+		return nil, err
+	}
+
+	return favorites, err
+}
```
ClayBenson94, line 21: Do we still need this now that we don't have a query to get all of a users plan favorites? (Thinking this code can go away, and can be re-introduced if we find a need for it)

pkg/graph/resolvers/plan_favorite.go
```diff
@@ -0,0 +1,66 @@
...
+
+//IsPlanFavorited checks if a model plan is a favorite.
+func IsPlanFavorited(logger *zap.Logger, principal authentication.Principal, store *storage.Store, modelPlandID uuid.UUID) (bool, error) {
```
ClayBenson94, line 24:
```suggestion
func IsPlanFavorited(logger *zap.Logger, principal authentication.Principal, store *storage.Store, modelPlanID uuid.UUID) (bool, error) {
```

Just a typo in a name here!

pkg/storage/plan_favoriteStore.go
```diff
@@ -0,0 +1,125 @@
...
+	err = stmt.Get(&delFavorite, arg)
+	if err != nil {
+		return nil, err
+	}
+
+	return &delFavorite, nil
+}
+
+//PlanFavoriteGet returns a plan favorite
+func (s *Store) PlanFavoriteGet(logger *zap.Logger, EUAID string, modelPlanID uuid.UUID) (*models.PlanFavorite, error) {
```
ClayBenson94, line 106: Should this function be named a bit more clearly to hint at what it's getting the plan by? Maybe something like

```suggestion
func (s *Store) PlanFavoriteGetByModelIDAndEUA(logger *zap.Logger, EUAID string, modelPlanID uuid.UUID) (*models.PlanFavorite, error) {
```

It's a bit verbose, but might help (and is in line with some other store functions, like [ModelPlanGetByID](https://github.com/CMSgov/mint-app/blob/e92e7ba3a39bd1e24feac6f06bf25648d4b13c17/pkg/storage/model_planStore.go#L99))

ClayBenson94 (changes requested):

migrations/V18__Add_Plan_Favorite.sql
StevenWadeOddball: @ClayBenson94, correct! It will already have a unique index on ID. This index is just to enforce that there is only one plan favorite per user per model. It should also make the query to get by model plan ID and user_id more performant.

pkg/graph/resolvers/plan_favorite.go
StevenWadeOddball: Good call, I removed the store, and resolver methods for this, as well as the related SQL.

OddTomBrooks:
@StevenWadeOddball When running the testing steps I run into the following error in addPlanFavorite:
image

OddTomBrooks:
Adding and deleting plan favorites do work but both give graphql errors, likely inputs just need to be nullable

pkg/graph/resolvers/plan_favorite.go
```diff
@@ -0,0 +1,74 @@
...
+
+	favorite, err := PlanFavoriteGet(logger, principal, store, modelPlanID)
+	if err != nil {
+		return false, err
+	}
+	isFavorite := false
+	if favorite != nil {
+		isFavorite = true
+
+	}
```
OddTomBrooks, line 24: isFavorite := (favorite != nil)

pkg/graph/resolvers/plan_favorite.go
```diff
@@ -0,0 +1,74 @@
...
+
+	favorite := models.PlanFavorite{
+		UserID: principal.ID(),
+		ModelPlanRelation: models.ModelPlanRelation{
+			ModelPlanID: modelPlandID,
+		},
+		BaseStruct: models.BaseStruct{
+			CreatedBy: principal.ID(),
+		},
+	}
```
OddTomBrooks, line 40: Extract to helper method or constructor

pkg/graph/schema.graphql
```diff
@@ -15,7 +15,7 @@ type CurrentUser {
   launchDarkly: LaunchDarklySettings!
 }
 """
-UUIDs are represented using 36 ASCII characters, for example B0511859-ADE6-4A67-8969-16EC280C0E1A
+UUIDs are represented using 36 ASCII characters, for example B05859-ADE6-4A67-8969-16EC280C0E1A
```
OddTomBrooks, line 18: Why was this changed?

pkg/graph/resolvers/plan_favorite.go
ClayBenson94: We saw this a bunch across the codebase - there's actually a few instances where embedded structs are created like this, so I think it might make sense to do this across the codebase. We could still opt to do it here (just for plan favorites), but if we do, we should track the work to do it everywhere else as well. (Perhaps a new function like `NewPlanFavorite` would do well?

pkg/storage/SQL/plan_favorite_create.sql
```diff
@@ -0,0 +1,26 @@
...
+    created_dts,
+    modified_by,
+    modified_dts
+)
+VALUES (
+    :id,
+    :model_plan_id,
+    :user_id,
+    :created_by,
+    :created_dts,
```
ClayBenson94, line 15: I think the issue [Tom brings up here](https://github.com/CMSgov/mint-app/pull/190#issuecomment-1230986948) is from this line of code.

Most (all?) of the other SQL files that create entities in our DB don't take `created_dts` or `modified_dts` as parameters, they just fall back on the defaults (which are `CURRENT_TIMESTAMP` and `NULL`, respectively), which I think makes sense here too.

ClayBenson94 (dismissed):

pkg/storage/SQL/plan_favorite_create.sql
StevenWadeOddball: @ClayBenson94, @OddTomBrooks , thanks for catching this. That is definitely the issue, and it should be using CURENT_tIMESTAMP. Great find!

pkg/graph/schema.graphql
StevenWadeOddball: huh, I think I just did this on accident. I'll revert it when I make the other changes.

pkg/graph/resolvers/plan_favorite.go
StevenWadeOddball: thanks for the suggestion, that is better

OddTomBrooks (approved):
LGTM

ClayBenson94 (approved):

PR #196, opened 2022-08-30
Easi 1883/email client

OddTomBrooks:
# EASI-1883

## Changes and Description

- Wrote a small reusable email service
- Built an interface layer to separate dependency injection from implementation

## How to test this change

- Install and run Mailcatcher as defined in https://mailcatcher.me/
- Run unit tests
- Assert that all unit tests succeed
- Assert that an email is captured in your Mailcatcher web interface

## PR Author Review Checklist

## PR Reviewer Guidelines


go.mod
```diff
@@ -78,7 +78,9 @@ require (
 	github.com/spf13/jwalterweatherman v1.1.0 // indirect
 	github.com/spf13/pflag v1.0.5 // indirect
 	github.com/subosito/gotenv v1.3.0 // indirect
+	github.com/toorop/go-dkim v0.0.0-20201103131630-e1cd1a0a5208 // indirect
```
ClayBenson94, line 81: Do you know why we need `go-dkim` in our go.mod here? I don't see it imported or used anywhere, so just wanted to double check that we still need this.

pkg/shared/oddmail/go_simple_mail_service.go
```diff
@@ -0,0 +1,85 @@
...
+import (
+	"errors"
+
+	mail "github.com/xhit/go-simple-mail/v2"
+)
+
+// GoSimpleMailService is an EmailService implementation for the GoSimpleMail library
+type GoSimpleMailService struct {
+	smtpServer *mail.SMTPServer
+	smtpClient *mail.SMTPClient
```
ClayBenson94, line 12: My main question around this package is how it works with SES. It seems like on EASi, we use the AWS SDK which seems to handle the credentialing through IAM (from what I can tell?) - seems like we'd need to get SMTP credentials for SES for this to work - do you know how this would look?

I think it's nice that we have a single client that always uses SMTP (whether thats locally for mailcatcher, or deployed with AWS SES), but I just want to make sure that this approach will work once we deploy, since it's not like how we've done it before!

ClayBenson94 (changes requested):

ClayBenson94:
@OddTomBrooks Also looks like there might be a failing server test. Not sure it looks related - perhaps updating the branch with `main` can fix it?

pkg/shared/oddmail/go_simple_mail_service.go
OddTomBrooks: This is the documentation I was reading regarding the AWS SES SMTP interface:
https://docs.aws.amazon.com/ses/latest/dg/send-email-smtp.html

It seems straightforward to me, but it's true this is an area of risk. I suggest we do a spike dedicated to AWS SES configuration and testing the SMTP interface with our production criteria.

go.mod
OddTomBrooks: This is likely an indirect dependency of GoSimpleMail to support certification of our SMTP server.

OddTomBrooks: Yep:
https://github.com/xhit/go-simple-mail/blob/bfa248936baba57cf414c64e6b10e2f5a1d01c34/go.mod#L3

pkg/shared/oddmail/go_simple_mail_service.go
ClayBenson94: I think as long as we're able to get SMTP credentials (endpoint, user, pass) we'll probably be fine. Guessing those are one-time, and we can generate them and configure/set them in SSM

StevenWadeOddball:
@OddTomBrooks , great work! I really like this approach for an email client.

Have you considered setting up mailcatcher as part of the local docker image? I believe EASI is set up that way, and it would be cleaner than everyone needing to install this manually.
[EASI docker- compose](https://github.com/CMSgov/easi-app/blob/main/docker-compose.yml#L88-L89 )

docker-compose.yml
```diff
@@ -83,3 +83,5 @@ services:
       /usr/bin/mc mb --quiet s3local/mint-app-file-uploads/;
       /usr/bin/mc policy set public s3local/mint-app-file-uploads;
       "
+  email:
+    image: dockage/mailcatcher:0.7.1
```
StevenWadeOddball, line 87:
```suggestion
    image: dockage/mailcatcher:0.7.1
        ports:
      - 1080:1080
```

When I tested, I couldn't access mail catcher without adding this port mapping. With it, I can access the web UI. Mind adding that line in?

StevenWadeOddball (changes requested):

docker-compose.yml
ClayBenson94: The [easi-app definition for MailCatcher](https://github.com/CMSgov/easi-app/blob/main/docker-compose.override.yml#L68-L71) also forwards port `1025`, which I think is the default SMTP port for mailcatcher. Might not be necessary, but is nice to have if you want to send an email to mailcatcher from oustide the docker-compose network!

StevenWadeOddball: that's a good point @ClayBenson94! It's a good thing to keep in mind for the future

StevenWadeOddball (approved):
Looks good to me! Thanks Tom

StevenWadeOddball (approved):

PR #198, opened 2022-08-30
EASI-2257/ ADR for worker/scheduler framework 

rileyanderson:
# EASI-2257

## Changes and Description

- Add ADR with decision on worker/scheduler framework

## PR Author Review Checklist

## PR Reviewer Guidelines


StevenWadeOddball:
This is looking good! Thanks @rileyanderson! I'm curious, is there a direction you are leaning for the solution?

rileyanderson:
> This is looking good! Thanks @rileyanderson! I'm curious, is there a direction you are leaning for the solution?

@StevenWadeOddball I am currently leaning towards Faktory. Main benefits I see are: 
- we initially won't need to add Redis (which I think we should eventually)
- Dedicated worker server
- ease of deploying server to ECS
- language agnostic (if we ever have a microservice 🤷‍♂️  we could use the same Faktory server)
- The paid features are features we would definitely use and are worth the cost imo.

Only real downside is the $199/month but I think it is worth it if we have the budget.

StevenWadeOddball (approved):
This sounds like a good approach to me assuming that it works within our budget! We will just need to verify that before we merge and commit to it.

ClayBenson94 (approved):
Thanks for making this! Super excited for us to dig into implementation!

PR #203, opened 2022-08-31
EASI-2093/db audit paradigm adr

StevenWadeOddball:
# EASI-2093

## Changes and Description

- Introduce [ADR](https://github.com/CMSgov/mint-app/blob/EASI-2093/db_audit_paradigm_adr/docs/adr/0007-adr-data-auditing.md) for auditing 
   -   [POC](EASI-2093/db_audit_paradigm) branch 

## PR Author Review Checklist

## PR Reviewer Guidelines


ClayBenson94:
Hey @StevenWadeOddball! I looked at the POC branch and saw a few migrations that looks like they were all comments - can you provide an overview of the different migrations? (I can't say I've worked much with triggers this complex, they're a bit over my head!)

StevenWadeOddball:
> Hey @StevenWadeOddball! I looked at the POC branch and saw a few migrations that looks like they were all comments - can you provide an overview of the different migrations? (I can't say I've worked much with triggers this complex, they're a bit over my head!)

Great question @ClayBenson94! So the first few migrations on the POC branch are some of my musings about how we might want to structure our specific implementation.  The bread in butter for functional testing is [migration 22](https://github.com/CMSgov/mint-app/blob/EASI-2093/db_audit_paradigm/migrations/V22__Example_Audit.sql), which is an example I got from https://wiki.postgresql.org/wiki/Audit_trigger_91plus. I think it is a pretty good starting point! But we'd want to change some details in it (eg, we don't need to store quite as much information as it's storing, and we probably want to abstract some data into practical fields (eg primary key, foreign key etc).

It creates the table [logged_actions](https://github.com/CMSgov/mint-app/blob/EASI-2093/db_audit_paradigm/migrations/V22__Example_Audit.sql#L39-L57), and inserts data there for every changed operation.  It sets up the trigger [here](https://github.com/CMSgov/mint-app/blob/EASI-2093/db_audit_paradigm/migrations/V22__Example_Audit.sql#L84-L143), and then creates some wrapper functions to set the auditing on on a table level. (I've implemented that [here](https://github.com/CMSgov/mint-app/blob/EASI-2093/db_audit_paradigm/migrations/V22__Example_Audit.sql#L255) on model plan).

The data right now is being stored in hstore format. You can see in [this area](https://github.com/CMSgov/mint-app/blob/EASI-2093/db_audit_paradigm/migrations/V22__Example_Audit.sql#L124-L142), the trigger is casting old and new values into an hstore, and writing to the table. 

So, every time the table is modified, a row gets written here. Further changes would be to add ignored fields (modified by, modified_dts etc). 

We can query this data with something like this
```sql
SELECT 
row_data ->'id' as ID,table_name,transaction_id, 
(EACH(changed_fields)).*
 

from audit.logged_actions where 

row_data ->'id' ='6e224030-09d5-46f7-ad04-4bb851b36eab'
```

That returns a result like this 
image

I'm happy to demo it sometime too if you like!

ClayBenson94 (approved):

PR #209, opened 2022-09-12
Content Loading

OddTomBrooks:
# NOREF

## Changes and Description

- A simple agnostic interface to allow content provisioning
- An implementation that loads files based on extension from a directory

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/pkg/shared/content/file_loader.go
+++ b/pkg/shared/content/file_loader.go
@@ -1,0 +1,102 @@
+package content
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+// FileLoader is a utility to load all files in a given directory filtered by extension
+type FileLoader struct {
+	contentDirectory    string
+	fileExtensionFilter string
+	fileData            map[string]string
+}
+
+// NewFileLoader creates and returns a new instance of a FileLoader
+func NewFileLoader(directory string, extension string) *FileLoader {
+	return &FileLoader{contentDirectory: directory, fileExtensionFilter: extension}
+}
+
+// Load searches for all files matching config specification and loads them into fileData
+func (f *FileLoader) Load() error {
+	files, err := f.findContentFilesMatchingExtensionFilter()
+	if err != nil {
+		return err
+	}
+
+	return f.loadContentFromFiles(files)
+}
+
+func (f *FileLoader) loadContentFromFiles(files []string) error {
+	_ = f.Reset()
+	for _, path := range files {
+		err := f.loadContentFromFile(path)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (f *FileLoader) loadContentFromFile(path string) error {
+	//nolint:gosec
+	data, err := ioutil.ReadFile(path)
+	if err != nil {
+		return err
+	}
+
+	fileName := strings.TrimSuffix(filepath.Base(path), filepath.Ext(path))
+	f.fileData[fileName] = string(data)
+	return nil
+}
+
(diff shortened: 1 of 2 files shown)


ClayBenson94:
@OddTomBrooks I am looking to confirm my understanding of this code, so feel free to confirm, or deny (with explanation)!

Are we going to instantiate this `FileLoader` struct with raw email templates so that we can do something like this?

```go
// Setup the FileLoader once
var fl := NewFileLoader("./email/templates", ".gohtml")

// Probably somewhere in our email client code
collabTemplate, _ := fl.Get("added_as_collaborator.gohtml")
// Perform templating operations with `collabTemplate`
```

With that as my understanding, I was curious if an existing file caching system ([like the one we use in EASi](https://github.com/CMSgov/easi-app/blob/main/pkg/cedar/core/client.go#L13)) would accomplish a similar thing, or if there was a reason to go for something hand-written.

StevenWadeOddball:
@OddTomBrooks, do you see a multiple places where this interface will be used? I'm not sure I see why we need this. For the email template use case, I think we can more simply embed specific templates individually, without needing to programmatically get them from a directory.

OddTomBrooks:
> @OddTomBrooks, do you see a multiple places where this interface will be used? I'm not sure I see why we need this. For the email template use case, I think we can more simply embed specific templates individually, without needing to programmatically get them from a directory.

It's not strictly necessary, there are multiple methods to approach this problem. I prefer this method as it is more data-driven. You are correct that it is technically simpler to explicitly embed templates, but then you must embed every template you add and remove embedding for every template you remove. By using a content loader you don't have this flexibility.

Conversely, embedding provides more compile-time safety whereas content loading can fail at runtime. This can be reconciled by adding unit tests for expected templates. Then you wind up needing an extra layer of wiring such as in embedding. An alternative approach would be a unit test which samples a list of expected templates and runs them through a unit test. In my opinion this is likely the best approach as it reduces wiring to a simple content file - although at that point you might as well load by configuration file rather than by directory + extension filter. There's a ton of nuance.

If you feel strongly about it I'll move to embedded files but in my opinion we only lose functionality without gaining anything.

StevenWadeOddball:
> > @OddTomBrooks, do you see a multiple places where this interface will be used? I'm not sure I see why we need this. For the email template use case, I think we can more simply embed specific templates individually, without needing to programmatically get them from a directory.
> 
> It's not strictly necessary, there are multiple methods to approach this problem. I prefer this method as it is more data-driven. You are correct that it is technically simpler to explicitly embed templates, but then you must embed every template you add and remove embedding for every template you remove. By using a content loader you don't have this flexibility.
> 
> Conversely, embedding provides more compile-time safety whereas content loading can fail at runtime. This can be reconciled by adding unit tests for expected templates. Then you wind up needing an extra layer of wiring such as in embedding. An alternative approach would be a unit test which samples a list of expected templates and runs them through a unit test. In my opinion this is likely the best approach as it reduces wiring to a simple content file - although at that point you might as well load by configuration file rather than by directory + extension filter. There's a ton of nuance.
> 
> If you feel strongly about it I'll move to embedded files but in my opinion we only lose functionality without gaining anything.

Thanks @OddTomBrooks , I think the main question for me is if this the best approach to get to the acceptance criteria of the email templating and added as a collaborator ticket (1. Send an email when added as a collaborator. 2. implement a templating paradigm to make the email content). Without the context of that ticket, it is hard to say if there is value adding this library to the code base. It's not the main part of that ticket, so it's hard to see if this is overcomplicating  the immediate use case or not.  

If you think this is the best option for loading the template, I'd almost be more inclined to see it in the scope of the ticket rather than a NOREF. I'm not convinced that this is the simplest option (I don't think there will be many templates at this point), but the context of the original ticket implementation would probably be enlightening.

PR #210, opened 2022-09-12
EASI-2296/add cr_tdl to model plan

rileyanderson:
# EASI-2296

## Changes and Description

Add CR and TDL plan model. Link them to a Model Plan. Add Create, Update, Get, Delete queries/mutations.

### Create PlanCrTdl

```gql
mutation CreatePlanCrTdl {
    createPlanCrTdl(
        input: {
            modelPlanID: "6e224030-09d5-46f7-ad04-4bb851b36eab",
            idNumber: "123-456",
            title: "My new CR",
            dateInitiated: "2022-09-13T00:00:00.00Z",
            note: "optionalComments"
        }
    )
    {
        modelPlanID
        title
        idNumber
        dateInitiated
        note
    }
}
```

### Update PlanCrTdl

```gql
mutation UpdatePlanCrTdl {
    updatePlanCrTdl(
        id: "f6d69c87-ca8f-4753-802c-9c96ec7be440",
        changes: {
            title: "My changed CR",
        }
    )
    {
        modelPlanID
        title
        idNumber
        dateInitiated
        note
    }
}
```

### Delete PlanCrTdl

```gql
mutation DeletePlanCrTdl {
    deletePlanCrTdl(
        id: "367e847c-18c4-4058-8bfc-1c850ed9da53"
    )
    {
        modelPlanID
        title
        idNumber
        dateInitiated
        note
    }
}
```

### Get CrTdls with ModelPlan

```gql
query findModelPlans{
  modelPlan(id: "6e224030-09d5-46f7-ad04-4bb851b36eab") {
    modelName
    crTdls {
        id
        modelPlanID
        title
        idNumber
        dateInitiated
        note
    }
  }
}
```

## PR Author Review Checklist

## PR Reviewer Guidelines


pkg/graph/resolvers/plan_cr_tdl_test.go
```diff
@@ -0,0 +1,108 @@
...
+	suite.NoError(err)
+	suite.EqualValues(retCrTdl.ID, crTdl.ID)
+
+}
+
+// TestPlanCrTdlsGetByModelPlanID tests returning all plan_cr_tdls associted with a plan
+func (suite *ResolverSuite) TestPlanCrTdlsGetByModelPlanID() {
+	plan := suite.createModelPlan("My Plan")
+
+	_ = suite.createPlanCrTdl(plan, "123-456", time.Now(), "My CR", "My comments")
```
ClayBenson94, line 101: I genuinely am not sure what the standard is for ignoring return values in go, but you can also opt to entirely omit the `_ =` from these lines, and it'll work the same as using `_`

Not requesting that this be changed, but commenting to open a discussion about it! My thoughts:
- `_ = func()` makes it clear that there IS a return value, we just don't care about it
- `func()` is a bit more terse, maybe more readable?

Honestly, in writing this out, I've already started to lean more towards what you have here!

pkg/graph/schema.graphql
```diff
@@ -46,6 +46,7 @@ type ModelPlan {
   itTools: PlanITTools!
   status: ModelStatus!
   isFavorite: Boolean!
+  crTdls: [PlanCrTdl!]
```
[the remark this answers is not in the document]
ClayBenson94: If you want to go buck wild (TDD, baby!) I could see value in writing a test in `plan_cr_tdl_test.go` that tests `PlanCrTdlsGetByModelPlanID` with no results to see if we get back `null` or if we get `[]`!

ClayBenson94 (changes requested):
Requesting one small change around the GQL schema, and just leaving another comment for discussion!

Awesome work on this, @rileyanderson!

pkg/graph/schema.graphql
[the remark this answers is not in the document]
rileyanderson: Ok cool, that makes sense. I _think_ this still returns an empty array but after reading the [schema documentation](https://graphql.org/learn/schema/#object-types-and-fields) it does make sense to add the bang after the array too since we want to ensure we get an array back -  
> [Episode!]! represents an array of Episode objects. Since it is also non-nullable, you can always expect an array (with zero or more items) when you query the appearsIn field. And since Episode! is also non-nullable, you can always expect every item of the array to be an Episode object.

I'll change this and add some tests!

pkg/graph/resolvers/plan_cr_tdl_test.go
rileyanderson: I honestly just did `_=` because I saw it in some other tests 😅

pkg/models/plan_cr_tdl.go
```diff
@@ -0,0 +1,15 @@
...
+)
+
+// PlanCrTdl represents CRs and TDLs related to a model plan
+type PlanCrTdl struct {
+	BaseStruct
+	ModelPlanRelation
+	IDNumber         string     `json:"idNumber" db:"id_number"`
+	DateInitiated    *time.Time `json:"dateInitiated" db:"date_initiated"`
+	Title            string     `json:"title" db:"title"`
+	OptionalComments string     `json:"optionalComments" db:"optional_comments"`
```
StevenWadeOddball, line 14: Non blocking, but it's more in line with existing tables to call the OptionalComments field note.

StevenWadeOddball (commented):
Overall this looks great! Thanks for adding seed data as well. I left a comment about renaming the optional comments. It's non blocking though (you can change if you agree and have time, up to you). After Clay's suggestions, I think it should be good to go!

pkg/models/plan_cr_tdl.go
rileyanderson: @StevenWadeOddball Changes made

StevenWadeOddball (approved):
The changes look great to me! Thanks Riley! I confirmed that the crTDL returns an empty array, and I see that it now says note instead of optional comments. unit tests pass as well!

ClayBenson94 (approved):
Looks good! Thanks for making these changes 😄

PR #211, opened 2022-09-14
[EASI-2187] Refactor DB Seeding to use resolver functions instead of .sql dump

ClayBenson94:
# EASI-2187

## Changes and Description

- Modify `scripts/dev` to no longer reference `pgdump` or `psql` for dev data
- Create `cmd/dbseed`, which lightly wraps `package resolver` code to create development data more easily.

The main goal behind this change was to make it easier for new changes to implement seeded data, while also giving more visibility into _what_ seeded data is being created.

## Notes for reviewers:
- This PR does _not_ fully implement every single kind of seeded data that our old DB dump had. Notably, it does not seed documents. This is likely to be far easier with EASI-2205
- This PR does _not_ implement as many variations of Model Plans as we previously had. The intent of this PR was to set up a paradigm that allows us to easily add and edit our seeded data, so if you work on a ticket that could seed some data, follow the patterns in this code to enable more kinds of seeded data! (i.e. seeding more task list sections than just `Plan Basics`)

## How to test this change

1. Start the application with `scripts/dev up`
2. Run the Cypress tests with `yarn cypress run`
  - Ensure you have followed [the instructions for setting up Cypress & Okta auth locally](https://github.com/CMSgov/mint-app/blob/0b35902a1f707477716706b5d7bb68428b26b4eb/docs/local_testing.md#cypress-tests-end-to-end-integration-tests), otherwise `cypress/e2e/login.spec.js` will fail!
3. Run `scripts/dev db:clean`
4. Open the UI
5. Log in as `MINT` (using local auth)
6. Note there is no data!
7. Run `scripts/dev db:seed`
8. Refresh the page
9. Note that there's now seeded data!

## PR Author Review Checklist

## PR Reviewer Guidelines


cmd/dbseed/main.go
```diff
@@ -0,0 +1,107 @@
...
+	err := rootCmd.Execute()
+	if err != nil {
+		panic(err)
+	}
+}
+
+// getResolverDependencies takes a Viper config and returns a Store and Logger object to be used
+// by various resolver functions.
+func getResolverDependencies(config *viper.Viper) (*storage.Store, *zap.Logger) {
+	logger, err := zap.NewProduction()
```
StevenWadeOddball, line 41: @ClayBenson94 , what do you think about this being a NOP logger?  The only thing I'm noticing is that it logs a lot of unneeded information to the console when you run the db:seed command.  That's not really the end of the world though, and I guess we could swallow the logging in the scripts/dev db:seed.  I don't really know if it's the end of the world to have the logging, and it might be useful. I also wonder how hard it would be to set the logging level here to just error or greater?

image

StevenWadeOddball (commented):
Thanks @ClayBenson94, this is looks good. I think this puts us in a better position to quickly update seeded data, or not even need to update seeded data files.

Just some thoughts here
 I was wondering where you landed on seeding a model plan with an id?

The only other question I have now is if this is enough seeded data for the front end to work with right now? Or should we hold off merging into main until we have a more complete data set? 

Future improvements could look at decoupling the hard coded values for the data to be a data import from file too. That could give us flexibility to seed different versions.

cmd/dbseed/main.go
ClayBenson94: I like this. Changing!

ClayBenson94:
> Just some thoughts here I was wondering where you landed on seeding a model plan with an id?

I thought the PR could go up without that, and we could either add it or save it as a problem for later. I am tempted to merge sooner rather than later, because I think this change is going to help us as we introduce new functionality (and be able to actually keep up with making seed data as we introduce new features)

> The only other question I have now is if this is enough seeded data for the front end to work with right now? Or should we hold off merging into main until we have a more complete data set?

Same point as above - we can confirm with @patrickseguraoddball / @garyjzhao if this is something they lean heavily on, but I wouldn't imagine it to be a  big enough blocker in comparison to the immediate benefit of readability / being able to maintain it a bit easier.

> Future improvements could look at decoupling the hard coded values for the data to be a data import from file too. That could give us flexibility to seed different versions.

100%! I think something like that sounds great, and would help even more with being able to maintain this data easily.

StevenWadeOddball (dismissed):
Thanks Clay! That this sounds good to me! I think this gives us a good place to iterate off of. If it doesn't interfere with the front end, I'm good to merge it in!

StevenWadeOddball (approved):
Looks good to me, thanks Clay!

PR #213, opened 2022-09-15
[NOREF] Created an initial postman collection

ClayBenson94:
# EASI-000

## Changes and Description

- Created an initial Postman collection
- Modified PR review checklist to have updating the collection as an item

## How to test this change

1. Download [Postman](https://www.postman.com/)
2. Import the collection added in this repository
3. Start the server with `scripts/dev up:backend`
4. Attempt to use the queries in the collection!

## PR Author Review Checklist

## PR Reviewer Guidelines


rileyanderson (commented):
Nice! I have a question about variables.. we don't need to add them now but would it be good to add them as we go? My thinking is if you add seeded data you should be adding that data to Postman as well. Thoughts?

StevenWadeOddball:
> Nice! I have a question about variables.. we don't need to add them now but would it be good to add them as we go? My thinking is if you add seeded data you should be adding that data to Postman as well. Thoughts?

I like that! I think @ClayBenson94 and I had discussed using variables in Postman off seeded data before as well. Right now with this [current PR](https://github.com/CMSgov/mint-app/pull/211#discussion_r971887321), we aren't seeding a id's. That might change in the future, but in the meantime, we'd need postman to set variable programmatically. 

I haven't tried it myself, but it looks like you can set variables using the test tab and parsing a response. ( So we could query model plan and get all the child task list section ids.

I do agree that it would probably be something to iterate on rather than need to ad it all now.

StevenWadeOddball (changes requested):
I think this  is a great addition. 

I noticed the currentUser is incomplete --> What I provide below should be valid

```gql
query currentUser {
    currentUser{
        launchDarkly {
            userKey
        }
    }
}
```

StevenWadeOddball:
@ClayBenson94 & @rileyanderson have either of you used the team feature for Postman to have a shared API there? Any thoughts of doing that verses importing the api from the Repo?

rileyanderson:
> @ClayBenson94 & @rileyanderson have either of you used the team feature for Postman to have a shared API there? Any thoughts of doing that verses importing the api from the Repo?

@StevenWadeOddball I have but it is a paid feature. I think that is what we are trying to avoid for the time being. It's nice, it has versioning that is basically just git where you can create pull requests and stuff. I think that would be the ideal solution but this is a good start I think.

OddTomBrooks (approved):

ClayBenson94:
@OddTomBrooks @StevenWadeOddball @rileyanderson I just pushed up changes to add CRs and TDLs, since that made it to `main`. Re-requesting review, please!

StevenWadeOddball (approved):
I tested this locally and it worked well!

PR #234, opened 2022-09-22
[EASI-2313] Email Templating / Collaborator Added Email

OddTomBrooks:
# EASI-2312

## Changes and Description

- Added a shared package for email templates
- Added mock unit testing for collaborator added emails
- Added email sending on collaborator added

To Do:
_Normally I wouldn't promote a PR while there are known issues but this one is large and I want people to be able to review it while I make last second fixes. Those can then be parsed in the specific commit to ease the burden of review._
- ~~Fix a bug with the HostName directive being malformed~~
- ~~Work with FrontEnd to create a URL alias for unfollowing (or at least make a ticket)~~ @patrickseguraoddball to the rescue!

## How to test this change

- Add a collaborator to a model and observe that MailCatcher captures associated email(s)
- Run the TestCreatePlanCollaborator unit test, observe that it succeeds
- Run the TestAddedAsCollaboratorEmail unit test, observe that it succeeds

## PR Author Review Checklist

## PR Reviewer Guidelines


ClayBenson94:
@OddTomBrooks Seems like the backend server tests are hanging in CI - the last one ran for about 5 1/2 hours before it was cancelled. Are you able to reproduce this locally, or identify if there's a specific test that might be causing this?

StevenWadeOddball:
> @OddTomBrooks Seems like the backend server tests are hanging in CI - the last one ran for about 5 1/2 hours before it was cancelled. Are you able to reproduce this locally, or identify if there's a specific test that might be causing this?

@OddTomBrooks, I tried running it locally, but the backend container just spins. Maybe there is some additional setup required?

config/emailServiceConfig.yaml
```diff
@@ -0,0 +1,2 @@
+host: "email"
```
ClayBenson94, line 1: I think this works for now, but I imagine in the future we're going to need to make these based in environment variables, because this won't fit our needs once we're deployed.

docker-compose.override.yml
```diff
@@ -67,3 +67,8 @@ services:
       - LICENSE_KEY
     volumes:
       - ./../mint-infra-modules/lambda/prince/build:/var/task:ro,delegated
+  email:
+    image: dockage/mailcatcher:0.7.1
+    ports:
+      - "1085:1080"
+      - "1030:1025"
```
ClayBenson94, line 74: Why is this being added when it already exists in `docker-compose.yml`? Seems like this is duplicative of what's in that file.

pkg/email/config.go
```diff
@@ -0,0 +1,16 @@
+package email
+
+var hostName = "mint.cms.gov"
```
ClayBenson94, line 3: Could the contents of this file be inside `pkg/shared/oddmail/email_service_config.go`? Seems like there's 2 spots in the codebase that hold email related config with this file and that one.

pkg/graph/resolvers/plan_collaborator.go
```diff
@@ -4,23 +4,83 @@ import (
...
+	}
+
+	emailSubject, err := emailTemplate.GetExecutedSubject(email.AddedAsCollaboratorSubjectContent{
+		ModelName: modelPlan.ModelName,
+	})
+	if err != nil {
+		return err
+	}
+
+	emailBody, err := emailTemplate.GetExecutedBody(email.AddedAsCollaboratorBodyContent{
```
ClayBenson94, line 70: I think we might need to add `CLIENT_PROTOCOL` to the link because without it the link from the email is relative, so it'll link to localhost:1025/model-plan/etc...

ClayBenson94 (changes requested):

pkg/email/templates/added_as_collaborator_body.template
```diff
@@ -0,0 +1,17 @@
...
+    <h1>
+      <b>MINT</b>
+    </h1>
+    The Model Innovation Tool
+  </title>
+
+  <body>
+    <p>You've been added as a collaborator on {{.ModelName}} in MINT.</p>
+
+    <p><a href="{{.Host}}/models/{{.ModelID}}/task-list">Start collaborating now</a></p>
```
StevenWadeOddball, line 12: @OddTomBrooks , could you double check the [FIGMA](https://www.figma.com/file/6tkIhfNfcsKEAkqXo1qvXH/MINT---Read-Only-View?node-id=805%3A73121) for this please? I think it's missing some areas from FIGMA. I think the FONT is different as well

StevenWadeOddball:
I also wanted to note, you have an action item listed to send the email to all who have favorited a model that a collaborator has been added. From my understanding, I think you currently have it correct, and the email is just meant for the collaborator, not for everyone following.

docker-compose.override.yml
OddTomBrooks: Good point. Leftovers from debugging the mailcatcher connectivity issues. Removed,

pkg/email/config.go
OddTomBrooks: I've been thinking about this over the weekend. Definitely agree. Wondering if it's best to load the configs then overwrite some with environment vars before injecting into the email service. Will ping you about this in Slack to get your thoughts.

.envrc
```diff
@@ -1,119 +1,119 @@
...
-export AIR_CONFIG_DEBUG_WAIT="-c .airDebugWait.conf"
-
-# Load a local overrides file. Any changes you want to make for your local
-# environment should live in that file.
-
-if [ -e .envrc.local ]
-then
-  source_env .envrc.local
-fi
+#! /usr/bin/env bash
```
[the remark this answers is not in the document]
OddTomBrooks: This one was tricky. No idea what the root cause was but even reverting to an old version with the IDE closed didn't fix it. Eventually fixed by running dos2unix across the file.

pkg/email/templates/added_as_collaborator_body.template
OddTomBrooks: I have figured out a good way to reflect Figma style in emails:

- Export Figma frame as PDF
- Open PDF in Adobe Acrobat
- Export PDF as HTML
- Format HTML and change generated fields to directives

pkg/graph/resolvers/plan_collaborator.go
OddTomBrooks: This is a great point. Rather than formatting our client protocol along with hostname I have swapped to using the CLIENT_ADDRESS env var to address both issues.

pkg/email/templates/added_as_collaborator_body.template
```diff
@@ -0,0 +1,28 @@
...
+      <p class="s1" style="padding-top: 11pt;padding-left: 5pt;text-indent: 0pt;text-align: left;">The Model Innovation Tool</p>
+      <p style="text-indent: 0pt;text-align: left;"><br/></p>
+      <p style="padding-top: 14pt;padding-left: 5pt;text-indent: 0pt;line-height: 140%;text-align: left;">You&#39;ve been added as a collaborator on {{.ModelName}} in MINT.</p>
+      <p style="text-indent: 0pt;text-align: left;"><br/></p>
+      <h2 style="padding-left: 5pt;text-indent: 0pt;text-align: left;"><a href="{{.ClientAddress}}/models/{{.ModelID}}/task-list">Start collaborating now</a></h2>
+      <p style="text-indent: 0pt;text-align: left;"><br/></p>
+      <p style="padding-top: 12pt;padding-left: 5pt;text-indent: 0pt;line-height: 140%;text-align: left;"><a href="mailto:MINTTeam@cms.hhs.gov" class="s2" target="_blank">If you have not already been given access to MINT or if you cannot log in with your EUA credentials, please contact the Model Assessment Team at </a><a href="mailto:MINTTeam@cms.hhs.gov" class="a" target="_blank">MINTTeam@cms.hhs.gov</a><a href="mailto:MINTTeam@cms.hhs.gov" class="s2" target="_blank">.</a></p>
```
ClayBenson94, line 23: This line appears to link `mailto:` twice, so the entire line is clickable. We should probably remove the first one and just have the email itself be hyperlinked

ClayBenson94 (changes requested):

pkg/email/templates/added_as_collaborator_body.template
OddTomBrooks: Good catch. Cleaned up, testing now.

pkg/graph/resolvers/resolver_test.go
```diff
@@ -50,15 +54,57 @@ func (suite *ResolverSuite) createDiscussionReply(pd *models.PlanDiscussion, con
...
+func (suite *ResolverSuite) createPlanCollaborator(mp *models.ModelPlan, EUAUserID string, fullName string, teamRole models.TeamRole, emailAddress string) *models.PlanCollaborator {
 	collaboratorInput := &model.PlanCollaboratorCreateInput{
 		ModelPlanID: mp.ID,
 		EuaUserID:   EUAUserID,
 		FullName:    fullName,
 		TeamRole:    teamRole,
-		Email:       email,
+		Email:       emailAddress,
+	}
+
```
StevenWadeOddball, line 65: `Non-blocking`  I'd argue that doing a unit test here is perhaps redundant with the `plan_collaborator_test.go` file. I think this function is just meant to be a utility to test other resolver code, and we test the collaborator in the other file.

pkg/email/templates/added_as_collaborator_body.template
```diff
@@ -0,0 +1,28 @@
...
+         h1 { color: #1A1A1A; font-family:Arial, sans-serif; font-style: normal; font-weight: bold; text-decoration: none; font-size: 39pt; }
+         .s1 { color: #707579; font-family:Arial, sans-serif; font-style: normal; font-weight: normal; text-decoration: none; font-size: 15.5pt; }
+         p { color: #1A1A1A; font-family:Arial, sans-serif; font-style: normal; font-weight: normal; text-decoration: none; font-size: 15.5pt; margin:0pt; }
+         h2 { color: #005BA0; font-family:Arial, sans-serif; font-style: normal; font-weight: bold; text-decoration: none; font-size: 15pt; }
+         .a { color: #005BA0; font-family:Arial, sans-serif; font-style: normal; font-weight: normal; text-decoration: none; font-size: 15.5pt; }
+         .s2 { color: #1A1A1A; font-family:Arial, sans-serif; font-style: normal; font-weight: normal; text-decoration: none; font-size: 15.5pt; }
+      </style>
+   </head>
+   <body>
+      <h1 style="padding-top: 3pt;padding-left: 5pt;text-indent: 0pt;text-align: left;">MINT</h1>
```
StevenWadeOddball, line 16: `Non-Blocking` Thanks for working on the formatting changes for the email @OddTomBrooks ! I think it looks a lot closer. It's not quite matching [FIGMA](https://www.figma.com/file/6tkIhfNfcsKEAkqXo1qvXH/MINT---Read-Only-View?node-id=902%3A73536), but I think that we can say it's close enough for now. We might consider requesting some front end assistance dialing in the final formatting to make sure the font, and font size matches and that we are following USWDS. There are a few other things I noticed, like figma has both words in `unfollow it` as the url. I think we can consider that part of a final pass ticket in the future.

StevenWadeOddball (approved):
Thanks @OddTomBrooks , I think this is in a good place to merge. I noted a couple of `non-blocking` observations, but I think that we can merge this so we can benefit from the functionality sooner, and come back to address these other observations at a later time.

PR #236, opened 2022-09-23
Easi 2338/data auditing

StevenWadeOddball:
# EASI-2338

### Overview
This is a a fairly large PR that introduce an auditing paradigm to MINT. At a high level, it uses triggers to capture insert, delete, and update actions and right a record to the audit table. 

#### Changes that are captured
1. Who inserted or updated a table
    a. deleting a row does not capture who performed the action, as we don't know who performed the action
2. Specific configurable columns entries on insert / delete
   a. Otherwise, we capture a complete record of the initial entry every insert / or delete, which is not needed
3. All changes in a transaction besides columns that are configured to be ignored. (META data changes do not need a separate entry written)

- Implement an auditing schema
    - audit.change -> Store transactional changes
    - audit.table_config -> Store configurations that are used by the audit trigger to audit each table with configurable behavior
- Model_Plan Name History
   - return a string array defaulted in `DESC` order for a plans name history, with optional sort order param.
- Generic GQL endpoint to get a change by primary key, and table name
   - there is no immediate use case for this. It is likely the signature of this will change to not return a map, but instead a generated type, so you can return an array with field name, old value, and new value.
   
 ## Future development
To protect the scope of this change, it is necessary to note what will be developed in the future as needed.
1. Returning generic change history to the front end as an array instead of a map
2. SQL to query based on a search parameter object which can return collection based on shared secondary key (eg all task list sections and model plans in one transaction, as well as a date range)
3. Analysis based on audit changes with who is subscribed to a change.
4. Possibility of other indexes added to the change table as needed (ok primary key and foreign key)
5. Custom View which combines table_config and change.
   

## How to test this change
1. `scripts/dev up:backend:debug && scripts/dev db:seed`
2. Update a model plan ( or other object as you desire) ~ Note, you will need to update the id to an existing model plan ID ~
```gql
mutation modelPlanUpdate{
  
  updateModelPlan(id:"6e224030-09d5-46f7-ad04-4bb851b36eab",changes
      :{
            modelName: "Another Cool Name NEW",
            # modelCategory: DEMONSTRATION,
            # cmsCenters: [CMMI,OTHER] ,           
            # cmsOther: "Excellent Good Center",
            # cmmiGroups: [PATIENT_CARE_MODELS_GROUP, POLICY_AND_PROGRAMS_GROUP],
            status: OMB_ASRF_CLEARANCE
            archived: false

            
      }
      )
      
      {
        id

        modelName
        nameHistory

        archived
        status
        createdBy
        createdDts
        modifiedBy
        modifiedDts
    
  }  
  
}

```
3. Note that the above returns the name history of the model in `DESC` order.
6. If you want to query other changes, you can use this generic GQL endpoint, you can also query the database to see the structure
```gql
query auditChanges{
  auditChanges(tableName: "model_plan",primaryKey: "6e224030-09d5-46f7-ad04-4bb851b36eab")
   {
        id
        primaryKey
        foreignKey
        tableName
        action
        fields
        modifiedBy
        modifiedDts
  }
}
```

## PR Author Review Checklist

## PR Reviewer Guidelines


patrickseguraoddball (commented):
Nice work Steven!  Thanks so much for the sync on FE needs for name history.  This works exactly as expected.  The FE would be able seamlessly integrate this implementation in areas when this type of audit are needed.  The `auditChanges` also works as expected, nice!  Going to leave this approval open for a bit to get other BE eyes on it

scripts/dev
```diff
@@ -323,8 +323,8 @@ namespace :db do
     sh "docker-compose start db_migrate"
     loop do
       sleep 0.5
-      status = `docker-compose ps | grep "db_migrate_1"`
-      match = /Exit (\d+)/.match(status)
+      status = `docker-compose ps | grep "db_migrate-1"`
+      match = /exited \((\d+)\)/.match(status)
```
ClayBenson94, line 327: Was this what was causing the backend to sometimes never spin up, as it never finished waiting for the migraitons?

ClayBenson94 (dismissed):

pkg/graph/schema.graphql
```diff
@@ -1332,6 +1343,27 @@ input PlanCrTdlChanges @goModel(model: "map[string]interface{}") {
...
+
+# type AuditFields {
+#   AuditField[]
+
+# }
+
+# type AuditField {
+#   old: String
+#   new: String
+# }
```
rileyanderson, line 1366: Can these lines be removed? 
```suggestion
```

pkg/graph/schema.graphql
```diff
@@ -1332,6 +1343,27 @@ input PlanCrTdlChanges @goModel(model: "map[string]interface{}") {
     note: String
 }
 
+
+type AuditChange {
```
rileyanderson, line 1347: This is something that doesn't have to be added right now but what it be beneficial to include the objects current values here too? Something like `currentFields: Map!`. That may be overkill but just a thought

scripts/dev
StevenWadeOddball: I wish it was! I don't think it will help that unfortunately, this is specifically with the `db:migrate` command. I don't think the up command calls this script. I ended up recreating the db and migrating it quite a few times, and this command was not analyzing if the migration was complete correctly. (Thanks to Riley for help with the Ruby REGEX!)

pkg/graph/schema.graphql
StevenWadeOddball: Thanks @rileyanderson , I removed it. I had left it because I thought we'd likely change to this in the future, but it is simple enough that we can recreate it later.

pkg/graph/schema.graphql
StevenWadeOddball: Thanks @rileyanderson , for now, I opted to at least not copy the current values in the DB record besides the fields that were changed. I think it would make it quite a bit more complex (though possible) to include it here at the moment. I'm inclined to leave it for now, and we can add that if a future use case calls for it.

PR #238, opened 2022-09-26
[EASI-2189] - Add date warning milestones

patrickseguraoddball:
# EASI-2189
# EASI-2190

## Changes and Description

- Added date warnings for dates in the past
- Ensured dates render on hard refresh
- Updated previous date implementations to not need state and useeffect

This is a small tech debt ticket to add `<DatePickerWarning />` and alerts for date entered in the past.  This is an existing component that needed to be added to Milestones.

Screen Shot 2022-09-26 at 10 18 47 AM

## How to test this change

- Navigate to Milestones on Model Basics
- Enter a date in the past and click outside the input (blur)
- Verify a alert icon and banner appear notifying a past date was entered
- Save the form and do a hard refresh to ensure the dates render on refresh

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/src/views/ModelPlan/TaskList/Basics/Milestones/index.tsx
+++ b/src/views/ModelPlan/TaskList/Basics/Milestones/index.tsx
@@ -33,19 +33,13 @@ import {
 } from 'queries/Basics/types/GetMilestones';
 import { UpdatePlanBasicsVariables } from 'queries/Basics/types/UpdatePlanBasics';
 import UpdatePlanBasics from 'queries/Basics/UpdatePlanBasics';
+import { isDateInPast } from 'utils/date';
 import flattenErrors from 'utils/flattenErrors';
 import sanitizeStatus from 'utils/status';
 import planBasicsSchema from 'validations/planBasics';
 import { NotFoundPartial } from 'views/NotFound';
 
 import './index.scss';
-
-const isDateInPast = (date: string | null): boolean => {
-  if (date && new Date() > new Date(date)) {
-    return true;
-  }
-  return false;
-};
 
 const Milestones = () => {
   const { t } = useTranslation('basics');
@@ -203,6 +197,7 @@ const Milestones = () => {
 
             const handleOnBlur = (e: string, field: string) => {
               if (e === '') {
+                setFieldValue(field, null);
                 return;
               }
               try {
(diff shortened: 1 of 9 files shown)


src/views/ModelPlan/TaskList/Basics/Milestones/index.tsx
```diff
@@ -245,19 +249,30 @@ const Milestones = () => {
...
+                        name="completeICIP"
+                        defaultValue={completeICIP}
+                        onBlur={(e: any) =>
+                          handleOnBlur(e.target.value, 'completeICIP')
+                        }
+                      />
+                      {isDateInPast(values.completeICIP) && (
+                        <DatePickerWarning label={h('dateWarning')} />
+                      )}
+                    </div>
```
ClayBenson94, line 267: With how frequently this shows up, do you think it'd be worth making a component that wraps `<Field as={DatePicker}` that handles the `isDateInPast` logic? I.e. a (naming aside) `<DateFieldWarnInPast>` component?

Not requesting changes per-se, but opening up for discussion!

ClayBenson94: Upon further review in other sections, this might get a bit hairy because there's multiple fields that map to a single warning in some spots.

ClayBenson94 (dismissed):
Tested this out locally and it worked exactly as expected!

Only note I could make is that it seems the values are being set `onBlur` rather than `onChange` - I presume this is because if someone is typing in the textbox rather than using the date-picker we don't want to validate each time they type a character (because there's no way it's valid when they start typing)

I wonder if there's a way to set values immediately on clicking something in the datepicker, but only `onBlur` when someone is typing manually 🤔

src/views/ModelPlan/TaskList/Basics/Milestones/index.tsx
patrickseguraoddball: That's a good idea, I should be able to make the component flexible enough to handle these cases.  I might take a. little time to wrap this in a custom component

patrickseguraoddball: Upon further inspection of the Truss DatePicker, there are only onChange and onBlur handlers.  I'm not seeing any prop for a onSelect callback.  I think I remember this was a discussion Gary and I were having with this component.

PR #242, opened 2022-09-29
[EASI-2224] Front End work for Read-only Model sub page: Payments

garyjzhao:
# EASI-2224

## Changes and Description

- Creates new Payment Read Only component
- Add GetAllPayments query and types
- 

## PR Author Review Checklist

- [ ] Met the ticket's acceptance criteria, or will meet them in a subsequent PR.
- [ ] Added or updated tests for backend resolvers or other functions as needed.
- [ ] Added or updated client tests for new components, parent components, functions, or e2e tests as necessary.
- [ ] Updated the [Postman Collection](../MINT.postman_collection.json) if necessary.

## PR Reviewer Guidelines
- It's best to pull the branch locally and test it, rather than just looking at the code online!
- Check that all code is adequately covered by tests - if it isn't feel free to suggest the addition of tests.
- Always make comments, even if it's minor, or if you don't understand why something was done.

PR #241, opened 2022-09-28
[EASI-2222] Read-only Beneficiaries - FE

garyjzhao:
# EASI-2222

## Changes and Description

- Creates new Beneficiaries Read Only component
- Add new `GetAllBeneficiaries` query and type
- Add `translateTriStateAnswer` to `src/utils/modelPlan.ts`
- Adds test and snaps 

## AC
- Render questions and answers from the task list section
- Should still render unanswered questions - populated with 'No answer entered'
- Testing

## How to test this change
1. Navigate to the Model Summary ( Models on nav > Complete Plan > View this Model Plan on sidebar)
2. Click `Beneficiaries` on sidebar
3. Ensure component is displayed correctly and matches figma

## PR Author Review Checklist

- [ ] Met the ticket's acceptance criteria, or will meet them in a subsequent PR.
- [ ] Added or updated tests for backend resolvers or other functions as needed.
- [ ] Added or updated client tests for new components, parent components, functions, or e2e tests as necessary.
- [ ] Updated the [Postman Collection](../MINT.postman_collection.json) if necessary.

## PR Reviewer Guidelines
- It's best to pull the branch locally and test it, rather than just looking at the code online!
- Check that all code is adequately covered by tests - if it isn't feel free to suggest the addition of tests.
- Always make comments, even if it's minor, or if you don't understand why something was done.

PR #240, opened 2022-09-28
[EASI-2218] Read-only Model sub-page: Team information - FE

garyjzhao:
# EASI-2218

## Changes and Description

- Creates new Team Info Read Only component
- Adds `teamRole` field to query to `src/queries/ReadOnly/GetModelSummary.ts`
- Update `ContactInfo` Side Navigation component to only show Model Leads 

## Acceptance Criteria

- Reusable component for rendering collaborator info
- Testing

## How to test this change

1. Navigate to the Model Summary ( Models on nav > Complete Plan > View this Model Plan on sidebar)
2. Click `Team` on sidebar
3. Ensure component is displayed correctly and matches figma

## PR Author Review Checklist

- [ ] Met the ticket's acceptance criteria, or will meet them in a subsequent PR.
- [ ] Added or updated tests for backend resolvers or other functions as needed.
- [ ] Added or updated client tests for new components, parent components, functions, or e2e tests as necessary.
- [ ] Updated the [Postman Collection](../MINT.postman_collection.json) if necessary.

## PR Reviewer Guidelines
- It's best to pull the branch locally and test it, rather than just looking at the code online!
- Check that all code is adequately covered by tests - if it isn't feel free to suggest the addition of tests.
- Always make comments, even if it's minor, or if you don't understand why something was done.

PR #231, opened 2022-09-21
[EASI-2216] FE work for Read Only Documents

garyjzhao:
# EASI-2216

## Changes and Description

- Creates new Participants and Providers Read Only component
- Adds test and snaps

## Acceptance Criteria

- Remove document should only be allowed with team members with edit access 
- Side contact bar should be hidden
- Testing

## How to test this change

1. Navigate to the Model Summary ( Models on nav > Complete Plan > View this Model Plan on sidebar)
2. Click `Documents` on sidebar
3. Ensure component is displayed correctly and matches figma

## PR Author Review Checklist

- [ ] Met the ticket's acceptance criteria, or will meet them in a subsequent PR.
- [ ] Added or updated tests for backend resolvers or other functions as needed.
- [ ] Added or updated client tests for new components, parent components, functions, or e2e tests as necessary.
- [ ] Updated the [Postman Collection](../MINT.postman_collection.json) if necessary.

## PR Reviewer Guidelines
- It's best to pull the branch locally and test it, rather than just looking at the code online!
- Check that all code is adequately covered by tests - if it isn't feel free to suggest the addition of tests.
- Always make comments, even if it's minor, or if you don't understand why something was done.

PR #243, opened 2022-09-29
[NOREF] More Specific Check for Undefined Query Data for Components

garyjzhao:
# NOREF

# This work is branched off many Read Only subpages. 

### To not be looked at until the following PRs are addressed/merged:
- #242
- #241
- #240
- #231

## Description
The way we currently implement rendering the component after querying for the backend does not properly identify if data is `undefined` (see below)
```js
if ((!loading && error) || (!loading && !data)) {
  return <NotFoundPartial />;
}
```
From the above statement, if `data` is equal to `undefined`, it will not render the `NotFoundPartial` component and continue and error out. It is because `!undefined` is equal to true.

This PR addresses this by adding `data === undefined` to the end of the if statement. See below:
```js
if ((!loading && error) || (!loading && !data) || data === undefined) {
  return <NotFoundPartial />;
}
```

As a result, all optional chaining is no longer needed and code is much cleaner and tighter

## Changes
- Adjusted data if statement check in all subpages (at time of writing)
- Remove optional chaining where applicable
- Added `translateBooleanOrNull` to better handle/translate boolean or null data values

## PR Author Review Checklist

## PR Reviewer Guidelines


src/views/ModelPlan/ReadOnly/GeneralCharacteristics/index.tsx
```diff
@@ -97,13 +109,13 @@ const ReadOnlyGeneralCharacteristics = ({ modelID }: { modelID: string }) => {
     >
       <div className="display-flex flex-justify flex-align-start">
         <h2 className="margin-top-0 margin-bottom-4">{t('heading')}</h2>
-        {status && <TaskListStatusTag status={status} />}
+        <TaskListStatusTag status={status} />
```
garyjzhao, line 112: Example of code cleaned up

src/views/ModelPlan/ReadOnly/ParticipantsAndProviders/index.tsx
```diff
@@ -100,14 +100,14 @@ const ReadOnlyParticipantsAndProviders = ({ modelID }: { modelID: string }) => {
...
-        {status && <TaskListStatusTag status={status} />}
+        <TaskListStatusTag status={status} />
       </div>
 
       <div className="margin-bottom-4 padding-bottom-2 border-bottom-1px border-base-light">
         <ReadOnlySection
           heading={t('whoAreParticipantsQuestion')}
           list
-          listItems={participants?.map(translateParticipantsType)}
+          listItems={participants.map(translateParticipantsType)}
```
garyjzhao, line 110: Another example of code clean up.

because data is for sure not going to be undefined (as per line 42 in this file), participants will never be undefined.

garyjzhao:
Clsoing this PR in favor for https://github.com/CMSgov/mint-app/pull/250

PR #249, opened 2022-10-06
EASI-2337/faktory implementation

rileyanderson:
# EASI-2337

## Changes and Description

-Implement Faktory worker server and manager 

## PR Author Review Checklist

## PR Reviewer Guidelines


pkg/models/base_task_list_section.go
```diff
@@ -38,7 +38,7 @@ func NewBaseTaskListSection(createdBy string, modelPlanID uuid.UUID) baseTaskLis
 
 }
 
-//GetBaseTaskListSection returns the BaseTaskListSection Object embedded in the struct
+// GetBaseTaskListSection returns the BaseTaskListSection Object embedded in the struct
```
rileyanderson, line 41: I think my editor went through and automatically changed all these comments. I can move them to a different PR if we don't want them in there.

ClayBenson94 (changes requested):
Tested locally, and was able to make my own job (and learn a bit about how Faktory works!)

Some small changes, but the functionality all seems to work great locally!

pkg/models/base_task_list_section.go
StevenWadeOddball: >  It's fine with me to leave them here since they are already here.

StevenWadeOddball (approved):
This looks good to me, thanks Riley! I was able to make a job and run it locally.

This would be for future tickets, but what do you think about defining worker functions in separate files? I think that might help keep the code organized so we only register jobs in worker.go, and define the worker functions in files specific to the use case.

rileyanderson:
> This looks good to me, thanks Riley! I was able to make a job and run it locally.
> 
> This would be for future tickets, but what do you think about defining worker functions in separate files? I think that might help keep the code organized so we only register jobs in worker.go, and define the worker functions in files specific to the use case.

@StevenWadeOddball Yep for sure. We could even split it out further to have files for regular job functions, scheduled job functions, etc

ClayBenson94 (approved):

PR #271, opened 2022-10-19
[EASI-2402] FE Updates to Remove Model Plans Modal

garyjzhao:
# EASI-2402

## Changes and Description

- Update warning copy and styles for Remove Modal Plans modal
- Add success message on Home page when successful removal

## Acceptance Criteria

- Updated warning model style per Figma
- Success message on home page after model plan is removed

## PR Author Review Checklist

## PR Reviewer Guidelines


src/views/Home/index.tsx
```diff
@@ -33,7 +34,7 @@ const Home = () => {
         <>
           <NDABanner collapsable />
           <GridContainer>
-            {message}
+            {message && <Expire delay={4000}>{message}</Expire>}
```
patrickseguraoddball, line 37: I'm a little hesitant to add the `Expire` component here.  Another component renders here within the message hook after a user removes themselves from a model plan. The success message on the homepage has crucial information on how to be added back as a collaborator if it was an accident.  I don't think 4 seconds is enough time to process the message.  I'm on the fence about adding a much longer expire time or just leaving it there until they decide to leave the page.  Just consulted with UX and said it's fine to leave it permanent.  

If we want to modify the useMessage hook with an expire state, we can conditionally render the expire wrapper.  In this case I think it would be fine for it to expire, but since another message is depending on it being permanent, we should leave it as permanent for now.

patrickseguraoddball (changes requested):

patrickseguraoddball (approved):
Nice work Gary!

PR #279, opened 2022-10-25
[EASI-2411] Unfollow Email Link

garyjzhao:
# EASI-2411

## Changes and Description

- Changed `NDAWrapper` to pass down previous URL as state to `NDA` component
- Changed `NDA` component to receive "previous URL" as state and push user to that original route
- Created new `Unfollow` component that will send mutation to remove unfollow based on url param
- Add message to show on `/models` page on either success or failure
    -  image
    - image

- Unfollow Cypress test

## Acceptance Criteria

- Route to handle unfollowing a model plan
- Send mutation to unfollow based on url param for modelPlanID
- Direct to /models after handler finishes
- Redirects for the user to accept the NDA if they haven't already done it
- Cypress testing

## PR Author Review Checklist

## PR Reviewer Guidelines


src/views/Unfollow/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+const UnfollowWrapper = ({ children }: UnfollowProps) => {
+  const history = useHistory();
+  const { t } = useTranslation('modelPlan');
+  const { pathname, search } = useLocation();
+  const { showMessageOnNextPage } = useMessage();
+
+  const modelIDToRemove = search
+    .replace(/.*modelID=(.*)\/?/g, '$1')
+    .replace(/\/+$/, '');
+
```
patrickseguraoddball, line 29: This param extraction is natively built in with `URLSearchParams` and can be used like so:

```
const params = new URLSearchParams(search);
const modelIDToRemove = params.get('modelID');
```

src/views/Unfollow/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+type UnfollowProps = {
+  children: React.ReactNode;
+};
+
+const UnfollowWrapper = ({ children }: UnfollowProps) => {
+  const history = useHistory();
+  const { t } = useTranslation('modelPlan');
+  const { pathname, search } = useLocation();
+  const { showMessageOnNextPage } = useMessage();
```
patrickseguraoddball, line 24: Gary, I'm not sure having a wrapper for a single route is the right method.  As is, this wrapper computes on every route change, as will changes to all the `useEffect` dependencies.  We only need to compute this logic if the router hits `/unfollow`.  This logic should be wrapped in a component that is only trigger within the `BrowserRouter`.  Wrappers like this are only necessary when you need to check states on every route in the app, like NDA and Subscriptions.  I think `/unfollow' should have its own designated route in our `App` component and only be called if the route is hit

patrickseguraoddball (changes requested):
Nice work Gary, I think it might need a rearchitecting.  I'm free to sync if you want to go over the proposed strategy.

src/views/App/index.tsx
```diff
@@ -80,6 +81,8 @@ const AppRoutes = () => {
       <Route path="/signin" exact component={Login} />
       <SecureRoute path="/user-diagnostics" component={UserInfo} />
 
+      <Route path="/unfollow" exact component={Unfollow} />
```
patrickseguraoddball, line 84: This should be a `SecureRoute` component.  Right now our dev login doesn't support maintaining an original url after login, but Okta does.  It should not be a problem in prod/once we get IdP connected

src/views/NDAWrapper/index.tsx
```diff
@@ -14,7 +14,10 @@ const NDAWrapper = ({ children }: NDAWrapperProps) => {
 
   useEffect(() => {
     if (user?.acceptedNDA && user?.acceptedNDA?.agreed === false) {
-      history.push('/pre-decisional-notice');
+      history.push({
+        pathname: '/pre-decisional-notice',
+        state: { nextState: pathname !== '/pre-decisional-notice' && pathname }
+      });
     }
```
patrickseguraoddball, line 21: I realize this piece of code doesn't preserve any `search` params.  For this to accurately work with NDA signing, you need to append `search` to `pathname` if the model plan ID is to be parsed in the `Unfollow` component.  Something like this:

`const { pathname, search } = useLocation();`

Then the `nextState` field could contain something like so:
`pathname !== '/pre-decisional-notice' && pathname + (search || '')`

patrickseguraoddball (changes requested):
Noticed a few more things, after this should be good to go!

patrickseguraoddball (approved):
LG Thanks Gary!

PR #296, opened 2022-11-08
[EASI-2423] Operational Solution Plan Document Links

OddTomBrooks:
# EASI-2423

## Changes and Description

Added endpoints to Create and Remove operational solution plan document links

- CreatePlanDocumentSolutionLinks
  - Creates an array of plan document solution links
  - Returns the resulting array of plan document solution links with the generated ID

- RemovePlanDocumentSolutionLink
  - Removes a plan document solution link by UUID
  
- PlanDocumentSolutionLinks
  - Gets a list of document solution links associated with a model plan ID

## How to test this change

Run the following endpoints and observe that they function as expected:

- createPlanDocumentSolutionLinks
- removePlanDocumentSolutionLink

## PR Author Review Checklist

## PR Reviewer Guidelines


migrations/V36__Add_Plan_Document_Solution_Link.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE plan_document_solution_link (
+    id UUID PRIMARY KEY NOT NULL,
+    model_plan_id UUID NOT NULL,
```
StevenWadeOddball, line 3: `model_plan_id` is a duplicate key here, I don't think it should be in this table as it is already in the plan_document table. Solution to Document is really the only link we need, we can query other data through joins.

migrations/V36__Add_Plan_Document_Solution_Link.sql
```diff
@@ -0,0 +1,12 @@
...
+    solution_id INTEGER NOT NULL,
+    document_id UUID NOT NULL,
+
+    --META DATA
+    created_by EUA_ID NOT NULL,
+    created_dts TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    modified_by EUA_ID,
+    modified_dts TIMESTAMP WITH TIME ZONE
+);
```
StevenWadeOddball, line 12: I think we should have some constraints and foreign key relationships added here, eg force a valid solution_id and document_id, as well as perhaps constraining that the combination of solution / document is unique.

pkg/graph/schema.resolvers.go
```diff
@@ -371,16 +372,12 @@ func (r *mutationResolver) AddOrUpdateCustomOperationalNeed(ctx context.Context,
 
 // UpdateCustomOperationalNeedByID is the resolver for the updateCustomOperationalNeedByID field.
 func (r *mutationResolver) UpdateCustomOperationalNeedByID(ctx context.Context, id uuid.UUID, customNeedType *string, needed bool) (*models.OperationalNeed, error) {
-	principal := appcontext.Principal(ctx)
-	logger := appcontext.ZLogger(ctx)
-	return resolvers.OperationalNeedCustomUpdateByID(logger, id, customNeedType, needed, principal, r.store)
```
StevenWadeOddball, line 376: I think changes got overwritten here?

pkg/graph/schema.resolvers.go
```diff
@@ -371,16 +372,12 @@ func (r *mutationResolver) AddOrUpdateCustomOperationalNeed(ctx context.Context,
...
+	panic(fmt.Errorf("not implemented: UpdateCustomOperationalNeedByID - updateCustomOperationalNeedByID"))
 }
 
 // AddOrUpdateOperationalSolution is the resolver for the addOrUpdateOperationalSolution field.
 func (r *mutationResolver) AddOrUpdateOperationalSolution(ctx context.Context, operationalNeedID uuid.UUID, solutionType models.OperationalSolutionKey, changes map[string]interface{}) (*models.OperationalSolution, error) {
-	principal := appcontext.Principal(ctx)
-	logger := appcontext.ZLogger(ctx)
-	return resolvers.OperationalSolutionInsertOrUpdate(logger, operationalNeedID, solutionType, changes, principal, r.store)
+	panic(fmt.Errorf("not implemented: AddOrUpdateOperationalSolution - addOrUpdateOperationalSolution"))
```
StevenWadeOddball, line 380: I think changes got overwritten here?

pkg/graph/resolvers/plan_document_solution_links.go
```diff
@@ -0,0 +1,24 @@
...
+	return store.PlanDocumentSolutionLinksCreate(logger, planDocumentSolutionLinks)
+}
+
+// PlanDocumentSolutionLinkRemove implements resolver logic to delete a plan document solution link
+func PlanDocumentSolutionLinkRemove(logger *zap.Logger, id uuid.UUID, store *storage.Store) (bool, error) {
+	return store.PlanDocumentSolutionLinkRemove(logger, id)
+}
+
+// PlanDocumentSolutionLinksGetByID implements resolver logic to get plan document solution links associated with a plan ID
+func PlanDocumentSolutionLinksGetByID(logger *zap.Logger, id uuid.UUID, store *storage.Store) ([]*models.PlanDocumentSolutionLink, error) {
```
StevenWadeOddball, line 22: I think that this should be getting by solution ID, not model plan ID

pkg/graph/schema.graphql
```diff
@@ -1,2301 +1,2326 @@
...
+updatePlanItTools(id: UUID!, changes:PlanITToolsChanges!): PlanITTools!
+@hasRole(role: MINT_USER)
+
+updatePlanOpsEvalAndLearning(id: UUID!, changes: PlanOpsEvalAndLearningChanges!): PlanOpsEvalAndLearning!
+@hasRole(role: MINT_USER)
+
+uploadNewPlanDocument(input: PlanDocumentInput!): PlanDocument!
+@hasRole(role: MINT_USER)
+
+deletePlanDocument(id: UUID!): Int!
```
ClayBenson94, line 1535: We should probably add functionality to this mutation to delete the plan document links as well.

pkg/graph/schema.graphql
```diff
@@ -1,2301 +1,2326 @@
...
+  foreignKey: UUID
+  tableName: String!
+  action: String!
+  fields: Map!
+  modifiedBy: String
+  modifiedDts: Time
+}
+
+
+type OperationalSolution {
```
ClayBenson94, line 1423: Could we introduce a `documents: [PlanDocument!]!` type here instead of having a discrete query?

ClayBenson94 (changes requested):

pkg/storage/SQL/plan_document_solution_links_create.sql
```diff
@@ -0,0 +1,25 @@
+INSERT INTO plan_document_solution_link (
```
StevenWadeOddball, line 1: I think we can update this to take an array vs doing this individually for each link

pkg/graph/schema.graphql
```diff
@@ -1489,7 +1482,8 @@ type Query {
   operationalNeed(id: UUID!): OperationalNeed!
   auditChanges(tableName: String!, primaryKey: UUID!): [AuditChange!]!
   possibleOperationalNeeds: [PossibleOperationalNeed!]!
-  planDocumentSolutionLinks(modelPlanID: UUID!): [PlanDocumentSolutionLink!]!
+  possibleOperationalSolutions: [PossibleOperationalSolution!]!
+  planDocumentSolutionLinks(solutionID: UUID!): [PlanDocumentSolutionLink!]!
```
StevenWadeOddball, line 1486: I'm not sure `planDocumentSolutionLinks` is needed on the query directly, if we are returning documents from the solution object.

MINT.postman_collection.json
```diff
@@ -1548,6 +1548,34 @@
 						}
 					},
 					"response": []
+				},
+				{
+					"name": "CreatePlanDocumentSolutionLinks",
```
StevenWadeOddball, line 1553: Can you also add the remove link to the collection please?

StevenWadeOddball (commented):
Thanks for making these changes Tom, I'm having issues testing because of the Minio envrc change ( I can't upload any documents :/) I wanted to provide some quick feedback for a couple of things, otherwise it's looking like I'd expect! I just want to make sure and test it

pkg/storage/SQL/plan_document_read_by_solution_id_not_restricted.sql
```diff
@@ -0,0 +1,23 @@
...
+    doc.other_type,
+    doc.optional_notes,
+    doc.deleted_at,
+    doc.created_by,
+    doc.created_dts,
+    doc.modified_by,
+    doc.modified_dts
+FROM plan_document AS doc
+INNER JOIN plan_document_solution_link AS link ON link.document_id = doc.id
+INNER JOIN operational_solution AS sol ON sol.id = link.id
```
StevenWadeOddball, line 22: I think this is what breaks the resolver. I think it should be handled like it is in the other  SQL file. I tested it by querying from my local SQL tool. 
```suggestion
FROM plan_document AS doc
INNER JOIN plan_document_solution_link AS link ON link.document_id = doc.id
WHERE link.solution_id = :solution_id AND doc.restricted = FALSE
```

StevenWadeOddball (changes requested):
Overall working good, there are some issues just with Plan Document Read by Solution ID that should be resolved.

Besides that just the parameter naming where solution id is still called model plan id, and it would be good to have an example of delete document_solution_link for the postman collection.

pkg/storage/plan_documentStore.go
```diff
@@ -107,6 +116,32 @@ func (s *Store) PlanDocumentsReadByModelPlanID(
...
+	modelPlanID uuid.UUID,
+	s3Client *upload.S3Client) ([]*models.PlanDocument, error) {
+
+	statement, err := s.db.PrepareNamed(planDocumentsGetBySolutionIDSQL)
+	if err != nil {
+		return nil, err
+	}
+
+	var documents []*models.PlanDocument
+	err = statement.Select(&documents, utilitySQL.CreateSolutionIDQueryMap(modelPlanID))
```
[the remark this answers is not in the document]
OddTomBrooks: Yes, good call!

pkg/graph/schema.graphql
OddTomBrooks: Hmm, at a glance I agree. This was from before we made the modification to add documents[] to the operational solution. Let me think about it a little then probably I'll remove it.

pkg/storage/SQL/plan_document_read_by_solution_id_not_restricted.sql
OddTomBrooks: Yes, I had a batch of changes to push up

cmd/dbseed/main.go
```diff
@@ -154,8 +156,8 @@ func seedData(config *viper.Viper) {
 
 	// Seed a plan with some documents
 	planWithDocuments := createModelPlan(store, logger, "Plan with Documents", "MINT")
-	_ = planDocumentCreate(store, logger, s3Client, planWithDocuments, "File (Unscanned)", "cmd/dbseed/data/sample.pdf", "application/pdf", models.DocumentTypeConceptPaper, false, nil, nil, false, false)
-	_ = planDocumentCreate(store, logger, s3Client, planWithDocuments, "File (Scanned - No Virus)", "cmd/dbseed/data/sample.pdf", "application/pdf", models.DocumentTypeMarketResearch, false, nil, zero.StringFrom("Company Presentation").Ptr(), true, false)
+	restrictedDocument := planDocumentCreate(store, logger, s3Client, planWithDocuments, "File (Unscanned)", "cmd/dbseed/data/sample.pdf", "application/pdf", models.DocumentTypeConceptPaper, true, nil, nil, false, false)
```
StevenWadeOddball, line 159: I think this is a little problematic, the seeding works, but it's creating data we wouldn't expect or want in reality. 

EG, you are linking a document that belongs to another model to a solution for a different model. 

I think it's ok that the resolver doesn't verify that  the document already belongs to the same model as the solution for now, as that should be controlled by the Front End (though it could be something we add in the future to harden the backend).

Could you update the seed data please so the need and solution are added onto `planWithDocuments` instead of `sampleModelPlan` please? Alternatively, you could make another model plan to demonstrate the document solution link.

StevenWadeOddball (changes requested):
Overall it looks good and I can see the document endpoint is correctly verifying if the user is a collaborator or not, and I can remove a link after it was added! 

I just noticed the seed data doesn't represent how links should behave in the real world. To not make the scope of the ticket expand too much, I don't think we should try to enforce that in this ticket, but would you mind updating the seed data?

cmd/dbseed/main.go
OddTomBrooks: done 👍

StevenWadeOddball (approved):
Thanks for making those changes Tom! This looks good!

PR #308, opened 2022-11-16
[NOREF] - Only submit dirty fields when sending mutations on forms

patrickseguraoddball:
NOREF

## Changes and Description

- Updated all forms to only pass dirty fields to update mutations

Currently, for IT Solutions, needs questions were being marked as 'Answered' because each form was sending the entire payload even if a field hadn't been touched.  Not we can have certainty over fields that have truly been answered.

## Notes

Formik does indeed have  a `dirty` value, but not per input - only per form.  This PR created a util function to return only dirty fields/values to mutation

## How to test this change

- Submit forms for question pertaining to IT Solutions without actually answering the question
- Verify on IT Solutions, that the relevant need still says `Not Answered`

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/src/utils/formDiff.ts
+++ b/src/utils/formDiff.ts
@@ -1,0 +1,24 @@
+/* 
+Util used to return only dirty formik fields
+Formik has no dirty value per input, only on whole form
+This returns a payload with only values that have changes
+*/
+
+type DirtyInputType = {
+  [key: string]: any;
+};
+
+export const dirtyInput = (initialValues: any, values: any) => {
+  if (!initialValues || !values) return {};
+
+  const onlyDirtyInput: DirtyInputType = {};
+
+  Object.keys(initialValues).forEach(field => {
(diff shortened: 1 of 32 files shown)


StevenWadeOddball:
Thanks Patrick! I tested by clicking through task list sections without submitting any answer. At a quick pass, this worked for most fields (notably the ones you showed me yesterday). I did notice 2 needs still get populated. I wonder if those have an additional complexity going on?

I looked in the table and I don't see them marked as needed, are they hard coded examples? They look like they are in there before you actually start changing any fields (right after starting a new model plan)

image

patrickseguraoddball:
Thanks Steven!  Yep previously those needs were independent of answers and were always a need.  I synced with Natasha, and this has changed, and will be included as a part of my current PR, so no need to worry about those.

StevenWadeOddball (approved):
Looks great form a functional view Patrick! Thanks so much for making this!

PR #310, opened 2022-11-17
[EASI-2207] Introduce WebSocket authentication for Okta.

ClayBenson94:
# EASI-2207

## Changes and Description

- Introduces functionality to allow Okta users to authenticate WebSocket connections

## How to test this change

- Log in with Okta authentication on two browsers (same account)
- Create a model plan
- Test out task list locking between the two browsers.

## PR Author Review Checklist

## PR Reviewer Guidelines


StevenWadeOddball (approved):
Thanks Clay, I tested this locally by signing in with multiple browsers as the same user, and as different users. I was able to confirm the task list locking behavior works with a mix of OKTA and local.

I did have a couple instances where I noticed a message was missed from one browser or another, but I think that was related to existing transient issues, not introduced by this PR. I suspect some of those issues I observed were due to 
a. duplicating tabs on a locked section, and then navigating away from the section
b. letting a session window go to sleep when I was working on something else for an extended period time.

Regardless, I think that we are good to merge this change in as is. Thanks for making this happen!

PR #326, opened 2022-12-01
[NOREF] - Organize SQL Queries

rileyanderson:
## Changes and Description

- Move SQL queries to folders
- structured in `SQL/{table_name}/{query}.sql` in a few cases it is structured from the name of the `*_Store.go` file where it makes more sense to do so (e.g. `SQL/access_controll/**`)

## PR Author Review Checklist

## PR Reviewer Guidelines


pkg/storage/SQL/check_if_collaborator_by_operational_need_id.sql
```diff
@@ -1,7 +0,0 @@
-SELECT EXISTS(
```
StevenWadeOddball, line 1: Why was this deleted?

rileyanderson: It wasn't used anywhere

StevenWadeOddball: I think it can be left for now, we are going to need it fairly soon (for operational solution access control).

rileyanderson: Added it back

ClayBenson94 (approved):
This looks good! Thanks for running through all these - should help when we introduce more .sql files as well 😄

PR #327, opened 2022-12-01
[NOREF] Use 'mint-groups' instead of 'groups' when parsing job codes on Frontend

ClayBenson94:
# NOREF

## Changes and Description

- Modify the key checked in the JWT from `groups` to `mint-groups` to reflect what actually gets returned from Okta

The backend change was made in this PR (https://github.com/CMSgov/mint-app/pull/300), and the corresponding frontend change was just missed until now!

## How to test this change

- Log in with Okta
- Navigate to http://localhost:3005/user-diagnostics
- Ensure that the right job codes are parsed!

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/src/views/UserInfoWrapper/index.tsx
+++ b/src/views/UserInfoWrapper/index.tsx
@@ -15,7 +15,7 @@ type UserInfoWrapperProps = {
 type oktaUserProps = {
   name?: string;
   euaId?: string;
-  groups?: string[];
+  'mint-groups'?: string[];
 };
 
 const UserInfoWrapper = ({ children }: UserInfoWrapperProps) => {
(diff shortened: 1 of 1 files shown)


src/views/UserInfoWrapper/index.tsx
```diff
@@ -34,7 +34,7 @@ const UserInfoWrapper = ({ children }: UserInfoWrapperProps) => {
       const user = {
         name: oktaUser.name,
         euaId: oktaUser.euaId || '',
-        groups: oktaUser.groups || [],
+        groups: oktaUser['mint-groups'] || [],
```
patrickseguraoddball, line 37: The local `getUser` method on `oktaAuth` returns `groups`, the `authState` returns `mint-groups`.

patrickseguraoddball (changes requested):

patrickseguraoddball:
Okta sign in works to propagate the roles/groups, just the local login doesn't (should still use `groups`)

patrickseguraoddball (approved):
Thanks Clay!

PR #338, opened 2022-12-07
[EASI-2491] - Change Job Codes Depending on Environment

OddTomBrooks:
# EASI-2491

## Changes and Description

- Okta middleware switches job codes depending on if the user is in a production environment or pre-production environment

## How to test this change

TBD

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/pkg/okta/authentication_middleware.go
+++ b/pkg/okta/authentication_middleware.go
@@ -19,16 +19,73 @@ import (
 	"github.com/cmsgov/mint-app/pkg/userhelpers"
 )
 
+// These job codes define the environmentally specific potential codes which a user may possess
 const (
-	jobCodeUser       = "MINT_USER_NONPROD"
-	jobCodeAssessment = "MINT_ASSESSMENT_NONPROD"
-	jobCodeMAC        = "MINT MAC Users"
+	JobCodeTestUser       = "MINT_USER_NONPROD"
+	JobCodeTestAssessment = "MINT_ASSESSMENT_NONPROD"
+	JobCodeTestMACUser    = "MINT MAC Users"
+	JobCodeProdUser       = "MINT_USER"
+	JobCodeProdAssessment = "MINT_ASSESSMENT"
+	JobCodeProdMACUser    = "MINT_MAC_USER"
 )
 
 // EnhancedJwt is the JWT and the auth token
 type EnhancedJwt struct {
 	JWT       *jwtverifier.Jwt
 	AuthToken string
+}
+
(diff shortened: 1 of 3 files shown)


pkg/okta/authentication_middleware.go
```diff
@@ -19,10 +19,14 @@ import (
...
 const (
-	jobCodeUser       = "MINT_USER_NONPROD"
-	jobCodeAssessment = "MINT_ASSESSMENT_NONPROD"
-	jobCodeMAC        = "MINT MAC Users"
+	JobCodeTestUser       = "MINT_USER_NONPROD"
+	JobCodeTestAssessment = "MINT_ASSESSMENT_NONPROD"
+	JobCodeTestMACUser    = "MINT MAC Users"
+	JobCodeProdUser       = "MINT_USER"
+	JobCodeProdAssessment = "MINT_ASSESSMENT"
+	JobCodeProdMACUser    = "MINT_MAC_USER"
```
ClayBenson94, line 29: I might need to confirm what this looks like across different environments. It's possible that this will be the same across prod/nonprod, but we should probably keep different const values for them just in case.

I should have some updates by tomorrow on what this value might look like, but have a feeling it'll also be `MINT MAC Users`

pkg/server/routes.go
```diff
@@ -90,10 +90,13 @@ func (s *Server) routes(
 	oktaConfig := s.NewOktaClientConfig()
 	jwtVerifier := okta.NewJwtVerifier(oktaConfig.OktaClientID, oktaConfig.OktaIssuer)
 
+	isLocalOrTestEnvironment := s.environment.Local() || s.environment.Test()
+
 	oktaMiddlewareFactory := okta.NewMiddlewareFactory(
 		handlers.NewHandlerBase(s.logger),
 		jwtVerifier,
 		store,
+		isLocalOrTestEnvironment,
```
ClayBenson94, line 99: I am thinking that using `!s.environment.Prod()` here might be more succinct because we probably want to check these `_NONPROD` codes in the case of `s.environment.Impl()` as well

ClayBenson94 (changes requested):

pkg/okta/authentication_middleware.go
OddTomBrooks: Updated, ping me with the final codes when you have them and I'll update 👍

pkg/server/routes.go
OddTomBrooks: Updated 👍

PR #344, opened 2022-12-08
[NOREF] Modify eslint to allow no warnings

ClayBenson94:
# NOREF

## Changes and Description

- Modify `eslint` to pass `--max-warnings=0`. This makes it behave just like `yarn run build`, which sets `CI=true` (which makes it so `eslint` warnings are treated as errors)

This change makes it so that eslint warnings are treated as errors, which would (before this change) not be caught until we run the production build during the deploy. Now, these errors should get caught in CI!

## How to test this change

Introduce an `eslint` warning into the code base, and attempt to `git commit` it!

## PR Author Review Checklist

## PR Reviewer Guidelines


OddTomBrooks (approved):

patrickseguraoddball (approved):
Works as expected, thanks Clay!

PR #345, opened 2022-12-08
[NOREF] Use default logout URI

ClayBenson94:
# NOREF

## Changes and Description

- Use default `post_logout_redirect_uri` when logging users out via a session timeout

Not 100% sure why, but depending on _how_ a user logged out (session timeout vs. clicking "Sign Out") we'd send the user to a different URI. This was annoying because it wasn't consistent, and was problematic because the IDM team only has `/` as a valid redirect URI (rather than `/` and `/signin`)

## How to test this change

Sadly, only `dev.mint.cms.gov` and `test.mint.cms.gov` are valid redirect URIs, not `localhost:3005/`, so testing this locally isn't possible yet, but ensuring that when you get a session timeout the `post_logout_redirect_uri` is what you expect it to be in the network console is sufficient!

## PR Author Review Checklist

## PR Reviewer Guidelines


garyjzhao (approved):
LGTM

PR #347, opened 2022-12-09
[NOREF] Create "testing" environment instead of using "test"

ClayBenson94:
# NOREF

## Changes and Description

- Changed references that used `test` as an environment to use `testing` instead
- Created new `testing` environment in `appconfig` package to facilitate identifying when we're running tests (what used to be `test`)

We needed to have some modifications to our environment configuration because EASi historically (and currently) only had `dev`, `impl`, and `prod` environments. Therefore, an environment was created in the appconfig called `test` which was used to identify when the application was being run for local test (i.e. `go test` or `cypress`)

However, MINT has an environment literally named `test`, so the "unit testing" environment was renamed `testing`, and all instances of `test` now refer to the _deployed_ test env, rather than the "unit testing" environment.

## How to test this change

- Ensure that when `APP_ENV` is set to `test`, a real connection to CEDAR LDAP is used.
- Ensure that when `APP_ENV` is set to `testing`, the CEDAR LDAP mock is used.

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -145,7 +145,7 @@ jobs:
         env:
           REACT_APP_API_ADDRESS: http://easi:8080/api/v1
           REACT_APP_GRAPHQL_ADDRESS: http://easi:8080/api/graph/query
-          REACT_APP_APP_ENV: test
+          REACT_APP_APP_ENV: testing
           REACT_APP_OKTA_CLIENT_ID: 0oad9awvebnsMwWNa297
           REACT_APP_OKTA_DOMAIN: https://test.idp.idm.cms.gov
           REACT_APP_OKTA_SERVER_ID: ausd980kt2CBBzStG297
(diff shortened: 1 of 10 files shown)


patrickseguraoddball:
Looks good from my end.  I only noticed one possible discrepancy - should we be including the local default for Faktory in testing as well?

https://github.com/CMSgov/mint-app/blob/45aaa6590de8cd47f16dbbc897e65813374fb136/pkg/server/routes.go#L254-L256

patrickseguraoddball (approved):

rileyanderson:
> > Looks good from my end. I only noticed one possible discrepancy - should we be including the local default for Faktory in testing as well?
> > https://github.com/CMSgov/mint-app/blob/45aaa6590de8cd47f16dbbc897e65813374fb136/pkg/server/routes.go#L254-L256
> 
> @patrickseguraoddball Good callout - @rileyanderson, do you know off-hand if we have any tests that rely on Faktory? Looks like the `server_test` job succeeded, so my guess is that we don't currently have anything, but maybe some of your current work might be introducing something like that?
> 
> EDIT: Also realizing - this code was created mainly to avoid Faktory from running in deployed envs (since it was crashing the app on startup)
> 
> We might just be able to change this to something like `if !s.environment.Deployed()`, that way local execution still works as expected.

@ClayBenson94 Not yet currently. The ticket I am working on now *may* introduce it but tbh i don't think we want to run Faktory processes in our tests. We mainly just want to test the functions that Faktory will call

rileyanderson (dismissed):

OddTomBrooks (approved):
LGTM

rileyanderson (changes requested):
Actually, it is kinda confusing to have a `test` and `testing` env. Should we re-name the deployed one to `staging` or something? https://github.com/CMSgov/mint-app/pull/350#pullrequestreview-1213860451

rileyanderson (approved):

PR #352, opened 2022-12-12
[EASI-2374] Update Status for Operational Needs

garyjzhao:
# EASI-2374

## Changes and Description

- Add new copy
- Add `isUpdatingStatus` prop to `SoltuionImplmentation` component and the corresponding conditional rendered copy
- Add new route for Update Status for the each operational need 
    - "/models/:modelID/task-list/it-solutions/:operationalNeedID/update-status"`
- Update `util.test.tsx`
- Pulled in `need id` in `util.tsx` Thanks @patrickseguraoddball 

## How to test this change

1. Pre-seed and create some operational needs 
2. Select a solution 
3. Click `Update status` and ensure you are navigated to the correct page
4. Update the status (dates and status)
5. Ensure data is saved and displayed properly

## PR Author Review Checklist

## PR Reviewer Guidelines


patrickseguraoddball (changes requested):
I think the route redirection after updating needs to accommodate for updating status and returning to the last page, not necessarily IT Solutions tracker.  For instance on my current branch I can update status from solution details.

The direction here could go something like: 

https://github.com/CMSgov/mint-app/blob/d28bedf9fee333357f7ea06251640a9a66252c05/src/views/ModelPlan/TaskList/ITSolutions/SolutionImplementation/index.tsx#L175

```
if (isUpdatingStatus) {
    history.goBack();
else {
    history.push(`/models/${modelID}/task-list/it-solutions`);
}
```

Because you can update the status from multiple places I think goBack fits better here.  I see in the Figma though, the link says 'Don’t update and return to solution details' even though you could be directed from the IT solutions tracker.

src/views/ModelPlan/TaskList/ITSolutions/SolutionImplementation/index.tsx
```diff
@@ -167,14 +171,26 @@ const SolutionImplementation = () => {
...
+            showMessageOnNextPage(
+              <Alert type="success" slim className="margin-y-4">
+                <span className="mandatory-fields-alert__text">
+                  {t('successStatusUpdated', {
+                    operationalNeedName: operationalNeed.name
+                  })}
+                </span>
+              </Alert>
+            );
+            history.goBack();
```
garyjzhao, line 192: @patrickseguraoddball, I made the conditional check and routing more verbose in order to incorporate a different success message when updating the status.

src/views/ModelPlan/TaskList/ITSolutions/SolutionImplementation/index.tsx
```diff
@@ -506,7 +532,9 @@ const SolutionImplementation = () => {
                               className="margin-right-1"
                               aria-hidden
                             />
-                            {t('dontAdd')}
+                            {isUpdatingStatus
```
patrickseguraoddball, line 535: This submit handler on this button needs to not be called if choosing to not update

```
onClick={() => {
    if (!isUpdatingStatus) handleFormSubmit(values, null, true);
  }}
```

The mutation needs to only be called on not adding, because that actually removed the solution from the need.  When updating, we still want the solution, but don't need to call any mutations

patrickseguraoddball (changes requested):
I think there still may need some further work on how adaptable we make this component.  This update page will be called from a a few different places, and we need to reflect that in the breadcrumbs, and well as the bottom-most link to return to previous location.  I think this could be in the scope of the work either by passing various hashes, or passing a variable through location state.  We can sync on implementation

src/views/ModelPlan/TaskList/ITSolutions/SolutionImplementation/index.tsx
```diff
@@ -157,7 +161,7 @@ const SolutionImplementation = () => {
 
         if (response && !errors) {
           // If successfully submitting solution details
-          if (!dontAdd && !redirect) {
+          if (!dontAdd && !redirect && !isUpdatingStatus) {
```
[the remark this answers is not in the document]
garyjzhao: @patrickseguraoddball the success messaging is a bit different when the user updates the status. I adjusted the code to better DRY/streamline the conditonals

src/views/ModelPlan/TaskList/ITSolutions/SolutionImplementation/index.tsx
garyjzhao: I took your suggestion and added another conditional to allow user to go back (behavior that we want) when the click. As your code sits, nothing will happen if user clicks and `!isUpdatingStatus` 
```js
if (isUpdatingStatus) {
  history.goBack();
} else {
  handleFormSubmit(values, null, true);
}
```

src/views/ModelPlan/TaskList/ITSolutions/index.tsx
```diff
@@ -58,10 +58,10 @@ const ITSolutions = () => {
             </Route>
 
             <Route
-              path="/models/:modelID/task-list/it-solutions/:operationalNeedID/update-status"
+              path="/models/:modelID/task-list/it-solutions/:operationalNeedID/update-status/:solutionId"
```
patrickseguraoddball, line 61: Within the tracker it's possible to update all solutions at once.  We need the ability to have the `solutionId` optional in this case.  If there is no solution id, and we are updating, we need to render all the solutions and have the text reflective of the update.

The flow to update all the solutions is as follows: 
From the tracker, click 'View Details', then in the solution details, click 'Update status and timing'.  Then you will comfirm the solutions on the next page and click 'Continue'.  Then you are taken to the update-status view where you should be able to update all the solutions

src/views/ModelPlan/TaskList/ITSolutions/SolutionImplementation/index.tsx
```diff
@@ -189,32 +200,82 @@ const SolutionImplementation = () => {
...
-            <span>{t('breadcrumb')}</span>
-          </BreadcrumbLink>
-        </Breadcrumb>
-        <Breadcrumb current>{t('selectSolution')}</Breadcrumb>
-      </BreadcrumbBar>
+      {fromSolutionDetails ? (
+        <BreadcrumbBar variant="wrap">
+          <Breadcrumb>
+            <BreadcrumbLink asCustom={UswdsReactLink} to="/">
+              <span>{h('home')}</span>
```
patrickseguraoddball, line 219: I think this can be refactored a bit.  We may want to look how EASI has created a much more condensed version of breadcrumbs.

src/views/ModelPlan/TaskList/ITSolutions/SolutionImplementation/index.tsx
```diff
@@ -319,158 +384,57 @@ const SolutionImplementation = () => {
...
+                                      handleOnBlur={handleOnBlur}
+                                      setFieldValue={setFieldValue}
+                                      isUpdatingStatus={isUpdatingStatus}
+                                      operationalNeed={operationalNeed}
+                                      values={values}
+                                      flatErrors={flatErrors}
+                                      loading={loading}
+                                    />
+                                  );
+                                })
```
patrickseguraoddball, line 412: This conditional is a little redundant.  You should map over this only once.  The solutions to render should all be in an array.  If updating one, it should map over [solution].  If over all the solutions, map over them normally.   If this doesn't make sense lets pair!

This filtering needs to happen outside of the formik context, because on the submission, it is updating all the solutions, when we are only updating one.  It is having unwanted mutations on solutions that arent being updated.

src/views/ModelPlan/TaskList/ITSolutions/SolutionImplementation/_components/Solution.tsx
```diff
@@ -0,0 +1,173 @@
...
+import ImplementationStatuses from '../../_components/ImplementationStatus';
+import SolutionCard from '../../_components/SolutionCard';
+
+type flatErrorsType = {
+  [key: string]: string;
+};
+
+type SolutionTypes = {
+  solution: OpertionalNeedSolutionTypes;
+  // flatErrors: [key: string]: string;
```
patrickseguraoddball, line 25: This might be old commented out code that can be removed

src/views/ModelPlan/TaskList/ITSolutions/SolutionImplementation/_components/Solution.tsx
```diff
@@ -0,0 +1,173 @@
...
+            </Label>
+
+            <div className="usa-hint margin-top-1">{h('datePlaceholder')}</div>
+
+            <FieldErrorMsg>{flatErrors.mustFinishDts}</FieldErrorMsg>
+
+            <div className="width-card-lg position-relative">
+              <Field
+                as={DatePicker}
+                error={!!flatErrors.mustFinishDts}
```
patrickseguraoddball, line 114: For both of these date errors, browser console throwing warning about an incorrect type.  The boolean here sohuld be expressed as:

```
error={+!!flatErrors.mustFinishDts}
```

patrickseguraoddball (changes requested):
Nice Gary, a few more changes.  We can pair on some as I think it may be easier to talk through.  

Also, I merged in my `SolutionDetailCard` component, so I think it's best to use that to display the combined solution on the update status page.  I know we said I would take care of this, but wasn't anticipating my PR to be merged first.  We can also sync on this!

patrickseguraoddball:
When on the update status page after navigating to update all solutions this link says it will take me back to the tracker, but it takes me back to the solution page:

Screenshot 2023-01-06 at 1 11 05 PM

patrickseguraoddball (changes requested):
The links to update and return to tracker dont work as expected - describe by the recent comment

src/views/ModelPlan/TaskList/ITSolutions/SolutionImplementation/index.tsx
```diff
@@ -189,32 +197,53 @@ const SolutionImplementation = () => {
...
+    { text: isUpdatingStatus ? t('updateStatus') : t('selectSolution') }
+  ];
+
+  const solutionsArray =
+    solutionId === undefined
+      ? operationalNeed.solutions
+      : operationalNeed.solutions.filter(
+          solution => solution.id === solutionId
+        );
+
```
patrickseguraoddball, line 237: This filtering on the solutions need to be passed to the Formik context.  All the solutions are still being passed to to Formik and attempting to update all solutions on the mutation.  `operationalNeed` is being passed, along with nested solutions on line 285.  The filtering needs to happen on this variable.  

```
const formikNeed = {...operationalNeed}
  if (operationalNeedID) {
    formikNeed.solutions = formikNeed.solutions.filter(solution => solution.id === solutionId)
  }

or

formikNeed.solutions = solutionsArray;
```

patrickseguraoddball (changes requested):

patrickseguraoddball (approved):
Nice work Gary, thanks!

PR #354, opened 2022-12-14
Read Only Operations Evaluation and Learning

downeyn-cms:
# EASI-2223

## Changes and Description

- Add GraphQL query for all OpsEvalAndLearning data
- Create ReadOnly component for OpsEvalAndLearning data and add it to ReadOnly ModelPlan component

## PR Author Review Checklist

## PR Reviewer Guidelines


patrickseguraoddball:
This status tag gets crunched a little bit.  I think adding `flex-2` util class on the adjacent <h2> element would give the tag the room it needs.

Screenshot 2022-12-14 at 11 09 50 AM

patrickseguraoddball (changes requested):
Thanks Nick!.  Just a few things in addition to comments.  There are a few other areas that are split into two columns other than misc info.  You should be able to reference other read only section to see how these are implemented when you are referencing Figma

src/i18n/en-US/draftModelPlan/operationsEvaulationAndLearning.ts
```diff
@@ -186,7 +186,7 @@ const operationsEvaluationAndLearning = {
     beneficiary: 'Beneficiary-level data',
     participant: 'Participant-level data',
     provider: 'Provider-level data',
-    other: 'Other-MIPS data',
```
downeyn-cms, line 189: I modified this from 'Other-MIPS data' to 'Other'. I did this to make it match the rest of the "Other" options across the forms and so that it would work properly with the ReadOnlySection component.

I am not sure if this was 'Other-MIps data' for a specific reason or if i was ok to change it.

@patrickseguraoddball @garyjzhao - we should probably bring this up with Natasha on Monday. I am happy to change it back and make modifications to make it work with ReadOnlySection, that wouldn't be too difficult.

garyjzhao (approved):
Looks great @downeyn-cms

PR #355, opened 2022-12-14
[NOREF] - User initial updates

patrickseguraoddball:
NOREF

## Changes and Description

User initials in icons were not displaying correctly with extra spaces and characters, ex: "John Doe (He/They)"

- Updated user initial icon to include first valid letter character of each word separated by a space character
- Unit tests

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/src/utils/modelPlan.test.tsx
+++ b/src/utils/modelPlan.test.tsx
@@ -1,0 +1,16 @@
+import { getUserInitials, returnValidLetter } from './modelPlan';
+
+describe('model plan util', () => {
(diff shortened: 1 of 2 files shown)


ClayBenson94 (approved):
Thanks for adding the unit tests in there - makes it so much easier to see what the behavior will be like in different cases!

I'm sure we'll continue to catch funny cases on this from LDAP data on users, but I think this is the right fix for now, given what we've seen!

PR #372, opened 2022-12-20
[Easi-2533] Multiple Role Directive and Assign MAC Role to Mutations 

OddTomBrooks:
# [Easi-2533] Multiple Role Directive and Assign MAC Role to Mutations 

## Changes and Description

- A new directive with support for multiple roles is created
- The MAC role is assigned to appropriate GraphQL mutations
  - agreeToNDA

## How to test this change

- Call the agreeToNDA mutation with the following job codes and results:
  - Job Code "MINT_USER_NONPROD" - Access
  - Job Code "MINT MAC Users" - Access
  - Job Code "MINT_ASSESSMENT" - No Access
 

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/pkg/services/auth.go
+++ b/pkg/services/auth.go
@@ -15,7 +15,7 @@ func HasAnyRole(ctx context.Context, roles []model.Role) (bool, error) {
 	for _, role := range roles {
 		userHasRole, err := HasRole(ctx, role)
 		if err != nil {
-			return userHasRole, err
+			return false, err
 		}
 
 		if userHasRole {
--- a/pkg/services/auth.go
+++ b/pkg/services/auth.go
@@ -10,6 +10,22 @@ import (
 	"github.com/cmsgov/mint-app/pkg/graph/model"
 )
 
+// HasAnyRole authorizes a user as having any in a collection of given roles
+func HasAnyRole(ctx context.Context, roles []model.Role) (bool, error) {
+	for _, role := range roles {
+		userHasRole, err := HasRole(ctx, role)
+		if err != nil {
+			return userHasRole, err
+		}
+
+		if userHasRole {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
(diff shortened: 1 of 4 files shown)


pkg/services/auth.go
```diff
@@ -10,6 +10,22 @@ import (
 	"github.com/cmsgov/mint-app/pkg/graph/model"
 )
 
+// HasAnyRole authorizes a user as having any in a collection of given roles
+func HasAnyRole(ctx context.Context, roles []model.Role) (bool, error) {
+	for _, role := range roles {
+		userHasRole, err := HasRole(ctx, role)
+		if err != nil {
+			return userHasRole, err
```
ClayBenson94, line 18:
```suggestion
			return false, err
```

Wondering if this might be safer - if `HasRole()` returns an error, it might be safe to assume that we shouldn't grant any roles (though, practically, it doesn't look like `HasRole()` ever actually returns an error... maybe we should just nix the `error` return from both of these functions and simplify them?

ClayBenson94 (changes requested):
These changes look great!

One small change, I'd say we should either
1. Implement the suggestion that handles errors by returning `false, err`
2. Simplify `HasAnyRole` and `HasRole` functions to not return errors at all, and just simply return a `bool`

@OddTomBrooks What are your thoughts?

pkg/services/auth.go
OddTomBrooks: Even if there is no current error implementation I think it's best to preserve the error state so that any systems using this method will already handle future error implementations.

I am on the fence about returning values on error states, I seem to recall we do both in different areas of the code base. In this case I agree that returning false is likely more rational. It would be good to have a discussion and project-scope decision on how we handle returning captured values vs default values on errors.

ClayBenson94 (approved):
Changes look good! Tested locally and seems to be working as expected!

PR #374, opened 2022-12-21
[EASI-2435] Upgrade flyway to v9

ClayBenson94:
# EASI-2435

## Changes and Description

- Upgrade flyway to V9 in Dockerfiles for DB Clean and DB Migrate
- Upgrade flyway to V9 by upgrading `joshuaavalon/flyway-action` to v3.0.0 (https://github.com/joshuaavalon/flyway-action/releases/tag/v3.0.0)

### Upgrade of the flyway-action

- Ensure that the CI/CD jobs pass, which use the flyway-action to run migrations!

### Upgrade in the `Dockerfile`s
- Start the service with `scripts/dev up`
  - This inherently runs the `Dockerfile.db_migrations` file due to [this entry](https://github.com/CMSgov/mint-app/blob/8401c8c2a62a26a15e236051ebb05755b3911851/docker-compose.yml#L9-L18) in our docker-compose.yml file.
- Ensure DB migrations ran as expected
- Use this nasty one-liner to build and run the DB Clean image:
```bash
FLYWAY_CLEAN_DISABLED=false FLYWAY_USER=postgres FLYWAY_PASSWORD=mysecretpassword FLYWAY_URL=jdbc:postgresql://host.docker.internal:5437/postgres docker run --env FLYWAY_CLEAN_DISABLED --env FLYWAY_USER --env FLYWAY_PASSWORD --env FLYWAY_URL $(docker build -q -f Dockerfile.db_clean .)
```
- Ensure DB Clean ran as expected

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -71,7 +71,7 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Migrate Database
-        uses: joshuaavalon/flyway-action@v1
+        uses: joshuaavalon/flyway-action@v3.0.0 # flyway-action 3.0.0 uses flyway v9, which should (if we do our jobs) match what's in Dockerfile.db_clean and Dockerfile.db_migrations
         with:
           url: jdbc:postgresql://postgres:5432/${{ env.POSTGRES_DB }}
           user: ${{ env.POSTGRES_USER }}
--- a/Dockerfile.db_clean
+++ b/Dockerfile.db_clean
@@ -1,4 +1,4 @@
-FROM flyway/flyway:7.8.2-alpine
+FROM flyway/flyway:9.10.1-alpine
 
 ENV FLYWAY_CONNECT_RETRIES=10
 ENV FLYWAY_IGNORE_INVALID_MIGRATION_NAMES="false"
--- a/Dockerfile.db_clean
+++ b/Dockerfile.db_clean
@@ -1,4 +1,4 @@
-FROM flyway/flyway:9.10.1-alpine
+FROM flyway/flyway:9.10-alpine
 
 ENV FLYWAY_CONNECT_RETRIES=10
 ENV FLYWAY_IGNORE_INVALID_MIGRATION_NAMES="false"
--- a/Dockerfile.db_migrations
+++ b/Dockerfile.db_migrations
@@ -1,4 +1,4 @@
-FROM flyway/flyway:9.10.1-alpine
+FROM flyway/flyway:9.10-alpine
 
 COPY migrations/ /flyway/sql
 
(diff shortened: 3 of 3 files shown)


Dockerfile.db_clean
```diff
@@ -1,4 +1,4 @@
-FROM flyway/flyway:7.8.2-alpine
+FROM flyway/flyway:9.10.1-alpine
```
rileyanderson, line 1: Do we want to lock this to a patch version? `flyway/flyway:9.10` will pull the latest patch version. We _could_ also do the same for minor version: `flyway/flyway:9`. It's a bit more risky but it _should_ be backwards compatible

.github/workflows/run_tests.yml
```diff
@@ -71,7 +71,7 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Migrate Database
-        uses: joshuaavalon/flyway-action@v1
+        uses: joshuaavalon/flyway-action@v3.0.0 # flyway-action 3.0.0 uses flyway v9, which should (if we do our jobs) match what's in Dockerfile.db_clean and Dockerfile.db_migrations
```
rileyanderson, line 74: Same comment. idk if we want to lock this to a minor and major version unless we need to

Dockerfile.db_clean
ClayBenson94: I could go either way. I'd probably rather do patch, since I saw in the release notes for Flyway that sometimes breaking changes are released in Minor versions 😬

.github/workflows/run_tests.yml
ClayBenson94: The only thing I noted is that I think GitHub actions don't actually _know_ what SemVer is, they just work on tags. The only reason `v1` worked before is there's a `v1` tag on that repo. Sadly, they stopped doing that on that action and now only have `v3.0.0`

https://github.com/joshuaavalon/flyway-action/tags
https://docs.github.com/en/actions/creating-actions/about-custom-actions#using-tags-for-release-management

GitHub recommends making multiple tags for each version, and including a "major only" tag that moves as updates get pushed, but seems like this action doesn't follow that.

rileyanderson: Cool cool TIL. Didn't realize that action was tagged like that 👍

Dockerfile.db_clean
rileyanderson: I'm good with patch

ClayBenson94: Sweet. I'll update to [9.10-alpine](https://hub.docker.com/layers/flyway/flyway/9.10-alpine/images/sha256-07f2358b0f75c8bcdf1b07deb5853ccf21152bb012c261f0f60151128638b008?context=explore)

rileyanderson (approved):

StevenWadeOddball:
BTW, there is actually a tech debt ticket in the backlog for this! [EASI-2435](https://jiraent.cms.gov/browse/EASI-2435)

PR #381, opened 2022-12-29
Easi 2550/add user account collaborator

StevenWadeOddball:
# EASI-2550

## Changes and Description

- Add functionality to add a user account using CEDAR information when a collaborator is added
- Add hasLoggedIn bool to user account
   - If user hasn't logged in, update the information with OKTA instead at the time of login.
- Decorate the context with the JWT, so we we can use that for reaching OKTA endpoints 

## How to test this change
Note, this change should function for both local and CEDAR. It is worth while to check with both
To check Cedar locally, you need to comment out line 127-129 in routes and connect to the vpn
``` go
	if s.environment.Local() || s.environment.Testing() {
		cedarLDAPClient = local.NewCedarLdapClient(s.logger)
	}
```

1. Create model plan.
2. Add a collaborator to the MODEL PLAN
    a. Verify that a user account was added for this collaborator, and that hasLoggedIn == false
3. Log in as that user. (Or send any request in postman)
    a. Verify that the user account has been updated to hasLoggedIn == true, and any missing information is filled in. 
     

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/migrations/V2.5__Add_Users_table.sql
+++ b/migrations/V2.5__Add_Users_table.sql
@@ -7,7 +7,8 @@ CREATE TABLE user_account (
     email ZERO_STRING NOT NULL,
     given_name ZERO_STRING NOT NULL,
     family_name ZERO_STRING NOT NULL,
-    zone_info ZERO_STRING NOT NULL
+    zone_info ZERO_STRING NOT NULL,
+    has_logged_in BOOLEAN NOT NULL
 );
 
 ALTER TABLE user_account
--- a/migrations/V2.5__Add_Users_table.sql
+++ b/migrations/V2.5__Add_Users_table.sql
@@ -7,8 +7,7 @@ CREATE TABLE user_account (
     email ZERO_STRING NOT NULL,
     given_name ZERO_STRING NOT NULL,
     family_name ZERO_STRING NOT NULL,
-    zone_info ZERO_STRING NOT NULL,
-    has_logged_in BOOLEAN NOT NULL
+    zone_info ZERO_STRING NOT NULL
 );
 
 ALTER TABLE user_account
--- a/migrations/V41__Add_Has_Logged_In_To_User_Account.sql
+++ b/migrations/V41__Add_Has_Logged_In_To_User_Account.sql
@@ -1,10 +1,2 @@
 ALTER TABLE user_account
-ADD COLUMN has_logged_in BOOLEAN DEFAULT FALSE; --ADD Column
-
-
-UPDATE user_account --Update missing information
-SET has_logged_in = FALSE
-WHERE has_logged_in IS NULL;
-
-ALTER TABLE user_account
-ALTER COLUMN has_logged_in SET NOT NULL; -- Add not null constraint
+ADD COLUMN has_logged_in BOOLEAN NOT NULL DEFAULT FALSE;
--- a/migrations/V41__Add_Has_Logged_In_To_User_Account.sql
+++ b/migrations/V41__Add_Has_Logged_In_To_User_Account.sql
@@ -1,0 +1,10 @@
+ALTER TABLE user_account
+ADD COLUMN has_logged_in BOOLEAN DEFAULT FALSE; --ADD Column
+
+
+UPDATE user_account --Update missing information
+SET has_logged_in = FALSE
+WHERE has_logged_in IS NULL;
+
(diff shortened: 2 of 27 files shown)


pkg/okta/authentication_middleware.go
```diff
@@ -32,12 +32,6 @@ const (
 	JobCodeProdMACUser    = "MINT MAC Users"
 )
 
-// EnhancedJwt is the JWT and the auth token
-type EnhancedJwt struct {
```
StevenWadeOddball, line 36: moved to authentication to avoid bad import cycle

StevenWadeOddball:
Note, this also is currently just modifying a migration. Assuming that we can't clear migrations and run this, I will need to update the changes to an update statement instead.

migrations/V2.5__Add_Users_table.sql
```diff
@@ -7,7 +7,8 @@ CREATE TABLE user_account (
     email ZERO_STRING NOT NULL,
     given_name ZERO_STRING NOT NULL,
     family_name ZERO_STRING NOT NULL,
-    zone_info ZERO_STRING NOT NULL
+    zone_info ZERO_STRING NOT NULL,
+    has_logged_in BOOLEAN NOT NULL
```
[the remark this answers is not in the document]
ClayBenson94: Per our discussion, let's go with `2.` for now!

ClayBenson94 (changes requested):

migrations/V2.5__Add_Users_table.sql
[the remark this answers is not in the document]
StevenWadeOddball: Thanks Clay! I made the change :)

StevenWadeOddball:
Thanks for the feedback @ClayBenson94!

Per our conversations I've done the following
1. Update migrations to a new migration
2. Type aliased the delegate functions
3. Made a new type AccountInfo
    a. Wrapper function for OKTA account information that returns just account information.

migrations/V41__Add_Has_Logged_In_To_User_Account.sql
```diff
@@ -0,0 +1,10 @@
+ALTER TABLE user_account
+ADD COLUMN has_logged_in BOOLEAN DEFAULT FALSE; --ADD Column
+
+
+UPDATE user_account --Update missing information
+SET has_logged_in = FALSE
+WHERE has_logged_in IS NULL;
+
+ALTER TABLE user_account
+ALTER COLUMN has_logged_in SET NOT NULL; -- Add not null constraint
```
rileyanderson, line 10: You could do this which would avoid an unsafe data migration:
```suggestion
ALTER TABLE user_account
ADD COLUMN has_logged_in BOOLEAN NOT NULL DEFAULT FALSE;
```

rileyanderson (changes requested):

pkg/userhelpers/user_account_utils.go
```diff
@@ -21,59 +26,127 @@ type OktaAccountInfo struct {
...
+
+	updatedAccount, updateErr := store.UserAccountUpdateByUserName(userAccount)
+	if updateErr != nil {
+		return nil, updateErr
+	}
+	return updatedAccount, nil
+}
+
+// GetUserInfoAccountInformationWrapperFunction returns a function that returns *AccountInfo with the input of a function that returns UserInfo
+func GetUserInfoAccountInformationWrapperFunction(getAccountInformation GetUserInfoFunc) GetAccountInfoFunc {
```
rileyanderson, line 85: Could we rename these functions with `AccountInformation` to `GetUserInfoAccountInfoWrapperFunction` or even `GetUserInfoAccountInfoWrapperFunc` to keep the naming consistent or was there a reason you put `AccountInformation`?

StevenWadeOddball: Thanks Riley! Good suggestions, I'll change it.

migrations/V41__Add_Has_Logged_In_To_User_Account.sql
rileyanderson: Unsafe in the sense that rolling back migrations that contain data backfill is not really possible. Where rolling back an ADD COLUMN only will just delete it.

rileyanderson (approved):
🚀

PR #383, opened 2023-01-03
[NOREF] Re-enable IMPL deployments. Make dev/test deploy concurrently

ClayBenson94:
# NOREF

## Changes and Description

- Re-enable IMPL deploys, now that pen-testing is complete in the IMPL environment
- Make `dev` and `test` environments deploy concurrently after `run_tests`. 
  - In the case of pushing a bug to `main`, this _does_ mean we will push it to both `dev` and `test`, but since `dev` is really meant to be an internal dev-only environment, and `test` is where most feature validation will happen, it feels like a good way to save some time on the deploy process.

## How to test this change

- There's not a great way to test GitHub actions locally (except for something like [act](https://github.com/nektos/act)), so it's recommended to just review the changes made to this file and ensure they're doing what we expect!
- Also, if folks disagree or have questions about making dev/test deploy at the same time, I'm happy to change this! It's not set in stone 😄 

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/.github/workflows/deploy_pipeline.yml
+++ b/.github/workflows/deploy_pipeline.yml
@@ -42,7 +42,7 @@ jobs:
       ECR_REGISTRY: ${{ secrets.ECR_REGISTRY }}
 
   deploy_test:
-    needs: deploy_dev
+    needs: run_tests
     uses: ./.github/workflows/deploy_to_environment.yml
     with:
       env: test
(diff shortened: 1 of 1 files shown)


StevenWadeOddball (approved):
looks good to me Clay, I think it makes sense to deploy dev and test concurrently.

PR #388, opened 2023-01-06
[EASI-2521] Discussion and Discussion Reply return user info full stack

OddTomBrooks:
# EASI-2521

## Changes and Description

- Added UserInfo to Discussion
- Added UserInfo to DiscussionReply

## How to test this change

- Create a Discussion, observe that CreatedByUserInfo populates appropriately and ModifiedByUserInfo is nil
- Update a Discussion, observe that CreatedByUserInfo populates without change and ModifiedByUserInfo populates appropriately
- Create a DiscussionReply, observe that CreatedByUserInfo populates appropriately and ModifiedByUserInfo is nil
- Update a DiscussionReply, observe that CreatedByUserInfo populatese without change and ModifiedByUserInfo populates appropriately

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/pkg/models/plan_discussion.go
+++ b/pkg/models/plan_discussion.go
@@ -8,19 +8,30 @@ import (
 type PlanDiscussion struct {
 	baseStruct
 	modelPlanRelation
-	Content      string           `json:"content" db:"content"`
-	Status       DiscussionStatus `json:"status" db:"status"`
-	IsAssessment bool             `json:"isAssessment" db:"is_assessment"`
+	Content            string           `json:"content" db:"content"`
+	Status             DiscussionStatus `json:"status" db:"status"`
+	IsAssessment       bool             `json:"isAssessment" db:"is_assessment"`
+	CreatedByUserInfo  UserInfo         `json:"createdByUserInfo" db:"created_by_user_info"`
+	ModifiedByUserInfo *UserInfo        `json:"modifiedByUserInfo" db:"modified_by_user_info"`
 }
 
 // NewPlanDiscussion returns a New PlanDiscussion with a status of UNANSWERED
-func NewPlanDiscussion(principal string, isAssessment bool, modelPlanID uuid.UUID, content string) *PlanDiscussion {
+func NewPlanDiscussion(
+	principal string,
+	isAssessment bool,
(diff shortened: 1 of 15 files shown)


pkg/models/plan_discussion.go
```diff
@@ -8,37 +8,60 @@ import (
...
 	baseStruct
 	modelPlanRelation
-	Content      string           `json:"content" db:"content"`
-	Status       DiscussionStatus `json:"status" db:"status"`
-	IsAssessment bool             `json:"isAssessment" db:"is_assessment"`
+	Content            string           `json:"content" db:"content"`
+	Status             DiscussionStatus `json:"status" db:"status"`
+	IsAssessment       bool             `json:"isAssessment" db:"is_assessment"`
+	CreatedByUserInfo  UserInfo         `json:"createdByUserInfo" db:"created_by_user_info"`
+	ModifiedByUserInfo *UserInfo        `json:"modifiedByUserInfo" db:"modified_by_user_info"`
```
ClayBenson94, line 15: I think it might end up being less overall code to not have these 2 fields added to the model, but just let GQL resolvers fill them out

ClayBenson94 (changes requested):
I think these changes work for the areas they touched, but a few things are missing that might be worth considering our approach:

- It doesn't look like the resolvers that fetch discussions & replies are returning this data.
- Since the model defines these properties and the resolvers (i.e. `CreatePlanDiscussion()`) are implementing the fetch, it means that we're fetching this information every time we call one of these resolvers, even if it's not requested in the GQL

For those reasons, I think it might be worth approaching this from the angle where we don't include these properties in the model in `pkg/models`, and instead implement resolvers that make this call to CEDAR LDAP only when it's requested.

This has the added benefit of only having to implement the 2 new resolvers, rather than having to add the fetch logic to each existing resolver.

StevenWadeOddball (changes requested):
So our conversation doesn't' get lost, I agree with the points Clay brought up. I think it would be simplest and most consistent to use the graphql resolvers for user info at this point.

StevenWadeOddball:
PR is being closed in favor of https://github.com/CMSgov/mint-app/pull/397

PR #397, opened 2023-01-12
Easi 2521/discussion and reply user info gql

StevenWadeOddball:
# EASI-2521

### Backend
- Add User Info to Discussion and discussion Reply 
   - Update Postman Collection to show the new options.

### Frontend
- Updated `GetModelPlanDiscussions` queries and its types to have `createdByUser` and `commonName`
- Updated `Discussion` and `SingleDiscussion` components to use new `createdByUser`
- Removed `collaborator` prop and query since we are no longer using that to query the name of users
- Updated corresponding tests 

### Frontend
1. Create a new discussion in a model plan
2. Ensure that eua id is no longer showing (i.e. MINT), instead it is the user name (i.e. mint Doe)
3. Continue to answer the new question and ensure the User Info is showing the correct name, not eua ID

## PR Author Review Checklist

## PR Reviewer Guidelines


src/queries/Discussions/GetModelPlanDiscussions.ts
```diff
@@ -17,6 +17,9 @@ export default gql`
         createdDts
         status
         isAssessment
+        createdByUser {
+          commonName
+        }
```
ClayBenson94, line 22: @garyjzhao With the introduction of this property, do we still need to query for `collaborators` in this query, or can it be removed?

ClayBenson94 (changes requested):

src/queries/Discussions/GetModelPlanDiscussions.ts
garyjzhao: @ClayBenson94 we do not need to query `collaborator` anymore, since we changed the way we are pulling the names due to this property. 

I will remove it from the query in an upcoming commit.

ClayBenson94 (approved):

PR #408, opened 2023-01-20
[EASI 2429] Operational Needs, Operational Solution Access control

rileyanderson:
# EASI-2429

## Changes and Description

- Add OperationalNeedRelation
- Enable/Update SolutionRelation

## PR Author Review Checklist

## PR Reviewer Guidelines


StevenWadeOddball (commented):
something to note, document solution link should probably use the solution link relation

rileyanderson:
> something to note, document solution link should probably use the solution link relation

@StevenWadeOddball [It does](https://github.com/CMSgov/mint-app/blob/main/pkg/models/plan_document_solution_link.go#L8). It was already there 😄

pkg/models/operational_solution.go
```diff
@@ -9,6 +9,7 @@ import (
 // OperationalSolution represents the specific implemented solution to an Operational Need
 type OperationalSolution struct {
 	baseStruct
+	operationalNeedRelation
 	OperationalNeedID uuid.UUID               `json:"operationalNeedID" db:"operational_need_id"`
```
StevenWadeOddball, line 13: This should be removed, as the OperationalNeedID field comes from the embedded struct already, so this is redundant.

StevenWadeOddball (changes requested):
Overall looks pretty good! There are a few places where pre-update still need to be enabled so access control works ( EG Operational_Solution.go (resolvers), the call to BaseStruct Preupdate should provide true instead of false for check access. 

I will make these updates.

pkg/models/operational_solution.go
StevenWadeOddball: When you run the code, one of the variables stays null unless you set it 
image

StevenWadeOddball: This is corrected removing the extraneous field
image

StevenWadeOddball: I resolved this

StevenWadeOddball:
@OddTomBrooks and @ClayBenson94 , can you please review this when you have a chance? I made enough changes that I think it should get approval from you two instead of me. It's ready for review.

ClayBenson94 (approved):

StevenWadeOddball (approved):

PR #410, opened 2023-01-23
[NOREF] Migration Generation Utility: scripts/dev genmigration

OddTomBrooks:
Added genmigration to scripts/dev which generates a migration prefixed with the current UTC timestamp in YYYYmmddHHMMSS format.

# NOREF

## Changes and Description

- Added scripts/dev genmigration

## How to test this change

- Call scripts/dev genmigration
- Observe that a migration is generated with the format "v%Y%m%d%H%M%S__.sql"

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/scripts/dev
+++ b/scripts/dev
@@ -474,7 +474,7 @@ end
 
 desc "Generate a new migration with the current utc timestamp prefix"
 task genmigration: :consume_args do |t, args|
-  migrationName = Time.now.utc.strftime("v%Y%m%d%H%M%S__.sql")
+  migrationName = Time.now.utc.strftime("V%Y%m%d%H%M%S__.sql")
   FileUtils.touch("migrations/#{migrationName}")
 end
 
--- a/scripts/dev
+++ b/scripts/dev
@@ -5,6 +5,7 @@ require "json"
 require "yaml"
 require "irb"        # debugging
 require "io/console" # for IO#getch
+require "fileutils"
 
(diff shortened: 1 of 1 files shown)


scripts/dev
```diff
@@ -471,6 +472,12 @@ task :lint do
   sh "pre-commit run -a"
 end
 
+desc "Generate a new migration with the current utc timestamp prefix"
+task genmigration: :consume_args do |t, args|
+  migrationName = Time.now.utc.strftime("v%Y%m%d%H%M%S__.sql")
```
ClayBenson94, line 477: Not sure if flyway actually cares or not, but for consistency can you make this a capital V?

```suggestion
  migrationName = Time.now.utc.strftime("V%Y%m%d%H%M%S__.sql")
```

ClayBenson94 (changes requested):

scripts/dev
OddTomBrooks: Done

StevenWadeOddball:
Thanks for making a proof of concept for this Tom. From conversation outside of this channel, I think we are leaning towards closing this PR in favor of addressing the underlying issue of migration conflicts. I think it's ok to close this out.

StevenWadeOddball (changes requested):
Hey Tom,

Thanks again for trying this out. The code here works fine and looks good, however, I don't think we should go with this approach. I think we should lean on GitHub configuration like Clay suggested (Which would mean we don't use this utility). Mind closing this out?

PR #422, opened 2023-01-31
[EASI-2426] Operational Solution Subtasks

OddTomBrooks:
# EASI-2426

## Changes and Description

- Implementation for Operational Solution Subtasks
  - GQL
  - Migration SQL
  - Store SQL
  - Store Methods
  - Resolver Methods
  - Solution Relation Access Control Implementation

## How to test this change

- Run operational solution subtask unit tests and observe that they all pass
- Use the operational solution subtask endpoints in Postman to:
  - Create an operational solution subtask(s)
  - Update the operational solution subtask(s)
  - Get the associated operational solution and confirm that it contains the created subtask(s)
  - Delete the created operational solution subtask(s)
  - Confirm that all endpoints behave as expected

## PR Author Review Checklist

## PR Reviewer Guidelines


pkg/graph/resolvers/operational_solution_subtask.go
```diff
@@ -0,0 +1,149 @@
...
+// OperationalSolutionSubtasksGetBySolutionID implements the resolver logic to
+// get operational solution subtasks by solution ID
+func OperationalSolutionSubtasksGetBySolutionID(
+	logger *zap.Logger,
+	store *storage.Store,
+	principal authentication.Principal,
+	solutionID uuid.UUID,
+) ([]*models.OperationalSolutionSubtask, error) {
+
+	isCollaborator, err := accesscontrol.IsCollaboratorSolutionID(logger, principal, store, solutionID)
```
StevenWadeOddball, line 71: We don't normally block access from viewing a model if the user isn't a collaborator (anyone can view). It's just updating and deleting that normally need that.  I think the access control check should be removed from this function.

pkg/graph/schema.graphql
```diff
@@ -1487,6 +1515,7 @@ type Query {
   possibleOperationalNeeds: [PossibleOperationalNeed!]!
   possibleOperationalSolutions: [PossibleOperationalSolution!]!
   userAccount(username: String!): UserAccount!
+  operationalSolutionSubtask(id: UUID!): OperationalSolutionSubtask
```
StevenWadeOddball, line 1518: I'm not sure this needs to be exposed. Was there something in FIGMA that suggests the front end would ever need to get a specific SubTask by ID? I could see the front end getting the collection by solution ID however. I think we mostly would need get by id for pre-update statements.

StevenWadeOddball (changes requested):
Overall this looks good Tom! Thanks for making this. I aded some areas that I think could be updated. 

Could you please also add a little bit of seeded data for this? I imagine the model that already has seeded operational Solutions could just have a few subtasks added on.

migrations/V50__Add_Operational_Solution_Subtask.sql
```diff
@@ -0,0 +1,25 @@
...
+    modified_by UUID REFERENCES public.user_account (id) MATCH SIMPLE,
+    modified_dts TIMESTAMP WITH TIME ZONE
+);
+
+
+ALTER TABLE operational_solution_subtasks
+ADD CONSTRAINT fk_operational_solution_subtasks FOREIGN KEY (solution_id)
+REFERENCES public.operational_solution (id) MATCH SIMPLE
+ON UPDATE NO ACTION
+ON DELETE NO ACTION
```
StevenWadeOddball, line 25: One other thing I realized is that auditing should be turned on for the table. Migration V23 is a good example how to turn that on. One caveat is that migration is using a function that is not updated now to include if the table uses the user table (just need to set that in the audit.table_config for this entry).

pkg/storage/SQL/operational_solution_subtask/update.sql
```diff
@@ -0,0 +1,17 @@
+UPDATE operational_solution_subtasks
+SET
+    solution_id = :solution_id,
```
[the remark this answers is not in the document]
OddTomBrooks: Great point. This was an oversight. Binding fields should be immutable.

pkg/storage/operational_solution_subtaskStore.go
```diff
@@ -0,0 +1,153 @@
...
+	err = tx.Commit()
+	if err != nil {
+		return nil, fmt.Errorf("could not commit subtask creation transaction: %w", err)
+	}
+
+	return results, nil
+}
+
+// OperationalSolutionSubtaskGetByID gets a models.OperationalSolutionSubtask by ID
+func (s *Store) OperationalSolutionSubtaskGetByID(_ *zap.Logger, subtaskID uuid.UUID) (*models.OperationalSolutionSubtask, error) {
```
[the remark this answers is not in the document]
OddTomBrooks: I agree with this, would definitely like to make the investment. I think it's outside the scope of this ticket but would be happy to do this in another ticket.

pkg/graph/schema.graphql
```diff
@@ -1465,6 +1466,33 @@ modifiedBy: String
...
+  status: OperationalSolutionSubtaskStatus!
+}
+
+type OperationalSolutionSubtask {
+  id: UUID!
+  solutionID: UUID!
+  name: String!
+  status: OperationalSolutionSubtaskStatus!
+
+  createdBy: UUID!
```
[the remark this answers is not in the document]
OddTomBrooks: Ah nice 👍

pkg/storage/operational_solution_subtaskStore.go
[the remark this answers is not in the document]
StevenWadeOddball: I 100% agree! I think we should ticket this for the future. I don't think it's probably the next priority either, but definitely before the feature goes live. Thanks Tom!

pkg/graph/schema.graphql
OddTomBrooks: We had briefly discussed it in our sync around operational solution subtasks. My understanding was that we intended to provide it as a matter of good practice / API completion. If you prefer to remove the endpoint it's certainly quick and easy to do, but it seems to me that we only lose platform capability.

StevenWadeOddball: hmm 🤔. I think the main thing I think of for this is if we aren't going to need it, we probably shouldn't have it. I know they AC did bring up collection by solution ID, I just don't know that this is needed. 

@patrickseguraoddball, do you mind weighing if about querying a singular solution subtask on the front end? Also, do anticipate a need to just get the subtasks by solution id? Or is nested in the solution sufficient?

patrickseguraoddball: I just looked through the Figma, and I don;t see any instance we we would query a singular subtask or query subtasks without needed the full need/solution.  I can see a future scenario where this may change, but I don't believe currently needed

StevenWadeOddball: Thanks so much @patrickseguraoddball! I think in that case @OddTomBrooks , I'd remove it so we don't need to support something we don't anticipate needing.

OddTomBrooks: Makes sense. The underlying capability is still there for internal purposes, so it's trivial if we ever need to restore it.

pkg/storage/truncate.go
```diff
@@ -22,6 +22,7 @@ func (s *Store) TruncateAllTablesDANGEROUS(logger *zap.Logger) error {
     plan_it_tools,
     plan_favorite,
     plan_cr_tdl,
+	operational_solution_subtask,
```
StevenWadeOddball, line 25: Sorry I think I forgot to submit this comment last time. There is an error message about this when you run `scripts/dev test:go` unit test, caused because this needs to be replicated there. Could you please add it there as well? 

```
ERROR:  cannot truncate a table referenced in a foreign key constraint
DETAIL:  Table "operational_solution_subtask" references "operational_solution".
HINT:  Truncate table "operational_solution_subtask" at the same time, or use TRUNCATE ... CASCADE.
```

StevenWadeOddball (changes requested):
Overall looks good! I think this is basically there! It just needs to grab the changes from main, and add to the truncate statement in the dev script!

StevenWadeOddball (changes requested):
Hey @OddTomBrooks,

Thanks for pulling in main. The generic account resolver seems to be working correctly! The only things remaining are
1. scripts/dev db:clean truncate statement
2. seed data

StevenWadeOddball (approved):
Looks great! Excellent job, thanks @OddTomBrooks

PR #444, opened 2023-02-09
Easi 2610/model plan to user table

StevenWadeOddball:
# EASI-2610

## Changes and Description

- Model Plan now uses the user table
- DB seed refactored to more easily allow for migration
- Plan Favorite, exposed the user account GQL

## How to test this change
1. run scripts/dev db:seed
   a. ensure data is is seeded as expected
2. Update model plans using the front end and make sure that it is working as expected.

## PR Author Review Checklist

## PR Reviewer Guidelines


pkg/graph/resolvers/model_plan.go
```diff
@@ -20,7 +20,7 @@ import (
 // TODO Revist this function, as we probably want to add all of these DB entries inthe scope of a single SQL transaction
```
OddTomBrooks, line 20:
```suggestion
// TODO: Revisit this function, as we probably want to add all of these DB entries in the scope of a single SQL transaction
```

pkg/models/model_helpers.go
```diff
@@ -1,6 +1,9 @@
 package models
 
-import "github.com/lib/pq"
+import (
+	"github.com/google/uuid"
+	"github.com/lib/pq"
+)
 
 // StringPointer returns a pointer to a string input
 func StringPointer(st string) *string {
```
OddTomBrooks, line 9: Why not simply use &st?

OddTomBrooks (changes requested):

pkg/graph/resolvers/model_plan.go
StevenWadeOddball: Hi @OddTomBrooks , that was a note from a previous time, not from this ticket, I don't this is the ticket to address it

pkg/models/model_helpers.go
StevenWadeOddball: Hi @OddTomBrooks , this is also not a change from this ticket, and if we changed it it should be in a different ticket.

pkg/graph/resolvers/model_plan.go
OddTomBrooks: Fair enough. Bad formatting on TODO: may skip IDEs from picking them up so it threw a large red flag for me.

pkg/models/model_helpers.go
OddTomBrooks: Fair enough. 👍

OddTomBrooks (approved):

PR #448, opened 2023-02-13
[NOREF] Table User Account Migration Tool Base Struct SQL Bugfix

OddTomBrooks:
# EASI-NOREF

## Changes and Description

- Updated base_struct sql to use input variable for table name generation rather than model_plan

## PR Author Review Checklist

## PR Reviewer Guidelines


StevenWadeOddball (changes requested):
Should be good besides the extra migration being in there. Also wanted to check, is it possible to make the branch off of origin main instead of local main?

There is an issue when you need to migrate the python script,  it requires you to commit to main before you run it. Ideally, you could modify the file without needing to commit to main.

StevenWadeOddball (approved):
The changes look great, thanks Tom! Thanks for making it just write the SQL too, that will be pretty helpful for my workflow.

BTW, I think [this pr](https://github.com/CMSgov/mint-app/pull/459) will correct the server tests.  Once that's merged into main, this should be good to go

PR #451, opened 2023-02-13
Easi 2612/user account migration plan ops eval and learning

StevenWadeOddball:
# EASI-2612

## Changes and Description

- Ops Eval and Learning updated to use the user table
- Auditing config updated
- Operational Needs trigger disabled for this task list section

## How to test this change
1. Bring the app up, and seed data.
2.  Verify that you can update Ops Eval and Learning Section
3. Verify Unit test pass as expected. 

## PR Author Review Checklist

## PR Reviewer Guidelines


pkg/graph/resolvers/operational_need_triggers_test.go
```diff
@@ -1,223 +1,223 @@
...
-
-}
+// import (
+// 	"reflect"
+
+// 	"github.com/cmsgov/mint-app/pkg/graph/model"
+// 	"github.com/cmsgov/mint-app/pkg/models"
+// )
+
+// func (suite *ResolverSuite) TestGeneralCharacteristicsNeeds() {
```
ClayBenson94, line 10: I think what might work well here is calling something like this at the start of each test:

```go
suite.T().Skip("TODO: Skipping while operational needs trigger is disabled")
```

This way you don't get out-of-date testing code if something changes related to a type/method used in our tests, and you get visibility in our testing suite as to the fact that tests exist that are being skipped (as opposed to comments, which the testing suite has no clue they even exist)

ClayBenson94: From local testing, this might also require a lint ignore on `findOpNeed` (the linter is smart enough to know that `t.Skip()` makes the following code unreachable!

```go
//lint:ignore U1000 We are currently calling `t.Skip()` on tests, and this method is unused because of it.
func findOpNeed(collection []*models.OperationalNeed, key models.OperationalNeedKey) *models.OperationalNeed {
...
```

migrations/V58__Update_Plan_Ops_Eval_and_Learning_To_User_Account_Table.sql
```diff
@@ -0,0 +1,80 @@
...
+
+/* turn on audit trigger */
+
+ALTER TABLE plan_ops_eval_and_learning
+ENABLE TRIGGER audit_trigger;
+
+/* disable need trigger as it is currently non-functional until needs are migrated */
+
+ALTER TABLE plan_ops_eval_and_learning
+DISABLE TRIGGER operational_need_trigger;
```
ClayBenson94, line 80: I think (though functionally equivalent) it makes sense to make this part of the migration V59.

The rest of this migration (as the name implies) is migrating the Ops Eval Learning section to the user table, but this part is responsible for something else entirely (disabling functionality while we write future migrations)

To ensure that this block is easy to find (and write the corresponding `enable trigger` migration later, can you break this out into a new migration?

ClayBenson94 (changes requested):

migrations/V58__Update_Plan_Ops_Eval_and_Learning_To_User_Account_Table.sql
StevenWadeOddball: Thanks Clay! Good call. I was disabling the trigger for each table in each user table migration, but I think I'm going to include disabling the trigger on each table in this new migration so it's all consolidated.

pkg/graph/resolvers/operational_need_triggers_test.go
StevenWadeOddball: Thanks Clay! That is a good call, I'll implement it.

StevenWadeOddball:
Closing in favor of https://github.com/CMSgov/mint-app/pull/458 which includes these changes.

PR #461, opened 2023-02-15
[NOREF] Support for removing multiple solutions

ClayBenson94:
# NOREF

## Changes and Description

- Modify `removePlanDocumentSolutionLinks` to take the same parameters as `createPlanDocumentSolutionLinks` (a single solution ID, multiple document IDs)

Follows up on https://github.com/CMSgov/mint-app/pull/296

## How to test this change

- Ensure go tests pass with `scripts/dev up:backend` and `scripts/dev test:go`

- Ensure functionality works as expected
  - Create a model plan
  - Create a custom operational solution
  - Link / unlink multiple documents at once

## PR Author Review Checklist

## PR Reviewer Guidelines


StevenWadeOddball (approved):
Thanks Clay,

This looks good. I pulled locally and used the endpoint, and it removed the links as anticipated.

One interesting behavior to note with the join condition, it only needs one combination of document_id and solution_id to work. Example, I sent the request with two `doc_ids`, one was linked to the solution, one was not. 

It successfully unlinked the document that was linked to the solution, and it doesn't do anything to the other combination (because there is no record). As long as one combination results in a link being removed, you will get a success message. I think this is likely as it should be expected.

ClayBenson94:
> Thanks Clay,
> 
> This looks good. I pulled locally and used the endpoint, and it removed the links as anticipated.
> 
> One interesting behavior to note with the join condition, it only needs one combination of document_id and solution_id to work. Example, I sent the request with two `doc_ids`, one was linked to the solution, one was not.
> 
> It successfully unlinked the document that was linked to the solution, and it doesn't do anything to the other combination (because there is no record). As long as one combination results in a link being removed, you will get a success message. I think this is likely as it should be expected.

Hmm - I think that makes sense, given that `DELETE ... WHERE ...` will succeed even if the `WHERE` returns nothing - though, I wonder if it'll error if one of the deletes actually fails 🤔

StevenWadeOddball:
> > Thanks Clay,
> > This looks good. I pulled locally and used the endpoint, and it removed the links as anticipated.
> > One interesting behavior to note with the join condition, it only needs one combination of document_id and solution_id to work. Example, I sent the request with two `doc_ids`, one was linked to the solution, one was not.
> > It successfully unlinked the document that was linked to the solution, and it doesn't do anything to the other combination (because there is no record). As long as one combination results in a link being removed, you will get a success message. I think this is likely as it should be expected.
> 
> Hmm - I think that makes sense, given that `DELETE ... WHERE ...` will succeed even if the `WHERE` returns nothing - though, I wonder if it'll error if one of the deletes actually fails 🤔

Good question, I don't know how that would happen 🤔 . If there isn't anything to check, you do get a no row in result set error (which I guess is different). I can't think of a situation where a valid matched entry would be deleted, but another valid match wouldn't be (unless we had some dependent foreign key relationships in the DB, which I'm pretty sure we don't have on this table)

PR #464, opened 2023-02-15
Easi 2622/existing model to user table

StevenWadeOddball:
# EASI-2622

## Changes and Description

- Migrate existing model to the user table
     - NOTE this didn't follow the exact migration as the other tables. Notably, we know the state of the created by field, so it is simplified to just set it as the system account
     - It should also be noted that this field doesn't implement base struct, so the user account fields had to be implemented in the model itself

## How to test this change
1. Bring up the app
2. Navigate to Plan General Characteristics
3. Select this plan resembles an existing model
4. Select some existing models
5. Verify that you can save the application

## PR Author Review Checklist

## PR Reviewer Guidelines


pkg/models/existing_model.go
```diff
@@ -23,8 +29,30 @@ type ExistingModel struct {
...
+}
+
+// ModifiedByUserAccount returns the user account of the user who created the struct from the DB using the UserAccount service
+func (eM *ExistingModel) ModifiedByUserAccount(ctx context.Context) *authentication.UserAccount {
+
+	if eM.ModifiedBy == nil {
+		return nil
+	}
+	service := appcontext.UserAccountService(ctx)
+	// service := authentication.UserAccountService(ctx)
```
OddTomBrooks, line 54: Does this need to be here?

StevenWadeOddball: Thanks Tom, good call, I removed it.

OddTomBrooks (approved):

PR #474, opened 2023-02-21
[EASI-2666] Updated existing_models Data to Current Spec

OddTomBrooks:
# EASI-2666

## Changes and Description

- Added migration to update existing_models table with current data

## How to test this change

Run scripts/dev db:seed and observe that the existing_models table is populated with current data

## PR Author Review Checklist

## PR Reviewer Guidelines


OddTomBrooks (commented):
The migration version will need to be updated. This feature is lower in the priority stack than the user account table migrations. Once those are in I will update the migration name to the appropriate version.

StevenWadeOddball (changes requested):
@OddTomBrooks, it looks like this migration has a syntax issue (like something is missing an id  somewhere). It might also be because the these ids are ints and they are quoted like a string in the mutation. Could you take a look please?

Also, do you know what caused the yarn.lock to be committed? It probably would be good to remove that.

OddTomBrooks:
> @OddTomBrooks, it looks like this migration has a syntax issue (like something is missing an id somewhere). It might also be because the these ids are ints and they are quoted like a string in the mutation. Could you take a look please?
> 
> Also, do you know what caused the yarn.lock to be committed? It probably would be good to remove that.

It's a good lesson that just because the query works in the console does not indicate that it will work in the migration.

The input data violates many of our constraints and is requiring quite a bit of sanitization. I will continue correcting these issues.

OddTomBrooks:
> > @OddTomBrooks, it looks like this migration has a syntax issue (like something is missing an id somewhere). It might also be because the these ids are ints and they are quoted like a string in the mutation. Could you take a look please?
> > Also, do you know what caused the yarn.lock to be committed? It probably would be good to remove that.
> 
> It's a good lesson that just because the query works in the console does not indicate that it will work in the migration.
> 
> The input data violates many of our constraints and is requiring quite a bit of sanitization. I will continue correcting these issues.

Well it certainly helps if you actually copy out of your console back into the migration. Luckily the solution is very straightforward. I will do some cleanup on the tool script and add that as well for any future work.

OddTomBrooks:
There were a few updates to migrations:

Naming convention
EUAID -> UUID
Removed unnecessary transaction block


I have also added the tool script I made to generate this migration. The tool is optional, I hope that it assists future work with this table until we have a more permanent installation. There were a few concessions which had to be made to sanitize the input data into something appropriate with our database constraints, please let me know if you prefer any other solutions.

StevenWadeOddball:
Thanks @OddTomBrooks, I brought the app up, and verified the modules against the modules provided in the spreadsheet. I would maybe not make the ids start so high (they actually started as they did because they had an id from the original source), but I  think it matters for this since we are just getting names.

I'm not sure we should include the python scripts in the repo. I think it's great you have it as a personal tool if you want to use it for generating a migration, but I don't think that this is something that we should maintain in the code base. Thoughts? What do you think @ClayBenson94?

OddTomBrooks:
> Thanks @OddTomBrooks, I brought the app up, and verified the modules against the modules provided in the spreadsheet. I would maybe not make the ids start so high (they actually started as they did because they had an id from the original source), but I think it matters for this since we are just getting names.
> 
> I'm not sure we should include the python scripts in the repo. I think it's great you have it as a personal tool if you want to use it for generating a migration, but I don't think that this is something that we should maintain in the code base. Thoughts? What do you think @ClayBenson94?

Ah, the large IDs. This was an interesting problem. 
Since we have the case of duplicate IDs which must be parsed as integers, I wanted to preserve the IDs as they were in a fashion where you can readily interpret the number of collisions without simply incrementing and running the risk of forcing another valid UID to change.

The idea here was to keep the base ID and add to it a multiple of an offset value. In this case, I put 100000 as there is little to no chance we will ever have a model ID at that scale.

Let's say we have 3 rows with ID 55.

The first row will simply maintain ID 55 as there is not yet a collision.
The second row will have one collision, so its ID will be 100055. At a glance we can quickly infer that the base ID is 55 and there is one collision.
The third row will have two collisions, so its ID will be 200055. At a glance we can tell there are two collisions and the base ID is still 55.

Please let me know if this makes sense to you and if there is another approach you would prefer.

StevenWadeOddball:
> > Thanks @OddTomBrooks, I brought the app up, and verified the modules against the modules provided in the spreadsheet. I would maybe not make the ids start so high (they actually started as they did because they had an id from the original source), but I think it matters for this since we are just getting names.
> > I'm not sure we should include the python scripts in the repo. I think it's great you have it as a personal tool if you want to use it for generating a migration, but I don't think that this is something that we should maintain in the code base. Thoughts? What do you think @ClayBenson94?
> 
> Ah, the large IDs. This was an interesting problem. Since we have the case of duplicate IDs which must be parsed as integers, I wanted to preserve the IDs as they were in a fashion where you can readily interpret the number of collisions without simply incrementing and running the risk of forcing another valid UID to change.
> 
> The idea here was to keep the base ID and add to it a multiple of an offset value. In this case, I put 100000 as there is little to no chance we will ever have a model ID at that scale.
> 
> Let's say we have 3 rows with ID 55.
> 
> The first row will simply maintain ID 55 as there is not yet a collision. The second row will have one collision, so its ID will be 100055. At a glance we can quickly infer that the base ID is 55 and there is one collision. The third row will have two collisions, so its ID will be 200055. At a glance we can tell there are two collisions and the base ID is still 55.
> 
> Please let me know if this makes sense to you and if there is another approach you would prefer.

That works! If this table stays around much longer, I'd consider changing to UUID instead of int, but I think for the sake of this ticket, I'm good with the migration as it stands. I'm just not sure about including the python scripts as part of this.

OddTomBrooks:
> That works! If this table stays around much longer, I'd consider changing to UUID instead of int, but I think for the sake of this ticket, I'm good with the migration as it stands. I'm just not sure about including the python scripts as part of this.

I also have gone back and forth on this point. I figured I'd make it available in case you wanted it. Another option I've seen is maintaining a separate repo for tools. I'm fine with whatever you guys want to do.

StevenWadeOddball:
> I also have gone back and forth on this point. I figured I'd make it available in case you wanted it. Another option I've seen is maintaining a separate repo for tools. I'm fine with whatever you guys want to do.

Thanks @OddTomBrooks , yeah I think I personally would remove it. Some way to share tools could be good in the future. I also think temporarily committing a script that will be used repeatedly and removing when it's done makes sense (like we did the user migration one).

ClayBenson94:
I think some context on this table itself is also helpful: I don't think we really plan to keep this `existing models` table around long term, and I really don't think we plan on making many more updates to it that fall outside "maybe add another one to the list"

Frankly, we could just hard-code this list of existing models as a `[]models.ExistingModel` and just nix the entire concept of making a DB query for this. We only ever grab the name of the model for use in the rest of the app (we don't FK by `id` or anything)
image


Either way - I think the simplest approach here is likely the best, so if we have an updated migration that's good, I think just committing that and moving forward makes sense - if we end up ever needing to make new updates to this list, I think just making it a hard-coded list rather than a DB table might just make our lives easier.

That work for y'all @OddTomBrooks & @StevenWadeOddball ?

ClayBenson94 (changes requested):
Requesting changes for this comment: https://github.com/CMSgov/mint-app/pull/474#issuecomment-1441885428

StevenWadeOddball:
> I think some context on this table itself is also helpful: I don't think we really plan to keep this `existing models` table around long term, and I really don't think we plan on making many more updates to it that fall outside "maybe add another one to the list"
> 
> Frankly, we could just hard-code this list of existing models as a `[]models.ExistingModel` and just nix the entire concept of making a DB query for this. We only ever grab the name of the model for use in the rest of the app (we don't FK by `id` or anything) image
> 
> Either way - I think the simplest approach here is likely the best, so if we have an updated migration that's good, I think just committing that and moving forward makes sense - if we end up ever needing to make new updates to this list, I think just making it a hard-coded list rather than a DB table might just make our lives easier.
> 
> That work for y'all @OddTomBrooks & @StevenWadeOddball ?

@ClayBenson94 , yeah that sounds good to me! I'm good with going with the migration and removing the python scripts for now. But I think you are right, there really is no reason for it to be in the DB with how we are using it. It could just be an embedded JSON file, or a hard coded list.

StevenWadeOddball:
@OddTomBrooks , I think this is good to go once the migration gets updated to V78

StevenWadeOddball (approved):
🚀

PR #481, opened 2023-02-24
Easi 2465/operational needs names trigger update

StevenWadeOddball:
# EASI-2465

## Changes and Description

- New Function to return a result set of Operational Needs to show if they are needed or not
  - previously, a cleared result would not set needed to NULL, this now will set it to NULL, if the new value is null
- Migration to set needed for all existing models

## How to test this change
1. Check that the need name changes have been updated as specified in the JIRA ticket
2. Make sure that the new configurations for the trigger logic in the possible operational need table matches the specs in the JIRA ticket.
3. Ensure that Operational Needs get set to needed / not needed as a result of this trigger

If you want to test the migration that sets existing needs as needed, you can do the following
1. Comment out the enable trigger migration (so that the trigger is not run when you make changes)
2. Make some model plans with answers that would normally trigger the changes
3. Run the SQL contained in v84 to set the needs for all existing models.
4.  Confirm that the existing needs were all set as expected.

## Calling the SQL Functions
1. If you want to explore the SQL functions individually, you can call them like this
```SQL
        SELECT 	* FROM	DETERMINE_SECTION_NEEDS('plan_ops_eval_and_learning',
                model_plan_idParam,
                (SELECT hstore(PPP) FROM plan_ops_eval_and_learning AS ppp WHERE ppp.model_plan_id = 'UUID HERE')
```
5. To see it for all models in the DB
```SQL
    SELECT OpNeeds.*  FROM model_plan mp, DETERMINE_MODEL_PLAN_NEEDS(mp.id) OpNeeds
```
6. To see for an individual model
```SQL

SELECT * FROM DETERMINE_MODEL_PLAN_NEEDS('915b9e0d-8ea5-439f-bc4a-f77291894e0f')

```

## PR Author Review Checklist

## PR Reviewer Guidelines


StevenWadeOddball:
Note, there is an issue with the logic for the composite needs.  It is due to using the h_changed hstore for getting the vals. Will refactor to provide the changed columns, and also pass the new values of the row.

migrations/V80__Add_Get_Needed_Function.SQL
```diff
@@ -0,0 +1,98 @@
+CREATE OR REPLACE FUNCTION GET_NEED_NEEDED( table_name TEXT, model_plan_idParam UUID, h_changed HSTORE ) --TODO: we could make this take an array if want
```
OddTomBrooks, line 1: The name is a bit confusing. Some alternatives might be DETERMINE_REQUIRED_NEED or GET_REQUIRED_NEED. What do you think? Do you have any other ideas?

StevenWadeOddball: That makes sense to me! Thanks for the suggestion @OddTomBrooks , I'll update it

StevenWadeOddball: I settled on `DETERMINE_SECTION_NEEDS`

migrations/V79__Update_Operational_Needs.sql
```diff
@@ -0,0 +1,79 @@
+/* Rename old key name */
+ALTER TYPE OPERATIONAL_NEED_KEY RENAME TO OPERATIONAL_NEED_KEY_OLD;
```
ClayBenson94, line 2: Can you remove the old enum at the end of this migration?

ClayBenson94 (changes requested):

migrations/V79__Update_Operational_Needs.sql
StevenWadeOddball: Thanks for catching this Clay! Yes, I'll drop this and update the keys elsewhere.

StevenWadeOddball: Updated! I ended up skipping the rename and dropped the type earlier instead.

OddTomBrooks (approved):

PR #483, opened 2023-02-28
[EASI-2556] Model Plan Created Email to MINT Dev Team

OddTomBrooks:
# EASI-2556

## Changes and Description

- Added MINT Dev Team Email to ENV VARS
- Added Dev Team Email to Email Service Config
- Added Model Plan Created Email to MINT Dev Team

## How to test this change

- Start the MINT dev environment
- Open MailCatcher at http://localhost:1085/
- Create a Model Plan
- Confirm that you received an email on MailCatcher to the MINT dev team

## PR Author Review Checklist

## PR Reviewer Guidelines


pkg/shared/oddmail/go_simple_mail_service_config.go
```diff
@@ -23,6 +23,7 @@ type GoSimpleMailServiceConfig struct {
 	SendTimeout    time.Duration
 	KeepAlive      bool
 	TLSConfig      *tls.Config
+	DevTeamEmail   string
```
ClayBenson94, line 26: Would it make more sense to call this `MINTTeamEmail`?

pkg/graph/resolvers/model_plan.go
```diff
@@ -108,9 +120,62 @@ func ModelPlanCreate(ctx context.Context, logger *zap.Logger, modelName string,
...
+		err = sendModelPlanCreatedEmail(
+			ctx,
+			emailService,
+			emailTemplateService,
+			emailService.GetConfig().GetDevTeamEmail(),
+			createdPlan,
+		)
+		if err != nil {
+			return nil, err
+		}
```
ClayBenson94, line 133: Do we want to throw an error here and block? Wondering if, since this isn't in Faktory yet, it makes more sense to just async this call and log the error if we get it, but otherwise just pass through this function and let the GQL mutation respond before the email is sent

ClayBenson94 (changes requested):

pkg/shared/oddmail/go_simple_mail_service_config.go
OddTomBrooks: Originally I had named it MINTTeamEmail but as this is in the shared package I thought it more appropriate to name it in an agnostic fashion so the package may be more readily shared outside our team at a later date.

pkg/graph/resolvers/model_plan.go
OddTomBrooks: Good point. In the course of development I ran into this error and it was disorienting, I will make this change.

pkg/graph/resolvers/model_plan.go
```diff
@@ -108,9 +120,63 @@ func ModelPlanCreate(ctx context.Context, logger *zap.Logger, modelName string,
...
+			emailService,
+			emailTemplateService,
+			emailService.GetConfig().GetDevTeamEmail(),
+			createdPlan,
+		)
+		if err != nil {
+			logger.Error("failed to send model plan created email to dev team", zap.Error(err))
+			err = nil
+		}
+	}
```
ClayBenson94, line 135: I probably should have clarified what I meant in my previous review - my feedback around not blocking the response was less specifically around how we handle the error, but was more around not blocking the response of this resolver based on the email being sent.

I think what we could do is wrap this code in an anonymous goroutine so that it runs asynchronously, that way we can respond as soon as we create the entity in the DB, and the email will fire off / log errors on its own.

ClayBenson94 (changes requested):

pkg/shared/oddmail/go_simple_mail_service_config.go
StevenWadeOddball: I find DevTeamEmail confusing as well. Dev team sounds like it is going to the developers of the application, eg us, instead of to but it is going to the business owner team, not the development team. 

Is there any reason that this needs to live in the simple mail service config? This really feels like an application config, and could be set in the appconfig package instead of the shared service config. Thoughts?

pkg/graph/resolvers/model_plan.go
StevenWadeOddball: I think this would be good too, and next step would be offloading this to faktory. I happened to test this and make the error pop up, and not block model plan creation which is great. I do think that offloading it further to make it asynchronous is important here.

StevenWadeOddball (changes requested):
overall, this works as expected! I agree with Clay's feedback. Verifying that the font etc matches what is documented in FIGMA (like you suggested) could be a good thing as well.

pkg/graph/resolvers/model_plan.go
OddTomBrooks: 👍

OddTomBrooks:
Made changes to include an email address book as discussed and wire it into various systems

StevenWadeOddball (approved):
This works as expected! Thanks @OddTomBrooks 🚀

PR #489, opened 2023-03-06
[EASI-2558] New question email to MINT dev team

OddTomBrooks:
# EASI-2558

Review [EASI-2556](https://github.com/CMSgov/mint-app/pull/483) before this PR as it shares commits

## Changes and Description

- Added email subject and body templates
- Updated code to send MINT dev team an email when a new question is asked

## PR Author Review Checklist

## PR Reviewer Guidelines


pkg/graph/resolvers/plan_discussion.go
```diff
@@ -12,7 +17,16 @@ import (
 )
 
 // CreatePlanDiscussion implements resolver logic to create a plan Discussion object
-func CreatePlanDiscussion(logger *zap.Logger, input *model.PlanDiscussionCreateInput, principal authentication.Principal, store *storage.Store) (*models.PlanDiscussion, error) {
+func CreatePlanDiscussion(
```
StevenWadeOddball, line 20: @OddTomBrooks , @ClayBenson94 , I wanted to hear what you both think about removing some of these extra parameters, and instead wrapping more of these dependencies in context?
 
This new method signature requires context be passed, `logger` and `principal` already happen to live on the context. I think it could make sense to decorate the context with `email service` etc as well.

pkg/graph/resolvers/plan_discussion.go
```diff
@@ -21,9 +35,77 @@ func CreatePlanDiscussion(logger *zap.Logger, input *model.PlanDiscussionCreateI
 	}
 
 	result, err := store.PlanDiscussionCreate(logger, planDiscussion)
+	if err != nil {
+		return nil, err
+	}
+
+	// Send email to MINT Dev Team
+	err = sendPlanDiscussionCreatedEmail(
```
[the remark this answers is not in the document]
OddTomBrooks: Good catch, this should have been in. Done.

pkg/graph/resolvers/plan_discussion.go
OddTomBrooks: That's an interesting wide-scale problem to solve. It's a balancing act. I'm not overly fond of the context pattern for passing services. As far as I can tell the primary use of context in Golang is to support preserving state in libraries. While it certainly can be used to transport services that is basically one step removed from the singleton or global pattern. It is possible to swap dependencies in the context stage but it is very limited and resource lifecycle is poorly managed.

I suspect it is an easier architectural question whether/how to group parameters into categorical objects (e.g. Services, CreatePlanDiscussionInput, etc). According to Clean Code this is a proper refinement. Another consideration is that functions with long parameter lists indicate multiple responsibility. Golang is more functional than I am accustomed, but one option may be to move resolver function groups into an OO model (PlanDiscussion struct) and use a factory pattern for dependency construction to simplify things.

StevenWadeOddball: Agreed! It's outside of the scope of this PR. It does start getting into paradigm choice, it's something that's been floating around the back of my mind. Maybe these options are something we could explore during a pairing session sometime?

Those are some good points too about dependency swapping, the way we have it probably makes testing a good deal simpler as well 🤔. It might not be worth changing, but good to explore.

StevenWadeOddball (dismissed):
Thanks @OddTomBrooks, 

This looks good! There is some oddity in the email  styling I noticed. The header looks different between the model plan created email and the discussion email. I think that there is something going on with the styling. If you have the time, it might be worth seeing if you can adjust it, but it doesn't have to block this.

I think the FIGMA also shows a bit more padding on the start of the text, but like above I don't think this needs to block this PR. The Daily Digest email seems to have the spacing working, using the table for a grid system. Might work here for spacing as well.

image

OddTomBrooks:
> Thanks @OddTomBrooks,
> 
> This looks good! There is some oddity in the email styling I noticed. The header looks different between the model plan created email and the discussion email. I think that there is something going on with the styling. If you have the time, it might be worth seeing if you can adjust it, but it doesn't have to block this.
> 
> I think the FIGMA also shows a bit more padding on the start of the text, but like above I don't think this needs to block this PR. The Daily Digest email seems to have the spacing working, using the table for a grid system. Might work here for spacing as well.
> 
> image

As I've been through these emails I've been making incremental improvements better conforming to the font sizing, styling, and padding. I haven't been sure how to precisely get the padding from Figma into an HTML format so I've been eyeballing it, that can likely be improved. I will create a NOREF to copy the CSS styling from New Question email to New Model email in order to unify styling. I attempted to link a common stylesheet last week and was unable to get it working in a small timebox, something to revisit later. I have some ideas how to do that using Go templating.

I will check the padding and get it closer.

StevenWadeOddball:
> > Thanks @OddTomBrooks,
> > This looks good! There is some oddity in the email styling I noticed. The header looks different between the model plan created email and the discussion email. I think that there is something going on with the styling. If you have the time, it might be worth seeing if you can adjust it, but it doesn't have to block this.
> > I think the FIGMA also shows a bit more padding on the start of the text, but like above I don't think this needs to block this PR. The Daily Digest email seems to have the spacing working, using the table for a grid system. Might work here for spacing as well.
> > image
> 
> As I've been through these emails I've been making incremental improvements better conforming to the font sizing, styling, and padding. I haven't been sure how to precisely get the padding from Figma into an HTML format so I've been eyeballing it, that can likely be improved. I will create a NOREF to copy the CSS styling from New Question email to New Model email in order to unify styling. I attempted to link a common stylesheet last week and was unable to get it working in a small timebox, something to revisit later. I have some ideas how to do that using Go templating.
> 
> I will check the padding and get it closer.

Thanks @OddTomBrooks, that sounds good, it works for me to merge this and revisit later. We could ticket this to get a shared style sheet as well, I think that could be a good idea.

StevenWadeOddball (approved):
Thanks @OddTomBrooks, I think we can revisit later. I think the left padding is still close to the edge, but I think it is good to get this deployed and in use. 

image

PR #518, opened 2023-03-30
[EASI-2764] Search Change History Proof of Concept

OddTomBrooks:
# EASI-2764

## Changes and Description

- Added Logstash, OpenSearch, OpenSearch-Dashboards
- Added Logstash data ingestion
- Added endpoints to search on OpenSearch
  - SearchChangeTable
    - Accepts an Elasticsearch DSL query
  - SearchChangeTableWithFreeText
  - SearchChangeTableByModelPlanID
  - SearchChangeTableByDateRange
  - SearchChangeTableByActor
  - SearchChangeTableByModelStatus

### Initialize Your Environment
- Run [scripts/set_vm_map.sh](https://github.com/CMSgov/mint-app/pull/518/files#diff-171c3036c8f86feab777a6cc9b100bde502a47facf5a7f1503f9dc41de068df7)

### Test Endpoints
- Open Postman
- Open MINT / Search filter
- Run each endpoint and confirm they function as expected

## PR Author Review Checklist

## PR Reviewer Guidelines


scripts/set_vm_map.sh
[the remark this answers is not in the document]
StevenWadeOddball: hmm I just tried to run the script, but I get an error on mac `sysctl: unknown oid 'vm.max_map_count'` . I'm going to walk through the other steps and see what I discover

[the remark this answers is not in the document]
StevenWadeOddball: per our discussion and local testing, it doesn't seem like you need to do this on mac :)

MINT.postman_collection.json
StevenWadeOddball, line 1: @OddTomBrooks, it looks like this might have overridden other postman file changes. Do you think you might have not re-imported the file before making the changes?

I also noticed both the queries that reference elastic search don't work (likely because of the switch to Open Search. Do you think it makes sense to remove them?

StevenWadeOddball:
Testing locally this is looking good! I need to continue reviewing, but wanted to ask some high level questions as I'm processing it.

1. What do you anticipate is complete as is vs what do you anticipate changing in subsequent tickets?
   a. (EG queries, ingestion etc)
   
2. Is there anything in here that is being committed as an example? Do you anticipate removing anything in the PR for the final implementation?

OddTomBrooks:
> 

What do you anticipate is complete as is vs what do you anticipate changing in subsequent tickets?
- EG queries: These are updated to the platform-agnostic SearchRequest type
- Ingestion:  This is complete for the Search Change History feature set iteration. We may later expand this to create multiple indexes. If it becomes important to have more granularity than than 1 minute cron-driven ingestion cycles we will likely migrate to debezium + CDC.
- Endpoints: These are mostly complete and ready for use in various features. They are driven primarily by our discussions and requests. Notable exceptions are as follow.
  - SearchChangeData: This is not explicitly necessary but it is very useful for testing and allows the front-end to make an Elasticsearch DSL query if flexibility or functionality is required which is not (yet) supported by a custom back-end query.
  - SearchChangeTableDateHistogramConsolidatedAggregations: While this call is mostly complete, it is still a work in progress and has a few edge cases which must be addressed before it can be used in a feature set. Notably, dynamic paging is a problem as merged bucket groups can span across fixed page segments resulting in segmented and unpredictable limit lengths. This can be solved iteratively but goes beyond the scope of the current ticket as this is focused on proof of concept and the basic concept of merging aggregation buckets is now proven.
- Coding / Query Implementation: I am still going back and forth a bit to determine the appropriateness of a formatted query string versus marshalling a strict typed nested map structure. I am leaning towards a formatted query approach but felt it warranted a discussion so I left both variants for the time being.
  
Is there anything in here that is being committed as an example? Do you anticipate removing anything in the PR for the final implementation?
- There have been some removals already. I think we are approaching the trimmest variant of this I want to push out as even the endpoints which don't serve an explicit purpose at the moment are foundational for expanding or specializing on for faster feature implementation later.

pkg/server/routes.go
```diff
@@ -198,6 +200,15 @@ func (s *Server) routes(
 
 	// gql.Use(requirePrincipalMiddleware)
 
+	osConfig := opensearch.Config{
+		Addresses: []string{s.Config.GetString(appconfig.OpenSearchHostKey)},
+	}
+
+	osClient, err := opensearch.NewClient(osConfig)
+	if err != nil {
+		s.logger.Fatal("Failed to create an OpenSearch client", zap.Error(err))
```
ClayBenson94, line 209: We should probably either
- Make a feature flag that prevents the creation/use of the OpenSearch client
- Change this log to not be `Fatal`

because when we push this code to our deployed envs, right now this will likely fail.

config/logstash/logstash.conf
```diff
@@ -0,0 +1,51 @@
...
+    jdbc_driver_library => "/usr/share/logstash/postgresql-42.6.0.jar"
+    jdbc_driver_class => "org.postgresql.Driver"
+    jdbc_connection_string => "jdbc:postgresql://db/postgres"
+    jdbc_user => "postgres"
+    jdbc_password => "mysecretpassword"
+    schedule => "* * * * *"
+
+    tracking_column => "id"
+    use_column_value => true
+    statement => "SELECT audit.change.id, audit.change.table_id, audit.change.primary_key, audit.change.foreign_key, audit.change.action, CAST(audit.change.fields AS text) as fields, audit.change.modified_dts, CAST(row_to_json(user_account.*) as text) as modified_by, CONCAT(audit.change.table_id::text, '_', audit.change.primary_key::text) AS guid FROM audit.change LEFT JOIN user_account ON audit.change.modified_by = user_account.id WHERE audit.change.id > :sql_last_value"
```
ClayBenson94, line 12: Can you add a comment explaining this query? Things to call out:

- How it determines _what_ to ingest (`sql_last_value`)
- How it pulls user info (use of `user_account` table)

config/logstash/logstash.conf
```diff
@@ -0,0 +1,51 @@
+input {
+  jdbc {
+    jdbc_driver_library => "/usr/share/logstash/postgresql-42.6.0.jar"
+    jdbc_driver_class => "org.postgresql.Driver"
+    jdbc_connection_string => "jdbc:postgresql://db/postgres"
+    jdbc_user => "postgres"
+    jdbc_password => "mysecretpassword"
+    schedule => "* * * * *"
```
ClayBenson94, line 8: I'd probably just add a `# Every 1 minute` note here so folks who don't know cron as well know what this means!

config/logstash/logstash.conf
```diff
@@ -0,0 +1,51 @@
...
+  json {
+    source => "fields"
+    target => "fields"
+  }
+  json {
+    source => "modified_by"
+    target => "modified_by"
+  }
+
+  #metrics {
```
ClayBenson94, line 26: Do we still need this?

docker-compose.yml
ClayBenson94, line 1: image
It looks like the docker-compose had it's executable bits changed? Just want to confirm that this was on purpose - if not we might just want to revert this

.envrc
ClayBenson94, line 1: image
Same here - looks like the permissions changed - can this be reverted if it wasn't on purpose?

ClayBenson94 (changes requested):

StevenWadeOddball (changes requested):
I think this is looking great Tom! I just added a couple of comments to remember what we talked about in our discussions. Thanks for your hard work on this!

StevenWadeOddball (changes requested):
Thanks @OddTomBrooks , overall this looks great! I'm only requesting a change because all the postman queries are erroring. Otherwise, this looks great!
I get the below error message for every search query.
`*errors.errorString {s: "unsupported protocol scheme \"\""}`

I suspect it is due to the change to use the templating library?

OddTomBrooks:
> Thanks @OddTomBrooks , overall this looks great! I'm only requesting a change because all the postman queries are erroring. Otherwise, this looks great! I get the below error message for every search query. `*errors.errorString {s: "unsupported protocol scheme \"\""}`
> 
> I suspect it is due to the change to use the templating library?

I have not experienced anything like this. Let's have a sync today and figure out the issue.

ClayBenson94 (approved):
Thanks for addressing the feedback!

This looks good - let's keep a close eye on the deploy and validate in `dev` and `test`!

PR #522, opened 2023-04-03
Replace print statements with Zap Logger Statements

StevenWadeOddball:
## Changes and Description

-This addresses some print statements that should be log statements
   - Print statement don't get a trace id, so it makes debugging from splunk difficult.
- Some other spots could use the logger in the future, but would need to have some restructuring to pass the logger around.  (For example, the pub sub doesn't have a reference to the logger)
 - Print statements are left in test code, as we run it with test logger that doesn't log any statements.

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/pkg/graph/resolvers/plan_task_list_section_locks.go
+++ b/pkg/graph/resolvers/plan_task_list_section_locks.go
@@ -253,7 +253,7 @@ func onLockTaskListSectionUnsubscribeComplete(
 		_, err := UnlockTaskListSection(ps, modelPlanID, section, subscriber.GetPrincipal().Account().ID, model.ActionTypeNormal)
 
 		if err != nil {
-			fmt.Printf("Uncapturable error on websocket disconnect: %v\n", err.Error())
+			fmt.Printf("Uncapturable error on websocket disconnect: %v\n", err.Error()) //TODO: can we pass a reference to the logger to the pubsub?
 		}
 	}
 }
--- a/pkg/local/authentication_middleware.go
+++ b/pkg/local/authentication_middleware.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
-	"fmt"
 	"net/http"
 	"strings"
 
@@ -95,8 +94,6 @@ func devUserContext(ctx context.Context, authHeader string, store *storage.Store
 func NewLocalWebSocketAuthenticationMiddleware(logger *zap.Logger, store *storage.Store) transport.WebsocketInitFunc {
 	return func(ctx context.Context, initPayload transport.InitPayload) (context.Context, error) {
 		// Get the token from payload
-		all := initPayload
-		fmt.Println("ALL OF EM BABY", all)
 		any := initPayload["authToken"]
 		token, ok := any.(string)
 		if !ok || token == "" {
--- a/pkg/okta/authentication_middleware.go
+++ b/pkg/okta/authentication_middleware.go
@@ -250,8 +250,6 @@ type MiddlewareFactory struct {
 func (f MiddlewareFactory) NewOktaWebSocketAuthenticationMiddleware(logger *zap.Logger) transport.WebsocketInitFunc {
 	return func(ctx context.Context, initPayload transport.InitPayload) (context.Context, error) {
 		// Get the token from payload
-		all := initPayload
-		fmt.Println("ALL OF EM BABY (OKTA)", all)
 		any := initPayload["authToken"]
 		token, ok := any.(string)
 		if !ok || token == "" {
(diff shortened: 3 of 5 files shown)


pkg/graph/resolvers/plan_task_list_section_locks.go
```diff
@@ -253,7 +253,7 @@ func onLockTaskListSectionUnsubscribeComplete(
 		_, err := UnlockTaskListSection(ps, modelPlanID, section, subscriber.GetPrincipal().Account().ID, model.ActionTypeNormal)
 
 		if err != nil {
-			fmt.Printf("Uncapturable error on websocket disconnect: %v\n", err.Error())
+			fmt.Printf("Uncapturable error on websocket disconnect: %v\n", err.Error()) //TODO: can we pass a reference to the logger to the pubsub?
```
OddTomBrooks, line 256: This should be provided to pubsub on service creation rather than passed as a dependency in function calls.

pkg/local/authentication_middleware.go
```diff
@@ -95,8 +94,6 @@ func devUserContext(ctx context.Context, authHeader string, store *storage.Store
 func NewLocalWebSocketAuthenticationMiddleware(logger *zap.Logger, store *storage.Store) transport.WebsocketInitFunc {
 	return func(ctx context.Context, initPayload transport.InitPayload) (context.Context, error) {
 		// Get the token from payload
-		all := initPayload
-		fmt.Println("ALL OF EM BABY", all)
```
OddTomBrooks, line 99: I will miss this.

pkg/graph/resolvers/plan_task_list_section_locks.go
StevenWadeOddball: agreed! I just wanted to add a note for now

pkg/okta/authentication_middleware.go
```diff
@@ -261,12 +259,10 @@ func (f MiddlewareFactory) NewOktaWebSocketAuthenticationMiddleware(logger *zap.
 		// Parse auth header into JWT object
 		jwt, err := f.jwt(logger, token)
 		if err != nil {
-			fmt.Println("ERROR PARSING JWT", err)
-			// TODO How to error handle here?
+			logger.Warn("could not parse jwt from token", zap.Error(err))
```
OddTomBrooks, line 262: Outside the scope of this immediate work, we need to evaluate how to gracefully handle this error case.

pkg/server/routes.go
```diff
@@ -292,29 +291,30 @@ func (s *Server) routes(
...
 	if ok, _ := strconv.ParseBool(os.Getenv("DEBUG_ROUTES")); ok {
+		s.logger.Debug("debugging routes")
 		// useful for debugging route issues
 		_ = s.router.Walk(func(route *mux.Route, router *mux.Router, ancestors []*mux.Route) error {
 			pathTemplate, err := route.GetPathTemplate()
 			if err == nil {
-				fmt.Println("ROUTE:", pathTemplate)
+				s.logger.Debug("ROUTE: ", zap.String("ROUTE", pathTemplate))
 			}
 			pathRegexp, err := route.GetPathRegexp()
```
OddTomBrooks, line 301: NIT: Unrelated to this ticket but I normally see and expect "Regex", not "Regexp"

pkg/services/auth.go
```diff
@@ -28,15 +27,9 @@ func HasAnyRole(ctx context.Context, roles []model.Role) (bool, error) {
...
-	fmt.Println("APP CONTEXT PRINCIPAL", appcontext.Principal(ctx))
-	fmt.Println("APP CONTEXT PRINCIPAL", appcontext.Principal(ctx))
-	fmt.Println("APP CONTEXT PRINCIPAL", appcontext.Principal(ctx))
-	fmt.Println("APP CONTEXT PRINCIPAL", appcontext.Principal(ctx))
-	fmt.Println("APP CONTEXT PRINCIPAL", appcontext.Principal(ctx))
-	fmt.Println("APP CONTEXT PRINCIPAL", appcontext.Principal(ctx))
-	fmt.Println("APP CONTEXT PRINCIPAL", appcontext.Principal(ctx))
+
 	logger := appcontext.ZLogger(ctx)
+	logger.Info("app context principal", zap.Any("principal", appcontext.Principal(ctx)))
```
OddTomBrooks, line 32: I would like to standardize our logging format, specifically around usage of caps. It's all over the place. I propose we form a habit that as a simple pass that for short-form logs (less than a sentence) we use all lower case as done in this line.

OddTomBrooks (approved):
There is nothing problematic with this PR. There are a few optional items for improvement. If you decide to make these improvements please ping me for more immediate attention.

pkg/services/auth.go
StevenWadeOddball: I think I like that, or at least don't start the message with upper case. GO specifically complains about errors starting with upper case.  Could be good to standardize for sure! It's also interesting, the message here is really a little redundant, because it is logging the principal object field in the log as well.

pkg/server/routes.go
StevenWadeOddball: yeah, I agree. Looks like this came from the `*mux.Router` library, so that's probably why they kept the log statement the same.

PR #529, opened 2023-04-05
Easi 2788/audit deleted actor implementation

StevenWadeOddball:
# EASI-2788

## Changes and Description

- All delete statements are wrapped in transactions
   - first part of the transaction sets the current user scope

## How to test this change
1. Create objects and delete them. 
2. Look at the audit logs, verify that you see an audit entry that records the correct user who deleted the entry.

## PR Author Review Checklist

## PR Reviewer Guidelines


OddTomBrooks (approved):

pkg/storage/utitiltyStore.go
```diff
@@ -0,0 +1,24 @@
...
+	_ "embed"
+
+	"github.com/google/uuid"
+	"github.com/jmoiron/sqlx"
+)
+
+//go:embed SQL/utility/set_session_current_user.sql
+var setSessionCurrentUserSQL string
+
+// setCurrentSessionUserVariable sets the userID for
```
ClayBenson94, line 13: Can you complete this comment? Might also be good to clarify the purpose of this function, as it's not immediately clear what we might be using this session variable for right now

pkg/storage/utitiltyStore.go
```diff
@@ -0,0 +1,24 @@
...
+func setCurrentSessionUserVariable(tx *sqlx.Tx, userID uuid.UUID) error {
+	argsUser := map[string]interface{}{
+		"user_id": userID.String(),
+	}
```
ClayBenson94, line 17: Can this use the new `CreateUserIDQueryMap` function? I think they're the same thing here?

ClayBenson94 (changes requested):
Some small changes requested!

Only other thing I noted - I went through and identified all the possible `DELETE` statements we make in our SQL, and noted that https://github.com/CMSgov/mint-app/blob/ba4d0cf68eed3ed213722786b45fbd7fdde6b24b/pkg/storage/analyzed_auditStore.go#L59-L75 isn't covered here, but that that method is also never called anywhere.

I don't think we need to add that here, for the following reasons:
- It's never called right now
- If it were, I don't think a user id would exist when calling it

Perhaps we should add a small comment on here to clarify that we're purposefully not including the user Id there?

pkg/storage/utitiltyStore.go
StevenWadeOddball: good catch! thanks for spotting it

pkg/storage/utitiltyStore.go
StevenWadeOddball: sure! I changed it. I'm often on the fence about using query map helper functions. I do like the clarity of instantiating the map inline, but I think the name of the function should be sufficiently clear as well.

PR #541, opened 2023-04-17
[NOREF] Adding filters for ACTIVE or STAGED users

ClayBenson94:
# EASI-000

## Changes and Description

- Added filter to user search to only include "ACTIVE" or "STAGED" users

## How to test this change

- `scripts/dev up`
- In `routes.go` ensure that a real okta client is being created by changing the if statement to `if !s.environment.Local()`
- Create a new model plan
- Add collaborators, search for users that should show up (yourself, for example!) and users that we don't expect to show up (Shane Clark, for example)

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/pkg/oktaapi/client_wrapper.go
+++ b/pkg/oktaapi/client_wrapper.go
@@ -40,6 +40,7 @@ type oktaUserResponse struct {
 	DisplayName string `json:"displayName"`
 	Email       string `json:"email"`
 	Login       string `json:"login"`
+	SourceType  string `json:"SourceType"`
 }
 
 func (cw *clientWrapper) parseOktaProfileResponse(profile *okta.UserProfile) (*oktaUserResponse, error) {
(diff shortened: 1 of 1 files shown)


StevenWadeOddball (approved):
Thanks Clay, I pulled this locally and tested it. I think it makes sense and looks good.

I was able to confirm deactivated users no longer show up in the search, but regular users show up as expected.

OddTomBrooks (dismissed):

pkg/oktaapi/client_wrapper.go
```diff
@@ -100,13 +106,18 @@ func (cw *clientWrapper) SearchByName(ctx context.Context, searchTerm string) ([
...
+	users := []*models.UserInfo{}
+	for _, user := range searchedUsers {
 		profile, err := cw.parseOktaProfileResponse(user.Profile)
 		if err != nil {
 			return nil, err
 		}
-		users[idx] = profile
+
+		// If we find EUA users that have logins longer than 4 characters, they're a test user (don't add them to the array)
+		if len(profile.Login) > 4 {
```
StevenWadeOddball, line 117: @ClayBenson94 , can you speak to this approach? This will remove all the test accounts, but it also excludes all non - eua users from search. I _think_ that makes sense with what we were already doing with this search.

I wonder if it makes sense to update the name of the search to note we are only searching for EUA users?

StevenWadeOddball (dismissed):
Thanks Clay, this looks good, I tested it and confirmed I couldn't see test user accounts. I had one tangentially related thought around non-EUA users I provided below. It's non-blocking, but maybe something we should consider in the future.

pkg/oktaapi/client_wrapper.go
ClayBenson94: Thoughts on this approach?
```go
	users := []*models.UserInfo{}
	for _, user := range searchedUsers {
		profile, err := cw.parseOktaProfileResponse(user.Profile)
		if err != nil {
			return nil, err
		}

		// If we find EUA users that have logins longer than 4 characters, they're a test user (don't add them to the array)
		if (profile.SourceType == euaSourceType || profile.SourceType == euaADSourceType) && len(profile.Login) > 4 {
			continue
		}
		users = append(users, profile.toUserInfo())
	}
```

ClayBenson94: This way the function stays agnostic to what we're searching, but it only filters out the EUA users? That way if we later add MAC/IDM User searching, this doesn't conflict with that

ClayBenson94: c762b413aceefc2fd5a4b5b4bd85be04523afdd6

StevenWadeOddball (approved):
🚀

PR #552, opened 2023-04-28
[EASI-2792] Search Change History GQL Result Typing

OddTomBrooks:
# EASI-2792

## Changes and Description

- Created a new type structure to contain search results

## How to test this change

Run any of the search endpoints. Verify that they work and return the changed fields as an array.

## PR Author Review Checklist

## PR Reviewer Guidelines


pkg/graph/schema.graphql
```diff
@@ -1357,13 +1357,30 @@ type OperationalSolutionSubtask {
...
+type ChangedFields {
+  changes: [Field!]!
+}
+
+type Field {
+  name: String!
+  value: ChangedFieldValue!
+}
+
+union ChangedFieldValue = ChangedFields | FieldValue
```
StevenWadeOddball, line 1369: @OddTomBrooks, would it make sense for this to be FieldValue and not a ChangdFields? We don't have a nested change hierarchy, so I'm not sure having this nested makes sense.

pkg/graph/schema.graphql
```diff
@@ -1357,13 +1357,30 @@ type OperationalSolutionSubtask {
...
+
+type FieldValue {
+  new: String
```
StevenWadeOddball, line 1372: What do you think about making `new` and `old` interface types? Not all values are strings.

pkg/models/change_table_record.go
```diff
@@ -12,10 +12,39 @@ import (
...
 	ForeignKey  *uuid.UUID                  `json:"foreign_key"`
 	Action      string                      `json:"action"`
-	Fields      map[string]interface{}      `json:"fields"`
+	Fields      ChangedFields               `json:"fields"`
 	ModifiedDts *time.Time                  `json:"modified_dts"`
 	ModifiedBy  *authentication.UserAccount `json:"modified_by"`
 }
+
+// ChangedFieldValue represents an interface for changed field values.
+type ChangedFieldValue interface {
```
StevenWadeOddball, line 25: Is this to support the union type?

pkg/models/change_table_record.go
```diff
@@ -12,10 +12,39 @@ import (
...
+
+// Field represents an individual field, including its name and associated value.
+type Field struct {
+	Name  string            `json:"name"`
+	Value ChangedFieldValue `json:"value"`
+}
+
+// FieldValue represents the new and old values of a changed field.
+type FieldValue struct {
+	New *string `json:"new"`
```
StevenWadeOddball, line 45: same thing here, this might make more sense as an interface type. Thoughts?

pkg/graph/schema.graphql
OddTomBrooks: This is something I thought about as well. It was originally designed with support for nested fields in mind. Rather than toss that work out I wanted to maintain it in case we needed it moving forward as it works the same either way. In fact, our processing on the server side does not currently support nested change fields right now - that would be the only change necessary I believe.

StevenWadeOddball: Thanks for the clarification Tom! I think it's clearer to remove the union type. I think leaving it there suggests that fields could be nested, but we don't have any nested fields, and I don't see how that would change (unless we were to drastically overhaul how we are doing auditing).

StevenWadeOddball (changes requested):
Thanks for your work on this Tom, I think it looks good. I think it makes the most sense to remove the union type, sense our data doesn't require it and likely won't require it. I also am curious about using `interface` instead of string for `New` and `Old`

pkg/models/change_table_record.go
OddTomBrooks: Yes

pkg/models/change_table_record.go
OddTomBrooks: Same as the other comment

StevenWadeOddball (approved):
Thanks Tom! 🚀

PR #555, opened 2023-05-01
Use multi-arch images when building locally for migrate and mailcatcher, use alpine when building for ECR

ClayBenson94:
# NOREF

## Changes and Description

- Modifies image tags used for mailcatcher to use 0.8.2, which is multi-arch (where the previous image was not)
- Modifies the Dockerfile for db migrations to support a build argument that allows modifying the tag used for the base Flyway image. This allows using the multi-arch image locally while also allowing the same dockerfile to build the alpine image to push to ECR.

The main goal for this PR was to reduce the frequency with which the DB Migrate container would fail to start locally. After some investigation, we're lead to believe that it's due to spinning up an amd64 image on arm computers (M1 Mac, for example). This isn't a guaranteed fix, but this is a good change to make even if it's not the fix for the hanging migrate cnotainers.

## How to test this change

- `scripts/dev up`
- Ensure there are no warnings in the console about targeting the wrong platform when building
- Ensure the DB Migrate image starts and runs as expected
- Ensure mailcatcher starts and runs as expected

## PR Reviewer Guidelines

### DIFF
--- a/.github/workflows/build_application_images.yml
+++ b/.github/workflows/build_application_images.yml
@@ -78,6 +78,8 @@ jobs:
           platforms: "linux/amd64"
           context: .
           cache-from: type=gha,scope=${{ env.GIT_REF_NAME }}-db_migrate
+          build-args: |
+            TAG=9.10-alpine
       - name: Announce failure
         if: ${{ failure() }}
         run: |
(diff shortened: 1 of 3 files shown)


StevenWadeOddball (approved):
This looks good to me! Thanks @ClayBenson94! I pulled this locally and brought the app up and down a few times. The containers started consistently, and ran faster than with the old image.

BuildEverything:
I have restarted the docker containers several times on WSL2 with no apparent issue.

OddTomBrooks (approved):
I have restarted the docker containers several times with no apparent issue.

PR #557, opened 2023-05-01
Easi 2592/mint assessment permissions overhaul

StevenWadeOddball:
# EASI-2592

## Changes and Description

- Assessment User now gets granted normal user role
- Unit tests to assert this new logic.
- I looked through the front end code to see if there needs to be changes there, It looks like this covers all the needs.
- 

## How to test this change
1. Log into the app / Run postman queries with only the ASSESSMENT Job code
2. Confirm that you can hit endpoints that require the mint user job code (like modelPlan)
3. Verify Unit test coverage matches as expected
4.  Verify that there aren't missed areas on the frontend that need code changes to work with this paradigm.

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/pkg/graph/resolvers/search_change_table.go
+++ b/pkg/graph/resolvers/search_change_table.go
@@ -33,6 +33,7 @@ const (
 )
 
 // Embed the template files
+//
 //go:embed searchquerytemplates/free_text.tmpl
 var freeTextSearchTmpl string
 
--- a/pkg/local/authentication_middleware.go
+++ b/pkg/local/authentication_middleware.go
@@ -73,11 +73,20 @@ func devUserContext(ctx context.Context, authHeader string, store *storage.Store
 	if parseErr := json.Unmarshal([]byte(devUserConfigJSON), &config); parseErr != nil {
 		return nil, errors.New("could not parse local auth JSON")
 	}
+	var jcUser bool
+	jcAssessment := swag.ContainsStrings(config.JobCodes, "MINT_ASSESSMENT_NONPROD")
+	if jcAssessment { //Assessment users automatically are granted the base user permissions
+		jcUser = true
+	} else {
+		jcUser = swag.ContainsStrings(config.JobCodes, "MINT_USER_NONPROD")
+	}
+	jcMAC := swag.ContainsStrings(config.JobCodes, "MINT MAC Users")
+
 	princ := &authentication.ApplicationPrincipal{
 		Username:          strings.ToUpper(config.EUA),
-		JobCodeUSER:       swag.ContainsStrings(config.JobCodes, "MINT_USER_NONPROD"),
-		JobCodeASSESSMENT: swag.ContainsStrings(config.JobCodes, "MINT_ASSESSMENT_NONPROD"),
-		JobCodeMAC:        swag.ContainsStrings(config.JobCodes, "MINT MAC Users"),
+		JobCodeUSER:       jcUser,
+		JobCodeASSESSMENT: jcAssessment,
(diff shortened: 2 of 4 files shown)


pkg/graph/resolvers/search_change_table.go
```diff
@@ -33,6 +33,7 @@ const (
 )
 
 // Embed the template files
+//
```
ClayBenson94, line 36: Can you revert this? (This is the definition of a nit pick 😂)

pkg/local/authentication_middleware.go
```diff
@@ -73,11 +73,20 @@ func devUserContext(ctx context.Context, authHeader string, store *storage.Store
...
 	}
+	var jcUser bool
+	jcAssessment := swag.ContainsStrings(config.JobCodes, "MINT_ASSESSMENT_NONPROD")
+	if jcAssessment { //Assessment users automatically are granted the base user permissions
+		jcUser = true
+	} else {
+		jcUser = swag.ContainsStrings(config.JobCodes, "MINT_USER_NONPROD")
+	}
+	jcMAC := swag.ContainsStrings(config.JobCodes, "MINT MAC Users")
+
```
ClayBenson94, line 84: This could be simplified, but functionally is the same! Keeps the `swag` lines together, then later overrides it. I could go either way, but thought I'd share the suggestion!

```suggestion
	// Pull job codes from config
	jcUser := swag.ContainsStrings(config.JobCodes, "MINT_USER_NONPROD")
	jcAssessment := swag.ContainsStrings(config.JobCodes, "MINT_ASSESSMENT_NONPROD")
	jcMAC := swag.ContainsStrings(config.JobCodes, "MINT MAC Users")
	
	// Always set assessment users to have base user permissions
	if jcAssessment {
		jcUser = true
	}
```

pkg/okta/authentication_middleware_test.go
```diff
@@ -68,7 +68,7 @@ func (tjv TestJwtVerifier) VerifyAccessToken(jwt string) (*jwtverifier.Jwt, erro
 	return tjv.verify(jwt)
 }
 
-func (s *AuthenticationMiddlewareTestSuite) buildMiddleware(verify func(jwt string) (*jwtverifier.Jwt, error)) func(http.Handler) http.Handler {
+func (s *AuthenticationMiddlewareTestSuite) buidMiddleWareFactory(verify func(jwt string) (*jwtverifier.Jwt, error)) *MiddlewareFactory {
```
ClayBenson94, line 71:
```suggestion
func (s *AuthenticationMiddlewareTestSuite) buildMiddleWareFactory(verify func(jwt string) (*jwtverifier.Jwt, error)) *MiddlewareFactory {
```

(Don't commit this directly, as it'll only rename the definition, not all the places it's used)

ClayBenson94 (changes requested):
Very small changes - great job on this!

pkg/graph/resolvers/search_change_table.go
StevenWadeOddball: @ClayBenson94, this one is a weird one! I actually introduced it because the linter is complaining about it (even though I didn't touch the file 😂 . So I'd have to force the commit if I don't commit this change

pkg/local/authentication_middleware.go
StevenWadeOddball: that does read nicer! Thanks!

OddTomBrooks (approved):

ClayBenson94 (approved):

PR #568, opened 2023-05-09
NOREF Update Operational Need Triggers

StevenWadeOddball:
## Changes and Description

- Correct triggers for
    - Update the plan’s contract
    - Educate beneficiaries
- Calculate the if needs are needed again for all existing needs based on question answers.

## How to test this change

1. Bring up the App and make a Model.
2.  Answer the question   ` Will the model have a learning strategy?` in plan ops, eval, and learning, with any below answer ( or combination) and ensure the need is shown as needed
    - `We plan to educate beneficiaries`
    - `We plan to use an IT platform (Connect)`
    - `We plan to have a learning contractor (cross-model or individual)`
    - `We plan to enable participant-to-participant collaboration`
    - `Other`
3. Answer the question   `Have you updated the plan’s contract?` in general characteristics with yes and ensure the need is shown as needed

## PR Author Review Checklist

## PR Reviewer Guidelines


pkg/graph/resolvers/operational_need_triggers_test.go
```diff
@@ -39,7 +40,8 @@ func (suite *ResolverSuite) TestGeneralCharacteristicsNeeds() {
 
 	changes = map[string]interface{}{
 		"managePartCDEnrollment": true,  // NEED 1, MANAGE_CD
-		"collectPlanBids":        false, // NEED 2, MANAGE_CD &&  3 UPDATE_CONTRACT
+		"collectPlanBids":        false, // NEED 2, MANAGE_CD
+		"planContractUpdated":    false, // NEED 3 UPDATE_CONTRACT
```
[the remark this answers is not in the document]
StevenWadeOddball: My pleasure! Thanks Clay

migrations/V90__Need_Trigger_Updates.sql
ClayBenson94, line 1: I think this conflicts with `V90__Existing_Model_Link_Table.sql`. Guessing this should be renamed to 92?

ClayBenson94 (changes requested):

migrations/V90__Need_Trigger_Updates.sql
StevenWadeOddball: Thanks! Yes, I think that PR was merged into main this morning. I'll update it

ClayBenson94 (approved):

pkg/graph/resolvers/operational_need_triggers_test.go
```diff
@@ -15,7 +15,8 @@ func (suite *ResolverSuite) TestGeneralCharacteristicsNeeds() {
 
 	changes := map[string]interface{}{
 		"managePartCDEnrollment": false, // NEED 1, MANAGE_CD
-		// "collectPlanBids":        true, // NEED 2, MANAGE_CD &&  3 UPDATE_CONTRACT
+		// "collectPlanBids":        true, // NEED 2, MANAGE_CD
+		// "planContractUpdated":        false, // NEED 3 UPDATE_CONTRACT
```
OddTomBrooks, line 19: Is this intended to be kept?

migrations/V92__Need_Trigger_Updates.sql
```diff
@@ -0,0 +1,46 @@
...
+    modified_dts = current_timestamp
+
+WHERE need_key = 'UPDATE_CONTRACT';
+
+
+
+UPDATE possible_operational_need
+SET
+    trigger_table = 'plan_ops_eval_and_learning', --same as previous
+    trigger_col = '{model_learning_systems}',  --same as previous
```
OddTomBrooks, line 19: What is --same as previous referring to?

StevenWadeOddball:
> I feel like these tests should effectively mimic/test the mapping of question -> op need (but idk if we want to do that for _every_ case - I'll leave it up to you, but otherwise the changes look good!)

Thanks @ClayBenson94 , I can see the benefit of that, but I think for now, I'm going to leave it with the current paradigm in the interest of unblocking the release cadence. These generally are just verifying the logic at this point, and I think the UX review was verifying the configuration.

If we come back to update the operational need trigger in the future, I think it might be worthwhile at thinking of a paradigm to test the configuration logic as well. For now, I think that we can have reasonable confidence in the accuracy of the trigger, and they shouldn't change without an explicit migration to change them.

migrations/V92__Need_Trigger_Updates.sql
```diff
@@ -0,0 +1,46 @@
+
+/* Update existing needs */
+UPDATE possible_operational_need
+SET
+    
+    trigger_table = 'plan_general_characteristics', --same as previous
```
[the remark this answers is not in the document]
StevenWadeOddball: this is denote the configuration is the same as the previous configuration. I put it in so it's easier to understand the configuration when looking at the migrations by themselves.

PR #572, opened 2023-05-10
Operational solution is common field

StevenWadeOddball:
## Changes and Description

Backend
- Operational Solution has a new field `isCommonSolution`
    - If this is listed as linked to the need in the `possible_need_solution_link`, it returns `true`, else `false` 
    -Postman queries updated
    - Unit test to assert Logic

FrontEnd

## PR Author Review Checklist

## PR Reviewer Guidelines


pkg/storage/SQL/operational_solution/and_possible_get_by_operational_need_id_LOADER.sql
```diff
@@ -29,7 +29,8 @@ possibleSolution AS (
         COALESCE(OpNd.created_by, '00000000-0000-0000-0000-000000000000') AS created_by, -- This is UUID.NIL, the same as the UNKNOWN_USER account in the DB
         COALESCE(OpSol.created_dts, CURRENT_TIMESTAMP) AS created_dts,
         OpSol.modified_by,
-        OpSol.modified_dts
+        OpSol.modified_dts,
+        TRUE AS is_common_solution
```
ClayBenson94, line 33: Can you add comments to these? I think it might just help clarify exactly how this is determined in each case. Maybe for here, something like

```suggestion
        TRUE AS is_common_solution -- Possible Solutions are always a common solution
```

pkg/storage/SQL/operational_solution/get_by_id.sql
```diff
@@ -16,8 +16,11 @@ SELECT
     OpSol.created_by,
     OpSol.created_dts,
     OpSol.modified_by,
-    OpSol.modified_dts
+    OpSol.modified_dts,
+    PNSL.id IS NOT NULL AS is_common_solution
```
ClayBenson94, line 20: Continuing from my previous note, I think comments here would be useful. Looking over this, I am not exactly sure why `is_common_solution` is based no whether or not a Possible Need Solution Link's ID isn't NULL (writing that out shows how complex this can be!)

pkg/models/operational_solution.go
```diff
@@ -10,18 +10,19 @@ import (
...
+	Needed           *bool                   `json:"needed" db:"needed"` // Null means that it is not an actual record
+	Name             *string                 `json:"name" db:"sol_name"`
+	Key              *OperationalSolutionKey `json:"key" db:"sol_key"`
+	NameOther        *string                 `json:"nameOther" db:"name_other"`
+	PocName          *string                 `json:"pocName" db:"poc_name"`
+	PocEmail         *string                 `json:"pocEmail" db:"poc_email"`
+	MustStartDts     *time.Time              `json:"mustStartDts" db:"must_start_dts"`
+	MustFinishDts    *time.Time              `json:"mustFinishDts" db:"must_finish_dts"`
+	IsOther          *bool                   `json:"isOther" db:"is_other"`
+	IsCommonSolution *bool                   `json:"isCommonSolution" db:"is_common_solution"` //This is determined by the query, and not stored in the operational_solution table
```
ClayBenson94, line 23: Just adding some additional clarity to this comment!

```suggestion
	IsCommonSolution *bool                   `json:"isCommonSolution" db:"is_common_solution"` // This is returned by the query, though it is not stored in the operational_solution table, so we still map it here with a "db:" struct tag
```

ClayBenson94 (changes requested):
Most of these changes make sense, just adding some comments around the clarity of the code!

The only other thing I'd note, is that the naming of this is slightly confusing: Are we saying that anything that `IsCommonSolution` is one that's a solution for a possible operational need? If so, we might be able to work on the naming to clarify that a bit more, but I want to make sure my understanding is correct first!

patrickseguraoddball (changes requested):
Adding the same 'other' solution, if it is a common solution, creates the duplicate as a common solution when it should be created as other

Screenshot 2023-05-11 at 9 19 57 AM

StevenWadeOddball:
> > Most of these changes make sense, just adding some comments around the clarity of the code!
> > The only other thing I'd note, is that the naming of this is slightly confusing: Are we saying that anything that `IsCommonSolution` is one that's a solution for a possible operational need? If so, we might be able to work on the naming to clarify that a bit more, but I want to make sure my understanding is correct first!
> 
> @ClayBenson94 The 'Common' solution comes from here for reference. It refers to only the default set of 'common' solutions for that particular need. None can be user-added/removed. If adding an 'other' solution or a solution that has preset details, those would go in "Other" solutions.
> 
> Screenshot 2023-05-11 at 9 00 19 AM

Thanks @patrickseguraoddball  and @ClayBenson94, yes, I think some of the confusion here is that I named the field with how it is referenced on the front end form, but we don't use the same terminology in backend.

@patrickseguraoddball , I have to double check the behavior with the the `treat_as_other` solutions like contract. For needs that have a solution, I would expect it to show up  like this. It is checking to see if the solution type is linked as a possible solution for the need. I don't think it would make sense to change that field value in the backend return value, because it really is a common solution. Do you think it would make sense to key off other fields as well here? ( EG if `is_other` and `is_common_solution` and `id` is not null, put them in the other solution section?)

pkg/storage/SQL/operational_solution/and_possible_get_by_operational_need_id_LOADER.sql
StevenWadeOddball: Good call, thanks Clay.

pkg/storage/SQL/operational_solution/get_by_id.sql
StevenWadeOddball: Good call! Thanks Clay, I'll add the comments elsewhere as well.

patrickseguraoddball (approved):
LG from FE!

PR #577, opened 2023-05-19
Easi 2900/temporal tables documentation

StevenWadeOddball:
# EASI-2900 Temporal Table Research

## Changes and Description

- [POC branch](https://github.com/CMSgov/mint-app/tree/EASI-2900/temporal_tables_poc).
- A write up of temporal table research [here](https://github.com/CMSgov/mint-app/blob/EASI-2900/temporal_tables_documentation/docs/temporal_tables.md)

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/docs/temporal_tables.md
+++ b/docs/temporal_tables.md
@@ -2,14 +2,14 @@
 
 ## Background
 
-Temporal tables are database tables that keep track of changes to data over time. They are commonly used in Microsoft SQL Server to maintain an audit trail of data changes. While PostgreSQL does not have native support for temporal tables, the same behavior can be replicated by using triggers to capture changes and store historical data in separate tables. As compared to a generic implementation, a temporal table utilizes a typed record that shows the state of a row at any given instant in time, instead of just saying what was changes.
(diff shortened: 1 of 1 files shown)


docs/temporal_tables.md
```diff
@@ -0,0 +1,55 @@
+# Utilization of temporal tables
+
+## Background
+
+Temporal tables are database tables that keep track of changes to data over time. They are commonly used in Microsoft SQL Server to maintain an audit trail of data changes. While PostgreSQL does not have native support for temporal tables, the same behavior can be replicated by using triggers to capture changes and store historical data in separate tables. As compared to a generic implementation, a temporal table utilizes a typed record that shows the state of a row at any given instant in time, instead of just saying what was changed.
```
OddTomBrooks, line 5:
```suggestion
Temporal tables are database tables that keep track of changes to data over time. They are commonly used in Microsoft SQL Server to maintain an audit trail of data changes. While PostgreSQL does not have native support for temporal tables, the same behavior can be replicated by using triggers to capture changes and store historical data in separate tables. As compared to a generic implementation, a temporal table utilizes a typed record that shows the state of a row at any given instant in time, instead of just saying what was changed. This is useful in conjunction with the audit change table to describe not just the delta slices, but express the state of our data at any point in its evolution.
```

It may be worth adding context such as this to describe the reason we would want to use audit tables in addition to what they are.

OddTomBrooks (approved):

docs/temporal_tables.md
StevenWadeOddball: Thanks Tom, I'm meaning for this to be a bit more generic in this section, so I think I'm going to leave that out. I know we also had the idea of combining the delta slice in the temporal table (vs having a separate table), but I'd like to leave this generic.

PR #580, opened 2023-05-24
[EASI-2793] Filter-Based Dynamic Query Generation

OddTomBrooks:
# EASI-2793

## Changes and Description

- Created dynamic query builder
  - Extensible based on 'Filter Handlers'
- Added concept of 'filters' and corresponding filter based search endpoint on the backend
- Abstracted dynamic query building logic into shared and project-impl elements
- Unit test coverage on query builder

## PR Author Review Checklist

## PR Reviewer Guidelines


pkg/graph/resolvers/search_change_table.go
```diff
@@ -276,6 +280,86 @@ func SearchChangeTable(
...
+
+	for _, searchFilter := range searchFilters {
+		err2 := qb.AddFilter(string(searchFilter.Type), searchFilter.Value)
+		if err2 != nil {
+			return nil, err
+		}
+	}
+
+	// Build the query
+	query := qb.Build().Size(uint64(limit)).From(uint64(offset))
```
OddTomBrooks, line 342: Move pagination within the query builder. a call to qb.Build() should generate a finalized query.

pkg/graph/resolvers/search_change_table.go
```diff
@@ -276,6 +280,86 @@ func SearchChangeTable(
...
+	}
+
+	// Debug print the query
+	logger.Debug("Search query", zap.String("query", string(queryJSON)))
+	print(string(queryJSON))
+
+	// Convert the JSON string to a reader
+	queryReader := strings.NewReader(string(queryJSON))
+
+	return searchChanges(logger, searchClient, queryReader, limit, offset, "modified_dts:desc")
```
OddTomBrooks, line 360: Add sorting parameters to the endpoing. One string or enumeration for the element to sort on and one enumeration for Ascending or Descending mode. These should be composed in a SortingConfiguration (or some other such name) type.

@patrickseguraoddball Do you prefer an enumeration (which may constrain the elements available to sort on until a backend updates to add elements to the enum) or do you prefer to pass a simple string such as "modified_dts", which requires you to pass the name of the field but allows you to search on any field?

pkg/graph/schema.graphql
```diff
@@ -1457,6 +1462,8 @@ type Query {
   @hasAnyRole(roles: [MINT_USER, MINT_MAC])
   existingModelLink(id: UUID!): ExistingModelLink!
   @hasAnyRole(roles:[MINT_USER, MINT_MAC])
+  searchChanges(filters: [SearchFilter!], limit: Int!, offset: Int!): [ChangeTableRecord!]!
```
OddTomBrooks, line 1465: Maybe this should be named something more expressive. Something like searchFilteredChanges? Thoughts or ideas?

OddTomBrooks:
Postman collection to be updated before upgrading this to full PR

pkg/graph/schema.graphql
StevenWadeOddball: searchChanges works for me! I think the params make it fairly clear that you provide a filter, so I think that works.

pkg/graph/schema.graphql
```diff
@@ -1457,6 +1462,8 @@ type Query {
   @hasAnyRole(roles: [MINT_USER, MINT_MAC])
   existingModelLink(id: UUID!): ExistingModelLink!
   @hasAnyRole(roles:[MINT_USER, MINT_MAC])
+  searchChanges(filters: [SearchFilter!], limit: Int!, offset: Int!): [ChangeTableRecord!]!
+  @hasAnyRole(roles: [MINT_USER, MINT_MAC])
   searchChangeTable(request: SearchRequest!, limit: Int!, offset: Int!): [ChangeTableRecord!]!
```
StevenWadeOddball, line 1467: I think it would make sense to remove the other queries at this point.

StevenWadeOddball (commented):
This looks pretty great Tom! I just added some thoughts, I'll test it locally with your testing steps when this is promoted. Great work tackling this 🚀 !

pkg/factories/query_builder_factory.go
```diff
@@ -0,0 +1,48 @@
+package factories
```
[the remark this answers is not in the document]
OddTomBrooks: As we discussed, this is unfortunately more related to Faktory being named something so ambiguous, there was an intentional choice to name this package "factories" rather than "factory". However, in keeping with Go idiomatic package naming, we have renamed this package to be more specific to query_builder. This is therefore incidentally resolved.

pkg/graph/schema.graphql
```diff
@@ -1416,6 +1416,11 @@ type DateHistogramAggregationBucket {
   minModifiedDts: Time!
 }
 
+input SearchFilter {
+  type: SearchFilterType!
+  value: Any!
+}
```
ClayBenson94, line 1422: Per our convo yesterday, I think it would greatly help the clarity of this schema to type the values for the different kinds of filters you can apply (or, at least document it in a comment on this input type)

For example, constructing a `CHANGED_AFTER` search - What format should I use for the date? Same as `scalar Time`? RFC3339? For a `MODEL_PLAN_SECTION`, should I pass an array of sections? Should they be strings with the same value as this enum? https://github.com/CMSgov/mint-app/blob/e4a024fc0a4af0e65bae9e51bd9b6b76759845ff/pkg/graph/schema.graphql#L1644-L1652

These are the kinds of questions that are helped by more strict typing, but in the short term if we can just document the expected input types and formats, that can be sufficient to push this through for the FE in my opinion

ClayBenson94 (dismissed):

pkg/graph/schema.graphql
OddTomBrooks: As discussed, further work on this is ticketed in EASI-2999.

StevenWadeOddball:
@OddTomBrooks , I think that LINTER error is just because of the dependabot PR for gqlgen. I think it should resolve if you run GQLgen again

pkg/querybuilderfilters/model_plan_section_filter.go
```diff
@@ -0,0 +1,76 @@
...
+			section := model.SearchableTaskListSection("")
+			if err := section.UnmarshalGQL(val); err != nil {
+				return err
+			}
+			if err := f.addSectionToQuery(section, sectionQuery); err != nil {
+				return err
+			}
+		}
+	case interface{}:
+		section := model.SearchableTaskListSection("")
```
StevenWadeOddball, line 49: Nice, I think this is a pretty clever way to validate the values.

pkg/graph/schema.graphql
StevenWadeOddball: I missed that you did do this in the resolver, so I think that this is could be ok as is. It's probably good to deprecate it or remove them sometime though.

pkg/querybuilderfilters/model_plan_section_filter.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+}
+
+// HandleFilter handles the filter value for the ModelPlanSectionFilter
+func (f *ModelPlanSectionFilter) HandleFilter(value interface{}) error {
+	sectionQuery := esquery.Bool()
+
+	switch v := value.(type) {
+	case []interface{}:
+		for _, val := range v {
```
StevenWadeOddball, line 39: Do you think that table_name should have this treatment as well? It probably can be added on later if needed. Right now, I couldn't specifically request plan_documents and model_plan changes only for example (because they aren't part of the task list, and putting in two table name parameter returns an empty array.)

pkg/graph/schema.graphql
```diff
@@ -2215,3 +2233,105 @@ enum ActionType {
...
+  TABLE_ID
+
+  """
+  Sort by the table name that was changed
+  """
+  TABLE_NAME
+}
+
+# lint-disable defined-types-are-used
+enum SearchableTaskListSection {
```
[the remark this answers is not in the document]
StevenWadeOddball: This could also be ok potentially just because there is the table_name query as well, however that doesn't allow for querying multiple tables at once.

StevenWadeOddball (approved):
This looks pretty good Tom, thanks for your hard work. I provided some feedback that could be implemented, but I don't think anything is blocking merging this in it's current state as there is a follow up ticket.

PR #590, opened 2023-06-06
[EASI-2899] Add AMS Model ID and other fields to General Characteristics

OddTomBrooks:
# EASI-2899

## Changes and Description

- Add fields to General Characteristics
  - AMS Model ID
  - Demo Code
  - Model Abbreviation
- Updated Store SQL to include new fields on General Characteristics
- There does not appear to be DB seeding for General Characteristics. That is outside the scope of the ticket so I have omitted it for now. This is a candidate for Tech Debt.

## How to test this change

- Run `scripts/dev up:backend && scripts/dev db:seed`
- Observe that the corresponding fields have been added to General Characteristics

## PR Author Review Checklist

## PR Reviewer Guidelines


migrations/V94__Update_General_Characteristics_AMSModelID_Abbreviation_DemoCode.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE plan_general_characteristics
+  ADD COLUMN demo_code zero_string,
+  ADD COLUMN ams_model_id uuid,
+  ADD COLUMN model_abbreviation zero_string;
```
StevenWadeOddball, line 4: @OddTomBrooks , did you think about including `model_abbreviation` on `model_plan`?  I think that it could make more sense there in terms of data organization (since it's an abbreviation of model name)

OddTomBrooks: I did think about it. Rationally it made sense to me for it to go on either general characteristics or model plan. Since I didn't have a strong feeling either way, I simply added it to general characteristics as it incurred fewer file changes. If you feel strongly about this I can migrate this field to model plan.

StevenWadeOddball: Thanks @OddTomBrooks , I think I prefer it in `model_plan`, but  `general_characterstics` could be ok. If you and @ClayBenson94 prefer general characteristics, that's ok. Thoughts?

StevenWadeOddball (changes requested):
This looks good! It just looks like you need to add model_abbreviation in schema.graphql's `model_plan` section.

pkg/graph/schema.resolvers.go
```diff
@@ -44,6 +44,11 @@ func (r *existingModelLinkResolver) CurrentModelPlan(ctx context.Context, obj *m
 	return resolvers.ModelPlanGetByIDLOADER(ctx, *obj.CurrentModelPlanID) //TODO, implement loader, or this will be many queries
 }
 
+// Abbreviation is the resolver for the abbreviation field.
+func (r *modelPlanResolver) Abbreviation(ctx context.Context, obj *models.ModelPlan) (*string, error) {
+	panic(fmt.Errorf("not implemented: Abbreviation - abbreviation"))
```
StevenWadeOddball, line 49: We have already synced about this, but just dropping a note so I get notified when you  have it updated! The field is named differently in model_plan.go, so this doesn't auto-resolve.

StevenWadeOddball (changes requested):

pkg/graph/schema.resolvers.go
```diff
@@ -1052,6 +1047,6 @@ type subscriptionResolver struct{ *Resolver }
 //  - When renaming or deleting a resolver the old code will be put in here. You can safely delete
 //    it when you're done.
 //  - You have helper methods in this file. Move them out to keep these resolver files clean.
-func (r *planGeneralCharacteristicsResolver) AmsModelID(ctx context.Context, obj *models.PlanGeneralCharacteristics) (*uuid.UUID, error) {
-	panic(fmt.Errorf("not implemented: AmsModelID - amsModelID"))
+func (r *modelPlanResolver) Abbreviation(ctx context.Context, obj *models.ModelPlan) (*string, error) {
```
StevenWadeOddball, line 1050: Can you remove this old resolver code please?

pkg/models/model_plan.go
```diff
@@ -5,9 +5,10 @@ import "github.com/google/uuid"
 // ModelPlan is the top-level object for an entire draft model plan
 type ModelPlan struct {
 	baseStruct
-	ModelName string      `json:"modelName" db:"model_name"`
-	Archived  bool        `json:"archived" db:"archived"`
-	Status    ModelStatus `json:"status" db:"status"`
+	ModelName    string      `json:"modelName" db:"model_name"`
+	Abbreviation *string     `json:"abbreviation" db:"abbreviation"`
```
StevenWadeOddball, line 9: Since you changed this to be just abbreviation, all the SQL will need to be updated as well, or the db tag needs to be `model_abbreviation`. I believe this is why db seed and the tests are failing.

StevenWadeOddball: I think it's probably best practice to just use the same field name in all spots (eg update db to just use abbreviation if that's what you want to use on the model)

StevenWadeOddball (changes requested):
Thanks @OddTomBrooks, this is almost there! Can you take a look at db seed and model_planTest.go? Could you also add these new fields to the postman collection please?

migrations/V94__Update_General_Characteristics_AMSModelID_DemoCode.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE plan_general_characteristics
+  ADD COLUMN demo_code zero_string,
+  ADD COLUMN ams_model_id uuid;
```
ClayBenson94, line 3: Do we know if the AMS Model ID is a UUID? I'm not sure we know the format of this quite yet, and it might make sense to just have this be `zero_string` as well, since it looks like the input type in GQL is already just `String`

ClayBenson94: (Same for the Go struct type)

pkg/graph/schema.graphql
```diff
@@ -431,6 +432,8 @@ PlanGeneralCharacteristics represents a plan general characteristics object
 type PlanGeneralCharacteristics {
   id: UUID!
   modelPlanID: UUID!
+  demoCode: String
+  amsModelID: String
```
ClayBenson94, line 436: I think we will need to add these fields to [PlanGeneralCharacteristicsChanges](https://github.com/CMSgov/mint-app/pull/590/files#diff-7b3b137cd7a3e86ef5717c83df62cd6be4aa416eb0d4d7eb1ebc7ecba3bf92c4L517) as well so that these can be modified/set be the frontend

pkg/graph/schema.graphql
```diff
@@ -49,6 +49,7 @@ ModelPlan represent the data point for plans about a model. It is the central da
 type ModelPlan {
   id: UUID!
   modelName: String!
+  abbreviation: String
```
ClayBenson94, line 52: Same here - will want to add `abbreviation` to [ModelPlanChanges](https://github.com/CMSgov/mint-app/pull/590/files#diff-7b3b137cd7a3e86ef5717c83df62cd6be4aa416eb0d4d7eb1ebc7ecba3bf92c4L131) so that the frontend can properly update these fields.

ClayBenson94: EASI-2899 also calls out updating unit tests - adding these new fields into our existing tests would probably be all we need!

ClayBenson94 (changes requested):

pkg/graph/schema.graphql
OddTomBrooks: I don't see any tests regarding plan general characteristics, but I have updated the model plan test.

StevenWadeOddball (approved):
🚀 Thanks Tom, I tested and I was able to add the new fields and return them.

ClayBenson94 (approved):

PR #599, opened 2023-06-20
[NOREF] Handle JWT creation erorrs more gracefully in

ClayBenson94:
# NOREF

## Changes and Description

- Handle JWT creation errors more gracefully.

## How to test this change

- Testing this locally is tough - best bet for testing locally is to force returning an error in this function and see if the UI breaks in any way that's unhandled!

## PR Author Review Checklist

## PR Reviewer Guidelines


StevenWadeOddball (approved):
Thanks for making this @ClayBenson94! The app continues to work as expected with this early return and it makes sense.

I do want to call out that the front end continues to hit this endpoint over and over again until it is successful. It is probably worth looking into handling this on front end as well, but this is a good first step.

PR #606, opened 2023-06-23
Noref/change history table name to type spike

StevenWadeOddball:
## Changes and Description

- Add a helper method to return GQL Table name for Change History 
   - Instead of camel casing the result, we utilized a map as it's possible it doesn't follow the same paradigm
- Add a helper method to return the camelCased Field name of a field
   - There are too many fields to do a direct mapping here, without redoing the mapping work that the front end is already doing. The Camel Case helper is used a close approximation

## How to test this change
1. Bring up the application.
2. Seed data `scripts/dev db:seed`
3. Run this query after a few minutes ( to let open search build the results)

```gql
query searchChanges($filters: [SearchFilter!], $sortBy: ChangeHistorySortParams, $page: PageParams) {
  searchChanges(filters: $filters, sortBy: $sortBy, page: $page) {
    guid
    modelPlanID
    tableID
    tableName
    primaryKey
    gqlTableName
    foreignKey
    fields {
      changes {
        name
        nameCamelCase
        value {
          ... on FieldValue {
            new
            old
          }
        }
      }
    }
    modifiedDts
    modifiedBy {
      id
      username
      isEUAID
      commonName
      locale
      email
      givenName
      familyName
      zoneInfo
      hasLoggedIn
    }
  }
}

```
Variables
```gql
{
  "filters": [
    {
        "type": "CHANGED_AFTER",
        "value": "2023-01-01T00:00:00.000Z"
    },
    {
        "type": "CHANGED_BEFORE",
        "value": "2024-01-01T00:00:00.000Z"
    },
    {
        "type": "FREE_TEXT",
        "value": "high level note"
    }
  ],
  "sortBy": {
    "field": "CHANGE_DATE",
    "order": "DESC"   
  },
  "page": {
    "limit": 200,
    "offset": 0
  }
}
```
4. Ensure the gqlTableName and Field nameCamelCase are cased correctly. It should look something like this.
```JSON
{
                "guid": "2_b53346e0-120a-4746-a9e1-8710c517426c",
                "modelPlanID": "c09e80d9-cfb1-4e32-ad2c-07838262cb57",
                "tableID": 2,
                "tableName": "plan_basics",
                "primaryKey": "b53346e0-120a-4746-a9e1-8710c517426c",
                "gqlTableName": "planBasics",
                "foreignKey": "c09e80d9-cfb1-4e32-ad2c-07838262cb57",
                "fields": {
                    "changes": [
                        {
                            "name": "phased_in",
                            "nameCamelCase": "phasedIn",
                            "value": {
                                "new": "t",
                                "old": null
                            }
                        }
 }
 
```

## PR Author Review Checklist

## PR Reviewer Guidelines


pkg/models/change_table_record.go
```diff
@@ -38,3 +96,14 @@ type FieldValue struct {
 	New interface{} `json:"new"`
 	Old interface{} `json:"old"`
 }
+
+// NameCamelCase converts the name database string value, to lower case camel. This is meant as an approximation of the GQL field name
+func (field *Field) NameCamelCase() string { //can include ctx context.Context if desired
```
OddTomBrooks, line 101: Does it make sense to include ctx here? If there is no present need we should remove the comment.

pkg/models/change_table_record.go
```diff
@@ -22,6 +23,63 @@ type ChangeTableRecord struct {
 	ModifiedBy  *authentication.UserAccount `json:"modified_by"`
 }
 
+// GQLTableName casts a table name into a strict type for GQL
+func (ctr *ChangeTableRecord) GQLTableName() GQLTableName { //can include ctx context.Context if desired
+
+	switch ctr.TableName { // TODO: this could be a map if desired, but GO doesn't allow constant maps
```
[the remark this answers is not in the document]
StevenWadeOddball: Thanks Tom, no I don't. I removed the comment for now. I think this works for this iteration.

OddTomBrooks (approved):

pkg/models/change_table_record.go
StevenWadeOddball: Thanks @OddTomBrooks, I don't think we need it here, but I wanted to note that it is an allowed signature if we need it. There is no need to log anything in this method or use anything from context, but I added the note since this largely started as a discovery task.

PR #609, opened 2023-06-27
[EASI-2713] Email Notification:  Model Plan Dates Changed

OddTomBrooks:
# EASI-2713

## Changes and Description

- Add email notifications when model plan dates change

## How to test this change

- Run `scripts/dev down && scripts/dev up && scripts/dev db:seed`
- Assert that the following database table is created: `config.model_plan_date_changed_email_recipients`
- Assert that the database table contains the following test recipients: `one@test.gov, two@test.gov`
- Open `Mailcatcher` at `http://localhost:1085/`
- Navigate to the `Model basics` task list section in the model plan `Plan with Basics`
- Assign any `Model Category` selection
- Click `Next`
- Assert that no email is generated
  - Faulty Date Change processing would generate a data as `Plan Basics` is updated here
- Click `Next`
- Permutation Testing
  - Assign various different dates for testing purposes
    - There are two types of dates - individual dates and date ranges. Ensure they work both independently and together.
  - Assert that you receive an email in `Mailcatcher` with changes according to your test and that the `To` column contains all recipients in your `config.model_plan_date_changed_email_recipients` table (`one@test.gov, two@test.gov` by default)

## PR Author Review Checklist

## PR Reviewer Guidelines


OddTomBrooks:
In order to avoid merge issues, this PR contains changes from the feature branch `EASI-3000/read_view_users_contribute_discussions`. Please ignore changes related to user roles as those changes will disappear once the feature branch is merged to main.

cmd/dbseed/main.go
```diff
@@ -242,4 +254,7 @@ func (s *Seeder) SeedData() {
 			},
 		},
 	)
+
+	s.createEmailRecipientModelDatesChanged("one@test.gov")
+	s.createEmailRecipientModelDatesChanged("two@test.gov")
```
ClayBenson94, line 259: Are these wrapers (and associated store methods) still needed now that we're configuring recipients by env vars?

pkg/graph/resolvers/plan_basics.go
```diff
@@ -12,14 +16,66 @@ import (
...
+	fieldNames := map[string]string{
+		"completeICIP":            "Complete ICIP",
+		"clearanceStarts":         "Clearance",
+		"clearanceEnds":           "Clearance",
+		"announced":               "Announce model",
+		"applicationsStart":       "Application period",
+		"applicationsEnd":         "Application period",
+		"performancePeriodStarts": "Performance period",
+		"performancePeriodEnds":   "Performance period",
+		"wrapUpEnds":              "Model wrap-up end date",
```
ClayBenson94, line 29: This feels similar / duplicative to what's in getDateFieldsToCheck. Would it be possible to use this map for both purposes? ([You can iterate over the keys of a map](https://goplay.tools/snippet/TlQMdVWPHA6), which might mean this function could be used to get the list of field names to check as well)

pkg/storage/SQL/model_plan_date_changed_email_recipients/create.sql
ClayBenson94, line 1: Same comment from above - we can probably get rid of this file, right?

pkg/storage/SQL/model_plan_date_changed_email_recipients/get.sql
ClayBenson94, line 1: Same comment from above - we can probably get rid of this file, right?

pkg/storage/model_plan_date_changed_email_recipientsStore.go
ClayBenson94, line 1: Same comment from above - we can probably get rid of this file, right?

ClayBenson94 (changes requested):

StevenWadeOddball (changes requested):
Thanks @OddTomBrooks , I had some weird behavior when I was testing this locally. It doesn't always recognize that a date is different than it was previously.  For example, if I have a complete ICIP date set, and I change it again to another date in the future, no email fires. I also confirmed with breakpoints in the code that the no date changes register. 

The email that did fire off looks like it might be getting the wrong data (eg providing the same date twice?)

image


Could you please take a look at the date comparison logic?  I think that would be a good place to add Unit tests to verify that this functionality is behaving as expected. 

Also, to verify, are all the store functions referencing an earlier approach which stored users in the database?

pkg/graph/resolvers/plan_basics_helper.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// DateProcessor is a struct that processes date changes
+type DateProcessor struct {
+	changes  map[string]interface{}
+	existing map[string]interface{}
+}
+
+// NewDateProcessor is a constructor for DateProcessor
+func NewDateProcessor(changes map[string]interface{}, existing *models.PlanBasics) (*DateProcessor, error) {
```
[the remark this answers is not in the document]
StevenWadeOddball: For what it's worth, you could use some table driven testing to validate the permutations that you mentioned above for manual testing. I think that be worthwhile.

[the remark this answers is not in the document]
OddTomBrooks: This is a great point. As this logic has been abstracted from the resolver it is unit testable.

OddTomBrooks:
> Thanks @OddTomBrooks , I had some weird behavior when I was testing this locally. It doesn't always recognize that a date is different than it was previously. For example, if I have a complete ICIP date set, and I change it again to another date in the future, no email fires. I also confirmed with breakpoints in the code that the no date changes register.
> 
> The email that did fire off looks like it might be getting the wrong data (eg providing the same date twice?)
> 
> image
> 
> Could you please take a look at the date comparison logic? I think that would be a good place to add Unit tests to verify that this functionality is behaving as expected.
> 
> Also, to verify, are all the store functions referencing an earlier approach which stored users in the database?

Strange, I have not seen such behaviour during local testing. Thank you for identifying this problem, I will go through this thoroughly to determine the root cause. Perhaps it will be exposed during construction of unit tests.

ClayBenson94 (changes requested):
image

Tested locally - looks like, at least for the ICIP date, it's showing the new date in both the old and new fields. Not sure if this is in the tested functions or not, or if it's cropping up somewhere else (like in the templates?)

OddTomBrooks:
> image
> 
> Tested locally - looks like, at least for the ICIP date, it's showing the new date in both the old and new fields. Not sure if this is in the tested functions or not, or if it's cropping up somewhere else (like in the templates?)


@ClayBenson94 
I'm struggling to reproduce this issue.

image
image
image
image


Can you repro it? If so, can you provide the repro steps?
It might be based on initial values. In that case, which model plan are you modifying?

Also, as a sanity check, do you have latest and did you rebuild your containers?

ClayBenson94 (approved):

PR #614, opened 2023-06-29
[EASI-3015] Update Salesforce Names

OddTomBrooks:
# EASI-3015

## Changes and Description

- Update Salesforce possible operational solution names

## How to test this change

- Run `scripts/dev down && scripts/dev up && scripts/dev db:seed`
- Open the possible operational solutions database table
- Assert that the Salesforce entry names are to spec

## PR Author Review Checklist

## PR Reviewer Guidelines


migrations/V96__Update_Salesforce_Connect_And_Portal_Names.sql
StevenWadeOddball, line 1: the names look good according to FIGMA! Have you considered using the solution key instead of the id for these update statements? I think it might be clearer when we look at this in the future what solution this was for instead of the hard coded id (though this will work as is)

StevenWadeOddball:
@OddTomBrooks , is this ticket also handling the front end changes for this name? Or just the backend?

StevenWadeOddball (dismissed):
Thanks @OddTomBrooks ! BE changes look good, this should be set to merge when FE changes get made.

StevenWadeOddball (approved):
Thanks @OddTomBrooks , BE changes look good! I focused on the BE migration since the FE was addressed in a separate PR.

PR #625, opened 2023-07-07
[NOREF] - Fixed conditional when updating custom solution

patrickseguraoddball:
# NOREF

## Changes and Description

`UpdateOperationalSolution` payload had an incorrect conditional for passing `nameOther` .  It was expecting `selectedSolution` to be falsy, when that isn't the case here

- Fixed the conditonal to explicity check for `OTHER_NEW_PROCESS` to pass `nameOther` to the mutation payload

Had to fix areas where we were treating custom solutions with a key.  We now pass null keys when creating a brand new custom solution.  For the case where we have the `OTHER_NEW_PROCESS` solution as a default, we pass that key and it behaves like `treatAsOther`

Also added a translation key for render `OTHER_NEW_PROCESS` as 'Other new process' on the card once created.

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/src/i18n/en-US/draftModelPlan/itSolutions.ts
+++ b/src/i18n/en-US/draftModelPlan/itSolutions.ts
@@ -236,6 +236,7 @@ const itSolutions = {
     existingCmsDataAndProcess: 'Existing CMS data and process',
     interalStaff: 'Internal staff',
     otherNewProcess: 'Other'
-  }
+  },
+  otherNewProcess: 'Other new process'
 };
 export default itSolutions;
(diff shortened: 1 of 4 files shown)


OddTomBrooks (dismissed):

StevenWadeOddball (approved):
Thanks @patrickseguraoddball , approving based on functional testing.

I pulled this down locally, and created both a custom solution and `other new process`. I confirmed that other new process is using the treatAsOther functionality, and the custom solution is passing a null need type as expected. 

I also updated both solutions and confirmed they update correctly as expected.

patrickseguraoddball:
> Thanks @patrickseguraoddball , approving based on functional testing.
> 
> I pulled this down locally, and created both a custom solution and `other new process`. I confirmed that other new process is using the treatAsOther functionality, and the custom solution is passing a null need type as expected.
> 
> I also updated both solutions and confirmed they update correctly as expected.

Thanks so much Steven for the help and quick review

PR #626, opened 2023-07-07
[EASI-3031] Filter Banner and Summary Updates

garyjzhao:
# EASI-3031

## Changes and Description

- Replacing the filter box with the filter banner that currently displays on the filtered content. 
- Update "Back to all models" link to "View all models"
- Updates to "Show model summary" and "Hide model summary" link icon and placement
- The overall status should now be incorporated into the summary below the name of the model
- A new "Edit Model Plan" link should show near the last updated information. 
    - Because of this change, we'll also need to update the side nav of the task list. Add a new "Actions" heading and a "Switch to the Read View for this Model Plan" link that takes the user to the read view. 

image

## How to test this change

1. Navigate to a task list
2. Check side navigation has "Actions" and new link to "Switch to Read View for this Model Plan"
3. Ensure new header looks like the figma

## PR Author Review Checklist

## PR Reviewer Guidelines


src/components/shared/CollapsableLink/index.tsx
```diff
@@ -68,8 +68,8 @@ const CollapsableLink = ({
       unstyled
       data-testid="collapsable-link"
     >
-      {iconPosition === 'left' ? selectedLabel : expandIcon}
-      {iconPosition === 'left' ? expandIcon : selectedLabel}
+      {iconPosition === 'right' ? selectedLabel : expandIcon}
+      {iconPosition === 'right' ? expandIcon : selectedLabel}
```
patrickseguraoddball, line 72: This change affected other instances of `<CollapsableLink /> other than readonly.  Need to make sure if you change a shared component directly like this, that it doesn't affect other components.  

`<NDABanner />` uses this component and not the icon is in the wrong position

Screenshot 2023-07-10 at 11 08 53 AM

patrickseguraoddball (changes requested):
Just a few small changes

The padding of `<SummaryBox />` on mobile seems a bit too large compared to Figma

Screenshot 2023-07-10 at 11 10 09 AM

Also seeing that the `modifiedDts` is not being populated and rendered when a model plan is updated.  I think this only triggers if the status is updated.  May want to check with Natasha/BE that this is intended functionality.  Basically should the last updated be based purely on the status change, or if changes were made anywhere on the task list
Screenshot 2023-07-10 at 11 14 30 AM

patrickseguraoddball (changes requested):
The responsiveness on the summarybox is still off at certain widths.  The header/status info should span all the way across grid-12 for each element

Screenshot 2023-07-12 at 9 11 58 AM

patrickseguraoddball:
Also on mobile there seems to be no space between the yelllow info box and the above banner.   This figma shows there should be space in the same way as desktop

Screenshot 2023-07-12 at 9 12 58 AM

patrickseguraoddball (approved):
LG, thanks!

PR #632, opened 2023-07-11
[EASI-3072] Payment Question Updates

OddTomBrooks:
# EASI-3072

## Changes and Description

- Added 'SEMIANNUALLY' enum
- Updated references from 'BIANNUALLY' to 'SEMIANNUALLY'
- Added new trust fund type
- Added trust fund columns to plan payments
- Changed enum casing for gql codegen
- Added FE checbox component fortrust fund types on Payments
- updated readonly render for Payments
- Updated unit tests

There is no readonly mockup for this, but I confirmed with UX that what is implemented is correct

Screenshot 2023-08-10 at 2 03 30 PM
Screenshot 2023-08-10 at 2 03 53 PM

## PR Author Review Checklist

## PR Reviewer Guidelines


migrations/V102__Add_Enum_Semiannually.sql
ClayBenson94, line 1: I think this and V103 could be combined into a single migration that looks more like [this one from EASi](https://github.com/CMSgov/easi-app/blob/0e390341fd5af9e74571d7873c1eda70aca02698/migrations/V142__Remove_PersonRole__Unknown.sql#L1-L18), that way we can do an "in place" substitution of one enum value for the other, i.e.:

1. Add the new enum value(s)
2. Rename the existing enums to `{enum}_old`
3. Create a new enum without BIANNUALLY
4. Alter tables to use the new enum type
5. Drop the old enum

migrations/V105__Update_Payment_Question_Remove_Funding_Structure_Add_Trust_Fund_Types.sql
```diff
@@ -0,0 +1,11 @@
...
+DROP COLUMN IF EXISTS funding_structure;
+
+CREATE TYPE trust_fund_type AS ENUM (
+  'MEDICARE_PART_A_HI_TRUST_FUND'
+  'MEDICARE_PART_B_SMI_TRUST_FUND'
+);
+
+ALTER TABLE public.plan_payments
+  ADD COLUMN funding_source_trust_fund_type trust_fund_type DEFAULT NULL,
+  ADD COLUMN funding_source_r_trust_fund_type trust_fund_type DEFAULT NULL;
```
ClayBenson94, line 11: Do these need to be arrays? It looks like we should be able to select HI and/or SMI, not just one or the other.

migrations/V103__Update_Enums_Biannually_To_Semiannually.sql
```diff
@@ -0,0 +1,16 @@
+-- Update the columns that use the old enum values to use the new ones
+UPDATE plan_beneficiaries
+SET beneficiary_selection_frequency = 'SEMIANNUALLY'
+WHERE beneficiary_selection_frequency = 'BIANNUALLY';
+UPDATE plan_participants_and_providers
+SET provider_addition_frequency = 'SEMIANNUALLY'
+WHERE provider_addition_frequency = 'BIANNUALLY';
+UPDATE plan_ops_eval_and_learning
+SET data_sharing_frequency = ARRAY ['SEMIANNUALLY']::data_frequency_type[]
+WHERE 'BIANNUALLY' = ANY (data_sharing_frequency);
```
ClayBenson94, line 10: Will this only update the entries from `BIANUALLY` to `SEMIANUALLY`, or will it set the entire array to one with a single entry of `SEMIANUALLY`?

It looks like it'd do the latter, but I haven't tested that - just looking at `SET data_sharing_frequency = ARRAY ['SEMIANNUALLY']::data_frequency_type[]` would make me think we'd always set an array with one item in it

ClayBenson94: I think we'd discussed holding off on the semiannual / biannual changes for now - if this proves to be a big lift, we can always just leave the enum on the backend and have FE translate it - this doesn't need to be changed on the backend if it involves a big lift

ClayBenson94 (changes requested):

ClayBenson94:
Removing the `blocked` label here - I think there's changes we can make to this before FE completes their work

migrations/V103__Update_Enums_Biannually_To_Semiannually.sql
OddTomBrooks: You're correct, this will only set a fixed array and must be updated. Apologies for the oversight. I thought the discussion was that we would not replace the enum type but would update the enum values.

I have code for updating the enum array and am testing it.

pkg/graph/resolvers/plan_payments_test.go
```diff
@@ -152,7 +151,6 @@ func (suite *ResolverSuite) TestPlanPaymentsReadByModelPlan() {
 	suite.Nil(pp.SharedSystemsInvolvedAdditionalClaimPaymentNote)
 	suite.Nil(pp.PlanningToUseInnovationPaymentContractor)
 	suite.Nil(pp.PlanningToUseInnovationPaymentContractorNote)
-	suite.Nil(pp.FundingStructure)
```
ClayBenson94, line 155: Can you add in a `Nil` check for the new fields added in the model?

pkg/storage/SQL/plan_payments/create.sql
ClayBenson94, line 1: This file is missing the new fields.

pkg/storage/SQL/plan_payments/get_by_id.sql
ClayBenson94, line 1: This file is missing the new fields.

pkg/storage/SQL/plan_payments/get_by_model_plan_id_LOADER.sql
ClayBenson94, line 1: This file is missing the new fields.

ClayBenson94 (dismissed):
There's a few spots that have enums that don't match and a few SQL statements missing the new field names. Do these changes work locally for you?

ClayBenson94 (approved):

PR #636, opened 2023-07-12
[NOREF] - Filter view view route restructure

patrickseguraoddball:
# NOREF

## Changes and Description

- Fixed app crash when passing an unmapped filter query param
- Reworked readonly routing to all for filter views to be passed without a page subkey

This is in preparation for the BE to begin supplying filter view links to readonly without needing to specify a subpage

Ex: `${host}/models/${modelID}/read-only`
Ex: `${host}/models/${modelID}/read-only?filter-view=ccw`

Previous the routing didn't allow for filter views without a specified subpage (`model-basics`, etc)

## How to test this change

Test out navigation to `${host}/models/${modelID}/read-only` without a subpage , should redirect to `model-basics`
Append a filter-view query param to `${host}/models/${modelID}/read-only` and verify it renders - `${host}/models/${modelID}/read-only?filter-view=ccw`
Try to type in an incorrect filter-view param - should reroute back to model basics

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/src/views/App/index.tsx
+++ b/src/views/App/index.tsx
@@ -104,14 +104,8 @@ const AppRoutes = () => {
       {/* Model Routes */}
       <SecureRoute path="/models" exact component={ModelPlan} />
 
-      <Redirect
-        exact
-        from="/models/:modelID/read-only"
-        to="/models/:modelID/read-only/model-basics"
-      />
-
       <SecureRoute
-        path="/models/:modelID/read-only/:subinfo"
+        path="/models/:modelID/read-only/:subinfo?"
         exact
         component={ReadOnly}
       />
--- a/src/views/ModelPlan/ReadOnly/_components/FilterView/Banner/index.tsx
+++ b/src/views/ModelPlan/ReadOnly/_components/FilterView/Banner/index.tsx
@@ -25,8 +25,6 @@ const FilterViewBanner = ({
 
   const history = useHistory();
 
-  const { pathname } = history.location;
-
   return (
     <div
       className="position-sticky z-100 top-0 bg-primary-darker text-white padding-105"
@@ -58,11 +56,7 @@ const FilterViewBanner = ({
                 type="button"
                 unstyled
                 className="text-white text-no-wrap"
-                onClick={() =>
-                  history.push(
-                    `${pathname}${pathname.length === 4 ? '/model-basics' : ''}`
-                  )
-                }
+                onClick={() => history.push(history.location.pathname)}
               >
                 {t('clearFilter')}
               </Button>
--- a/src/views/ModelPlan/ReadOnly/_components/FilterView/Banner/index.tsx
+++ b/src/views/ModelPlan/ReadOnly/_components/FilterView/Banner/index.tsx
@@ -22,7 +22,11 @@ const FilterViewBanner = ({
   openFilterModal
 }: FilterViewBannerProps) => {
   const { t } = useTranslation('filterView');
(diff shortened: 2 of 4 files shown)


src/views/ModelPlan/ReadOnly/_components/FilterView/Banner/index.tsx
```diff
@@ -54,7 +56,7 @@ const FilterViewBanner = ({
                 type="button"
                 unstyled
                 className="text-white text-no-wrap"
-                onClick={() => history.push(`${history.location.pathname}`)}
+                onClick={() => history.push(history.location.pathname)}
```
ClayBenson94, line 59: Was this a functional change, or just stylistic? Only case I could see this mattering is pushing `undefined` or `null` vs `"undefined"` or `"null"`

src/views/App/index.tsx
```diff
@@ -104,14 +104,8 @@ const AppRoutes = () => {
...
 
-      <Redirect
-        exact
-        from="/models/:modelID/read-only"
-        to="/models/:modelID/read-only/model-basics"
-      />
-
       <SecureRoute
-        path="/models/:modelID/read-only/:subinfo"
+        path="/models/:modelID/read-only/:subinfo?"
```
ClayBenson94, line 108: What does the addition of `?` do to this route?

patrickseguraoddball: The `?` makes the parameter optional.  The component can still be rendered without passing any of the subinfo (`model-basics`, `payments`, etc)

src/views/ModelPlan/ReadOnly/index.tsx
```diff
@@ -260,10 +267,14 @@ const ReadOnly = ({ isHelpArticle }: { isHelpArticle?: boolean }) => {
   const subComponent = subComponents[subinfo];
 
   if ((!loading && error) || (!loading && !data?.modelPlan)) {
-    return <NotFoundPartial />;
+    return <NotFound />;
+  }
+
+  if (!subinfo && !isViewingFilteredGroup) {
+    history.replace(`${location.pathname}/model-basics`);
```
ClayBenson94, line 274: Nit: Might be nice to pull `model-basics` into a `const` at the top of this component under something like

```typescript
const defaultSection = "model-basics"
```

just for visibility.

ClayBenson94 (changes requested):
Requesting changes for the default section comments

src/views/ModelPlan/ReadOnly/_components/FilterView/Banner/index.tsx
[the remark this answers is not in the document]
patrickseguraoddball: Reply

src/views/App/index.tsx
patrickseguraoddball: test

ClayBenson94 (changes requested):
Not sure how important this really is, but going to `http://localhost:3005/models/92b151fc-3cdc-4999-ae4d-abfe8503603b/read-only/` (note the trailing slash) ends up booting me to `http://localhost:3005/models/92b151fc-3cdc-4999-ae4d-abfe8503603b/read-only//model-basics`, which gives me a "Page not found"

Not sure there's an easy way to fix this, but could be a headache for folks that chop off the section in the URL themselves 🤷

ClayBenson94 (approved):

PR #655, opened 2023-07-28
[NOREF] Use new graphql-codegen tool

ClayBenson94:
# EASI-000

## Changes and Description

- Introduce new `queriesCodegen` folder, where all queries will reside for the new codegen tool

The high level approach used here is to introduce `graphql-codegen`, which is more supported than the apollo codegen tool we're using (see [this issue](https://github.com/apollographql/apollo-tooling/issues/2053)). The goal is to introduce a way to slowly cut over queries to this new codegen tool without having to completely and immediately deprecate apollo codegen.

Looking for feedback on the following:
- Organization of files (one file per query? Multiple? Both are supported!)
- Naming (Not sure if LOVE `queriesCodegen` - could be convinced to name this something else
- Any additional configs/settings that the new codegen tool might provide that we should enable here!

## How to test this change

- Run `scripts/dev gql` to ensure no files change
- Try adding a new query to `src/queriesCodegen` and using it within the app!

## PR Author Review Checklist

## PR Reviewer Guidelines


package.json
```diff
@@ -170,7 +173,15 @@
...
-    "node-fetch": "2.6.1"
+    "node-fetch": "2.6.1",
+    "graphql": "15.8.0",
+    "@graphql-typed-document-node/core": "3.2.0"
+  },
+  "comments": {
+    "on_resolutions": {
+      "graphql": "Having graphql fixed at 15.8.0 allows apollo codegen and graphql-codegen to work together. Once the former is deprecated, you can remove this",
+      "typed_document_node": "https://github.com/dotansimha/graphql-code-generator/issues/9293"
+    }
```
ClayBenson94, line 184: Trying something here - I wanted to document a few things that I did in the package.json, but since it's just JSON, it doesn't support actual comments, and some places online recommended just throwing a custom object in there for stuff like this!

patrickseguraoddball:
Thanks for this, can't wait to use it!

I like the idea of one file for all the generated types.  

Also not sure if I also love the name `queriesCodegen`.  I think `queries` isn't encapsulating (mutations, subscriptions).  Maybe something like `apolloGQL`.  Eventually we will deprecate gql-gen and naming it Codegen won't really matter.  Idk just thoughts.

Are `fragment-asking.ts`  generated or externally sourced code?  If so can we get a comment here or in the code of the source?

Works great!

patrickseguraoddball (approved):
Works great, and the structure makes sense

PR #658, opened 2023-08-02
[EASI-3100] Update Model Categories

OddTomBrooks:
# EASI-3100

## Changes and Description

- Update model categories to match CMS spec
- Add additional categories

## PR Author Review Checklist

## PR Reviewer Guidelines


pkg/graph/schema.graphql
```diff
@@ -1744,15 +1746,30 @@ enum ModelType
...
+  INIT_SPEED_ADOPT_BEST_PRACTICE
+  PRIMARY_CARE_TRANSFORMATION
+  UNKNOWN
+  DISEASE_SPECIFIC_AND_EPISODIC
+  HEALTH_PLAN
+  PRESCRIPTION_DRUG
+  STATE_BASED
+  STATUTORY
+  TO_BE_DETERMINED
+}
```
ClayBenson94, line 1764: I think this enum has too many items in it - from what I can see, we only support 7 options (per the Figma), which look to be the ones that are currently in `AdditionalModelCategory` (except for `TO_BE_DETERMINED`, which is missing from that one)

ClayBenson94 (changes requested):

migrations/V102__Update_Model_Categories.sql
ClayBenson94, line 1: We'll probably want this migration to handle any old values that no longer exist getting properly set to `TO_BE_DETERMINED`. (i.e. `Initiatives to Accelerate the Development and Testing of New Payment and Service Delivery Mode` no longer exists, adn should be set to `TO_BE_DETERMINED`, that way we can update it in prod later)

ClayBenson94 (changes requested):

pkg/graph/schema.graphql
OddTomBrooks: Do you think it's appropriate to mask out the TO_BE_DETERMINED option from the FE side for AdditionalModelCategory selections?

logstash/V102__Update_Model_Categories.sql
ClayBenson94, line 1: Move to `migrations` (looks like a change from local testing)

pkg/graph/schema.resolvers.go
```diff
@@ -486,6 +486,11 @@ func (r *operationalSolutionResolver) OperationalSolutionSubtasks(ctx context.Co
 	return resolvers.OperationalSolutionSubtaskGetBySolutionIDLOADER(ctx, obj.ID)
 }
 
+// AdditionalModelCategories is the resolver for the additionalModelCategories field.
+func (r *planBasicsResolver) AdditionalModelCategories(ctx context.Context, obj *models.PlanBasics) ([]models.ModelCategory, error) {
+	panic(fmt.Errorf("not implemented: AdditionalModelCategories - additionalModelCategories"))
```
ClayBenson94, line 491: Call resolver function here

pkg/models/model_plan.go
```diff
@@ -32,15 +32,34 @@ type ModelCategory string
...
+type AdditionalModelCategory string
+
+// These constants represent the different values of AdditionalModelCategory
+const (
+	AMCAccountableCare            AdditionalModelCategory = "ACCOUNTABLE_CARE"
+	AMCDiseaseSpecificAndEpisodic AdditionalModelCategory = "DISEASE_SPECIFIC_AND_EPISODIC"
+	AMCHealthPlan                 AdditionalModelCategory = "HEALTH_PLAN"
+	AMCPrescriptionDrug           AdditionalModelCategory = "PRESCRIPTION_DRUG"
+	AMCStateBased                 AdditionalModelCategory = "STATE_BASED"
+	AMCStatutory                  AdditionalModelCategory = "STATUTORY"
```
ClayBenson94, line 62: Don't need `AdditionalModelCategory` - can move these types to `ModelCategory` and ensure they align with SQL / GQL enums

ClayBenson94 (dismissed):

ClayBenson94 (commented):
@OddTomBrooks Changes look good! Going to hold off approving until FE changes are also in, but this seems like it should be all set for FE!

garyjzhao:
@OddTomBrooks @ClayBenson94 Check Migrations was failing because it detected a duplicate file. `V102...`. I renamed the file in [commit](https://github.com/CMSgov/mint-app/pull/658/commits/83e8d9a19797868086ff660edeaf3acfafa85e82). 

Is that right? And if is there anythign else to do to address that failing test?

patrickseguraoddball (changes requested):
Really nice work Gary, works great.

Aside from the small changes, I notice that clicking `Back` on basics `Overview` returns to task list instead of previous page.  I may have introduced this?  If you could add this small change that would be much appreciated as well.

patrickseguraoddball:
Noticed that the model categories on the home page are not rendering

Screenshot 2023-08-17 at 1 01 49 PM

src/views/ModelPlan/ReadOnly/Table/index.tsx
```diff
@@ -116,11 +115,27 @@ const Table = ({
...
-        Cell: ({ value }: any) => {
+        Cell: ({ row, value }: any) => {
+          const additionalModelCategory =
+            row.original.basics.additionalModelCategories;
+
           if (!value) {
             return <div>{h('noAnswer.tBD')}</div>;
           }
-          return translateModelCategory(value);
+          if (value) {
```
patrickseguraoddball, line 125: This value check isn't necessary as it's already being checked in the line above

src/views/ModelPlan/ReadOnly/Table/index.tsx
```diff
@@ -53,6 +51,7 @@ const Table = ({
   const { t } = useTranslation('readOnlyModelPlan');
   const { t: h } = useTranslation('modelSummary');
   const { t: f } = useTranslation('home');
+  const { t: planBasicsT } = useTranslation('planBasics');
```
patrickseguraoddball, line 54: I think this is the incorrect translation file, it should be

const { t: basicsT } = useTranslation('basics');

ClayBenson94 (approved):
Except for Patricks feedback these changes look good. Did a dump of data with a model plan with each category from `main`, ran the migration, updated to this branch, and everything looks as expected!

src/views/ModelPlan/TaskList/Basics/index.test.tsx
```diff
@@ -71,6 +76,7 @@ describe('Model Plan Documents page', () => {
       ).toBeInTheDocument();
     });
   });
+
```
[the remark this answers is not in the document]
ClayBenson94: Added in 0545f1f0fa2d7a5029107963573b17f4c39576b0

src/views/ModelPlan/ReadOnly/ModelBasics/index.tsx
```diff
@@ -207,9 +211,54 @@ const ReadOnlyModelBasics = ({
...
+                  <Tooltip
+                    label={basicsT(
+                      `modelCategory.optionsLabels.${modelCategory}`
+                    )}
+                    position="right"
+                  >
+                    <IconInfo className="text-base-light" />
+                  </Tooltip>
+                )}
+              </span>
```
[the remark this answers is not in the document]
ClayBenson94: Added in 0545f1f0fa2d7a5029107963573b17f4c39576b0

src/views/ModelPlan/ReadOnly/Table/index.tsx
ClayBenson94: Addressed in 6256b5d3c93dea3c07cc1ce0a457ef279bbc3f49

ClayBenson94 (approved):

PR #659, opened 2023-08-03
[NOREF] - Changed post login redirect from react-router to native browser redirect

patrickseguraoddball:
# NOREF

## Changes and Description

MINT deployed and local env were experiencing issues with gql subscription authentication due to a missing okta storage token post login.  The app was using `react-router-dom` for post redirect, not allowing the app to ingest the auth token into its ws connections

This was carried over from EASI, so the same issue may exist there as well.

This resolves the issue, but we may want to dig further to see if it's possible to still use `react-router-dom` for a cleaner user experience

- Changed the `AuthenticationWrapper` redirect to use native browser redirect/hard refresh - post login

# UPDATE

- Reverted the  native browser redirect/hard refresh - post login
- Added a lazy config to ws connection to only attempt to init on first subscription
- Wrapped `authToken` in `connectionParams` function

This re-inits successfully the okta token in the ws connection without the need for a browser refresh

Reviewers may want to test both methods (via commit history).  I still see the initial ws connection fail with #2 but it successfully retries and connects on second attempt

## How to test this change

- Log in with okta
- Navigate to a task list section form
- Copy url to clipboard
- Log out
- Paste url
- Sign in again
- Verify ws connection that auth token was passed

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -88,9 +88,10 @@ const wsProtocol = protocol === 'https' ? 'wss' : 'ws'; // Use WSS when connecti
 const wsLink = new WebSocketLink(
   new SubscriptionClient(`${wsProtocol}://${gqlAddressWithoutProtocol}`, {
     reconnect: true,
-    connectionParams: {
+    lazy: true,
+    connectionParams: () => ({
       authToken: getAuthHeader(process.env.REACT_APP_GRAPHQL_ADDRESS as string)
-    }
+    })
   })
 );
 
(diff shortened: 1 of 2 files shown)


Jdwoodson:
👏🏻👏🏻👏🏻

patrickseguraoddball:
I think* we can apply a middleware to reconnect websocket connection post login with new okta token

PR #665, opened 2023-08-16
[EASI-3098] Share Model Plan

OddTomBrooks:
# EASI-3098

## Changes and Description

- Share Model email template
- View Filter enum

FE Changes
- Added UI for `Share` on modal
- Added clipboard copy for entire model plan as well as filtered view link
- Mutation to send model plan
- Success/error states on tasklist/readonly
- Updated unit tests

Screenshot 2023-08-29 at 10 49 59 AM

Screenshot 2023-08-29 at 10 50 27 AM

Screenshot 2023-08-29 at 10 50 59 AM

Screenshot 2023-08-29 at 11 03 40 AM

## How to test this change

*From both task list and readonly

- Verify if opening modal from filtered view, that is gets autopopulated
- Verify button is disabled if not filter view is selected
- Verify mutation passes successfully if valid email - success Alert component rendered
- Verify mutation fail on inproper email address - error Alert component rendered
- Verify copy link to clipboard is successful on both all plan and filter view selections

## PR Author Review Checklist

## PR Reviewer Guidelines


pkg/graph/resolvers/model_plan.go
```diff
@@ -279,3 +282,118 @@ func ModelPlanNameHistory(logger *zap.Logger, modelPlanID uuid.UUID, sortDir mod
...
+	receiverEmails []string,
+	optionalMessage *string,
+) (bool, error) {
+	modelPlan, err := store.ModelPlanGetByID(logger, modelPlanID)
+	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			logger.Warn("No model plan found for the given modelPlanID",
+				zap.String("modelPlanID", modelPlanID.String()))
+			return false, fmt.Errorf("no model plan found for the given modelPlanID: %w", err)
+		}
```
ClayBenson94, line 305: I don't think we need to have specific error handling for `ErrNoRows` here - we should probably just keep this simple, since we're returning an error either way, we'll be able to see what the source of the issue was without having to check it here

pkg/graph/resolvers/model_plan.go
```diff
@@ -279,3 +282,118 @@ func ModelPlanNameHistory(logger *zap.Logger, modelPlanID uuid.UUID, sortDir mod
...
+		UserName: principal.Account().CommonName,
+	})
+	if err != nil {
+		return false, fmt.Errorf("failed to execute email subject: %w", err)
+	}
+
+	var modelPlanCategories []models.ModelCategory
+	if planBasics.ModelCategory != nil {
+		modelPlanCategories = append(modelPlanCategories, *planBasics.ModelCategory)
+	}
```
ClayBenson94, line 344: With the recent merge of the Model Category updates I think we should add the additional categories here as well (which is likely why it was configured to be a slice of Categories here!)

ClayBenson94 (dismissed):
Great job on these changes! 🎉 

Just noting some changes we'll want to pull in regarding multiple model categories, and some smaller feedback around error handling that isn't as big of a priority

ClayBenson94:
We might also want to handle the case of an empty string as the message in the backend. Potentially a `zero.String` could be useful for that check!

image

pkg/graph/resolvers/model_plan.go
OddTomBrooks: The thinking here was to provide more expressive errors so that we could understand the context of the 'pq: no rows' error. Please let me know if you still want this removed and I will remove it.

pkg/graph/resolvers/model_plan.go
```diff
@@ -279,3 +282,118 @@ func ModelPlanNameHistory(logger *zap.Logger, modelPlanID uuid.UUID, sortDir mod
...
+		return false, fmt.Errorf("failed to fetch the model plan: %w", err)
+	}
+
+	planBasics, err := PlanBasicsGetByModelPlanIDLOADER(ctx, modelPlanID)
+	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			logger.Warn("No plan basics found for the given modelPlanID",
+				zap.String("modelPlanID", modelPlanID.String()))
+			return false, fmt.Errorf("no plan basics found for the given modelPlanID: %w", err)
+		}
```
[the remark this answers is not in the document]
OddTomBrooks: Same as above, given the reasoning please confirm if you would like this removed.

pkg/graph/resolvers/model_plan.go
ClayBenson94: I think we can do that, but if we want to wrap the errors I'd say doing it in the `store` method might make more sense rather than having to do it in each resolver that calls it, that way we can just forward the error up the call stack in the resolvers

OddTomBrooks: Good point.

OddTomBrooks:
> We might also want to handle the case of an empty string as the message in the backend. Potentially a `zero.String` could be useful for that check!
> 
> image

In this case I was able to resolve it with a simple template modification

garyjzhao (approved):
FE is working as described

PR #678, opened 2023-09-01
[EASI-3098] Use default sender when sharing model plans

ClayBenson94:
# EASI-3098

## Changes and Description

- Use default sender (configured per-env through env vars) instead of trying to send emails from the user who made the GQL call.

## How to test this change

- `scripts/dev up`
- `scripts/dev db:seed`
- Share a model plan through the UI
- Ensure that mailcatcher shows it's coming from `no-reply-mint-localhost@cms.hhs.gov`

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/pkg/graph/resolvers/model_plan.go
+++ b/pkg/graph/resolvers/model_plan.go
@@ -289,6 +289,7 @@ func ModelPlanShare(
 	principal authentication.Principal,
 	emailService oddmail.EmailService,
 	emailTemplateService email.TemplateService,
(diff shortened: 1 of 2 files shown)


StevenWadeOddball (approved):
Thanks Clay, I confirmed that this is sending from the correct sender now.

Out of scope here but, I noticed some failed sendings for invalid email addresses. I think it could be beneficial in a future ticket to validate the emails are valid on the front end before sending the mutation to the backend.

PR #681, opened 2023-09-05
[EASI-3008] - Added usernames to discussions,replies and model plan CSV

patrickseguraoddball:
# EASI-3008

## Changes and Description

- Added human readable user names where users are referenced in CSV model plan
  - Discussions, replies, and model plan
- Updated unit tests/snapshots
  -  Changed mocked data where time was set at midnight to be more consistent

CSV headers will be addressed in another ticket

## How to test this change

Download a model plan csv
Verify that user's names are rendered

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/src/utils/export/CsvData.ts
+++ b/src/utils/export/CsvData.ts
@@ -21,7 +21,7 @@ const csvFields = [
   },
   {
     label: 'Created By',
-    value: 'createdBy'
+    value: 'createdByUserAccount.commonName'
   },
   {
     label: 'Created At',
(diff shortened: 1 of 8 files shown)


src/utils/export/CsvData.ts
```diff
@@ -21,7 +21,7 @@ const csvFields = [
   },
   {
     label: 'Created By',
-    value: 'createdBy'
+    value: 'createdByUserAccount.commonName'
```
ClayBenson94, line 24: Should we also make this change in the queries & this file for `modifiedDts`?
Looks like we do have a [modifiedByUserAccount](https://github.com/CMSgov/mint-app/blob/78e5debd8e78a778ba4b03480d42d84d4ca79551/pkg/graph/schema.graphql#L58) in the schema, but what I'm not sure about is what happens if we use that and `modifiedByUserAccount` is `null` (which it can be if nothing's been modified - `scripts/dev db:seed` creates a few like this!)

I'm thinking we should, since this is what renders in the column as-is
image

ClayBenson94 (changes requested):
Changes look good, probably just want to update to handle this as well!

src/utils/export/CsvData.ts
patrickseguraoddball: Honestly thinking of nixing this branch/PR as the subsequent PR is kinda a rework of the csv fields and will be replaced anyway/merge conflicts

PR #707, opened 2023-09-20
[NOREF] Split docker compose files

ClayBenson94:
# NOREF

## Changes and Description

- This PR aims to clean up the docker compose files so that we can more easily spin up ONLY the services we need, and prevents having a bunch of compose files that re-define the same services in lots of different contexts
- Removes most (not all) duplicative environment variables that were in both .envrc and docker-compose files

## How to test this change

- Ensure the app works locally with `scripts/dev up`
- Ensure only the backend works with `scripts/dev up:backend`
- Ensure E2E tests pass in CI
- Ensure all server tests pass with `scripts/dev test:go`

## PR Author Review Checklist

## PR Reviewer Guidelines


StevenWadeOddball (approved):
Thanks @ClayBenson94 ! This looks good. I tested it locally and the functionality worked as expected.  

I wonder if we could expand this further in the future? It would be great not to need to spin up open search or faktory when we aren't developing on that tooling.

Jdwoodson:
> Thanks @ClayBenson94 ! This looks good. I tested it locally and the functionality worked as expected.
> 
> I wonder if we could expand this further in the future? It would be great not to need to spin up open search or faktory when we aren't developing on that tooling.

I agree and that is exactly the discussion Clay and I had during our pairing session on this, as well as my comment/thought during Retro. I definitely think further improvements can be made here!

StevenWadeOddball (approved):
Thanks @ClayBenson94! These additional changes look great as well. The front end seems to build really fast in the container after it's first build!

PR #710, opened 2023-09-20
[EASi 3154] links as documents be

StevenWadeOddball:
# EASI-3154

## Changes and Description

- Update Documents in GQL and DB to include new fields
    - isLink
    - URL 
- Validate the url with strict validation of format 
- Introduce a new mutation to create a linked Document
- Update relevant resolvers regarding getting download link for linked documents

Note, that the end document has no validation. We don't make any scans, we just store the link

## How to test this change
1. Create a model plan in the backend
2. Create a document link . 
3. View the documents on the front end
4. Upload a document on the front end
5.  Ensure both types of documents still work (note that the linked document formatting will be updated in a separate frontend ticket when the feature is released)

## PR Author Review Checklist

## PR Reviewer Guidelines


ClayBenson94 (commented):
I am good with this implementation (it functions as we need it to), but wanted to ask: Had you considered a separate type for these documents (and returning a union type of something like `PlanDocumentFile | PlanDocumentURL`)?

I assume so, but this implementation is much simpler to get rolling, but wanted to check! I ask because a _lot_ of the fields are no longer needed when making a plan document as a URL (instead of a file) and they even have different input types and mutations

StevenWadeOddball:
> I am good with this implementation (it functions as we need it to), but wanted to ask: Had you considered a separate type for these documents (and returning a union type of something like `PlanDocumentFile | PlanDocumentURL`)?
> 
> I assume so, but this implementation is much simpler to get rolling, but wanted to check! I ask because a _lot_ of the fields are no longer needed when making a plan document as a URL (instead of a file) and they even have different input types and mutations


Thanks @ClayBenson94 , yes, I thought about it but decided against it. I did a quick sync in standup with frontend yesterday, and also went away from using a union type to distinguish the different type of document. It sounded like there wasn't a benefit to them over simply returning a `document type` / `isLink` field.

I also considered creating an entirely new type and database tables, but I don't know if it's worth the time to implement. It's true the fields are different (and not needed for things like virus scan status). However, we'd need at least two new tables for this (`plan_document_url`, `plan_document_url_solution_link` (Or modify the link table)).

It's definitely doable, but I'm not sure that it is worth the extra effort it would take to develop on the frontend and on the backend. With both staying in the same table, we get a lot of functionality (linking a document, viewing the document table) for free. What do you think?

ClayBenson94:
> > I am good with this implementation (it functions as we need it to), but wanted to ask: Had you considered a separate type for these documents (and returning a union type of something like `PlanDocumentFile | PlanDocumentURL`)?
> > I assume so, but this implementation is much simpler to get rolling, but wanted to check! I ask because a _lot_ of the fields are no longer needed when making a plan document as a URL (instead of a file) and they even have different input types and mutations
> 
> Thanks @ClayBenson94 , yes, I thought about it but decided against it. I did a quick sync in standup with frontend yesterday, and also went away from using a union type to distinguish the different type of document. It sounded like there wasn't a benefit to them over simply returning a `document type` / `isLink` field.
> 
> I also considered creating an entirely new type and database tables, but I don't know if it's worth the time to implement. It's true the fields are different (and not needed for things like virus scan status). However, we'd need at least two new tables for this (`plan_document_url`, `plan_document_url_solution_link` (Or modify the link table)).
> 
> It's definitely doable, but I'm not sure that it is worth the extra effort it would take to develop on the frontend and on the backend. With both staying in the same table, we get a lot of functionality (linking a document, viewing the document table) for free. What do you think?

Absolutely - let's keep it as-is. Just wanted to make sure it was something you'd considered as you implemented! If you assessed it to be more work than is necessary (which I'd tend to agree, looking at it from my perspective), then we're all good with this 😁

ClayBenson94 (approved):

ClayBenson94 (approved):

PR #723, opened 2023-09-28
EASI-3163: Refactor MINT Dockerfiles

Jdwoodson:
# EASI-3163: Refactor MINT Dockerfiles

## Changes and Description

- Dockerfile: 
  - consolidate COPY lines for go.mod go.sum to one line
  - consolidate COPY lines for cmd and pkg to one line
  - reorganize COPY lines for certs and tzinfo to pull from base to increase cache hit chance
- Dockerfile.db_clean:
  - consolidate all ENV variables to one line for a single layer!
- Dockerfile.db_migrations:
  - specify default ARG TAG=9.10 for best practices
  - consolidate all ENV variables to one line for single layer!
  - remove duplicate default ARG from docker-compose.backend.yml when building mint-db-migrate image
- Remove Dockerfile.db_seed as completely unused in MINT
- remove scripts/run-cypress-test-docker as outdated and now unused in CI pipeline/local development

## PR Author Review Checklist

## PR Reviewer Guidelines


Dockerfile
```diff
@@ -1,49 +1,47 @@
...
-	# Copy build args to envs for version, datetime, and timestamp
-	ARG ARG_APPLICATION_VERSION
-	ARG ARG_APPLICATION_DATETIME
-	ARG ARG_APPLICATION_TS
-	ENV APPLICATION_VERSION=${ARG_APPLICATION_VERSION}
-	ENV APPLICATION_DATETIME=${ARG_APPLICATION_DATETIME}
-	ENV APPLICATION_TS=${ARG_APPLICATION_TS}
+COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
+COPY --from=build /usr/share/zoneinfo /usr/share/zoneinfo
+COPY --from=build /mint/bin/mint ./
```
ClayBenson94, line 34: Does re-ordering (or adding the trailing slash to the destination) accomplish anything here? Fine with the change itself, but was curious if there was a performance benefit to doing it in this order (I suppose the binary is more likely to differ than the zoneinfo / certs, so putting it later results in better caching of the final image?)

Dockerfile.db_migrations
```diff
@@ -1,11 +1,11 @@
-ARG TAG=
+ARG TAG=9.10
```
ClayBenson94, line 1: Is there a reason we default to `9.10` here but `9.10-alpine` in `Dockerfile.db_clean`? Perhaps this was added when building locally?

ClayBenson94: TIL: Defining the arg in the dockerfile still allows it to be overriden, this is basically just the default value. 

@Jdwoodson Can you:
- Add `9.10` to the DB Clean Dockerfile
- Modify the pipeline to build `9.10-alpine` for deploying to ECS
- Remove references to `9.10` in the compose files (since we shouldn't need to specify it if it's in the dockerfile, only if we want to build a different image, i.e. for ECS)

ClayBenson94 (changes requested):

Dockerfile
Jdwoodson: This was a mistake! I wanted to reorganize these layers, but ALSO pull certs and tzinfo from base instead of build. This should help caching of those layers.

Dockerfile.db_migrations
Jdwoodson: @ClayBenson94 correct about the default ARG value. Just makes the dockerfile complete to run.

There was a reason I went with alpine for clean and not for migrations. trying to find my notes on it because I thought I marked down why.

Jdwoodson: TL;DR for historical purposes - db_clean runs only in CI and we can use alpine image successfully 100% of time. db_migrations runs locally and in ci.  As part of local dev process, using mutli-arch image allows for devs on different architectures to not have to specifically define their arch. in CI, we can use alpine 100%.

docker-compose.backend.yml
Jdwoodson, line 1: Changes resolved

ClayBenson94 (approved):
Changes look good to me, and CI builds passing gives me confidence in the dockerfiles themselves!

PR #740, opened 2023-10-09
[EASI-3215] Allow multiple roles for collaborators

OddTomBrooks:
# EASI-3215

## Changes and Description

- Added flyway migration updating team_role to team_roles array
- Updated gql schema
- Updated resolvers
- Updated data model

## How to test this change

- Bring the systems up
- Create a plan collaborator using Postman
- Ensure the expected roles exist within Postman
- Update a plan collaborator's roles using Postman
- Delete the final Model Lead plan collaborator using Postman
  - This should throw an error

## PR Author Review Checklist

## PR Reviewer Guidelines


migrations/V107__Alter_Team_Role_To_Array.sql
```diff
@@ -0,0 +1,24 @@
...
+  BEFORE UPDATE ON plan_collaborator
+  FOR EACH ROW
+  WHEN ('MODEL_LEAD' = ANY(old.team_roles) AND 'MODEL_LEAD' != ANY(new.team_roles))
+EXECUTE FUNCTION COLLABORATOR_ROLE_CHECK_TRIGGER();
+
+CREATE TRIGGER collaborator_lead_req_delete
+  BEFORE DELETE ON plan_collaborator
+  FOR EACH ROW
+  WHEN ('MODEL_LEAD' = ANY(old.team_roles))
+EXECUTE FUNCTION COLLABORATOR_ROLE_CHECK_TRIGGER();
```
ClayBenson94, line 22: I don't think this function will continue to work as it was written [in this migration](https://github.com/CMSgov/mint-app/blob/8ad718c30bf3513130d9f27b6285869207797ba0/migrations/V8__Plan_Collaborator_Lead_Trigger.sql#L1-L18) - I think it'll need to be updated to handle the new column name and type

pkg/email/template_service_impl.go
```diff
@@ -70,6 +70,9 @@ var modelPlanShareSubjectTemplate string
 //go:embed templates/model_plan_share_body.html
 var modelPlanShareBodyTemplate string
 
+//go:embed templates/shared_header.html
+var sharedHeaderTemplate string
+
```
ClayBenson94, line 75: Guessing this will resolve itself when `main` is merged in, but seeing a lot of email template changes in this PR as-is

pkg/models/plan_collaborator.go
```diff
@@ -40,3 +41,19 @@ const (
...
 	TeamRolePayment    TeamRole = "PAYMENT"
 )
+
+func ConvertStringArrayToTeamRoles(sa pq.StringArray) []TeamRole {
+	roles := make([]TeamRole, len(sa))
+	for i, s := range sa {
+		roles[i] = TeamRole(s)
+	}
+	return roles
+}
```
ClayBenson94, line 51: We already actually have a generic function for this!

https://github.com/CMSgov/mint-app/blob/79190855a4ab76afe7a2257c6c1054fb179ab2ba/pkg/models/model_helpers.go#L44-L51

pkg/shared/utilityCollections/search.go
```diff
@@ -0,0 +1,10 @@
+package utilityCollections
+
+func Contains[T comparable](s []T, elem T) bool {
+	for _, a := range s {
+		if a == elem {
+			return true
+		}
+	}
+	return false
+}
```
ClayBenson94, line 10: Is this used anywhere? If not, we should delete it

We also often use [samber/lo](https://github.com/samber/lo) for functions like [this](https://github.com/samber/lo#contains)

ClayBenson94 (changes requested):

pkg/email/template_service_impl.go
OddTomBrooks: Yes. I branched off of the shared email branch on accident 😮‍💨 Surprised there are diffs though, expected it to not have diff against main.

pkg/shared/utilityCollections/search.go
OddTomBrooks: Good catch. I didn't like the process of looping through a collection but this attempt did not work. Will give samber a try next time.

migrations/V109__Alter_Team_Role_To_Array.sql
ClayBenson94, line 1: image

It's probably worth adding a check here to ensure that you can't have multiple of the same role for one collaborator

migrations/V109__Alter_Team_Role_To_Array.sql
```diff
@@ -0,0 +1,40 @@
...
+    RAISE EXCEPTION 'There must be at least one MODEL_LEAD assigned to each model plan';
+  END IF;
+
+  IF (TG_OP = 'DELETE') THEN
+    RETURN OLD;
+  ELSE
+    RETURN NEW;
+  END IF;
+END
+$role_check$ LANGUAGE plpgsql;
```
ClayBenson94, line 28: image

Something isn't working correctly with this function locally. When I attempt to modify a collaborator (only collaborator on the plan, currently only has `["MODEL_LEAD"]` as their roles, I get this error even though the target roles also contain `["MODEL_LEAD"]`

However, I _don't_ get that error when the _only_ role I target is `"MODEL_LEAD"` - seems to be something with the `newRoles` array containing things that aren't `"MODEL_LEAD"` (seen below)

image

ClayBenson94 (changes requested):

StevenWadeOddball (commented):
Thanks Tom, I did some quick functional testing which worked as expected. I will continue with reviewing the code in the morning.

StevenWadeOddball (approved):
Thanks Tom, this looks good. I tested according the testing instructions and it worked as expected.

OddTomBrooks:
This feature is complete from a backend perspective. I used git CLI to merge EASI-3217 into this branch which surprisingly caused GitHub to close PR #762. @garyjzhao please address the feedback from @patrickseguraoddball in this branch, as soon as we have front-end approval we will merge this to main.

patrickseguraoddball (changes requested):
@garyjzhao Per the [figma](https://www.figma.com/file/M72DReE73SSXMN48WZfaKN/MINT---New-Model-Process-(Model-Intake)?type=design&node-id=6093-119354&mode=design&t=oJLtJ6UvhM3IB8rL-0) if there is only one collaborator/model lead, there should still be the ability to edit.  Once on the edit page, the ability to deselect a model lead should be disabled.  As well as the tag to remove a role - that X should be gone.  This is not a small change so we might want to loop back with UX and team to make sure we can invest in this.

patrickseguraoddball:
Also the model lead always needs to be listed first [as shown here](https://www.figma.com/file/M72DReE73SSXMN48WZfaKN/MINT---New-Model-Process-(Model-Intake)?type=design&node-id=6169-124578&mode=design&t=rK6pbZmh26qPij4q-0)

The combobox "x" need to not render if there is only model lead in the selection [as shown here](https://www.figma.com/file/M72DReE73SSXMN48WZfaKN/MINT---New-Model-Process-(Model-Intake)?node-id=6093%3A119224&mode=dev)

patrickseguraoddball (changes requested):
I'm running into an issue where I have a model lead already created.  Then I go to create another collaborator and select them as a model lead.  Then before I submit, I can't remove the tag for model lead, because it still thinks there's only one model lead

Screenshot 2023-10-26 at 12 44 06 PM

patrickseguraoddball:
The app is also crashing on other multiselect components.  Here is the output

Screenshot 2023-10-26 at 12 45 09 PM

patrickseguraoddball:
I notice that the clear indicator on the multiselect is never visible.  It should be visible if there are any selections and there is at least one model lead.  Here is what the multiselect looks like in dev

Screenshot 2023-10-26 at 12 47 47 PM

patrickseguraoddball:
I notice that the clear indicator on the multiselect is not visible.  This seems related to the tag closing functionality listed above.  The clear indicator only appears if there are two model leads selected.  It should be visible if there are any selections and there is at least one model lead.  Here is what the multiselect looks like in dev.  

Screenshot 2023-10-26 at 12 47 47 PM

src/components/shared/MultiSelect/index.tsx
```diff
@@ -223,6 +236,28 @@ const MultiSelect = ({
...
+          a.value.localeCompare(b.value)
+        );
+      default:
+        return [
+          ...selected.filter(tag => tag.value === order),
+          ...selected.filter(tag => tag.value !== order)
+        ];
+    }
+  };
+
```
patrickseguraoddball, line 260: I think we don't want to default to ASC for sorting.  In almost all existing cases, we want to leave the tag in the order they are selected, so no sort necessary.

patrickseguraoddball (changes requested):

patrickseguraoddball (changes requested):
Sorry I missed this earlier but the alert isn't reflective of the current model lead.  This alert is reading as if the current collaborator is a model lead, but they arent

Screenshot 2023-10-26 at 3 22 32 PM

patrickseguraoddball:
This should say `Roles`  instead of `Selected options`

Screenshot 2023-10-26 at 3 24 44 PM

https://www.figma.com/file/M72DReE73SSXMN48WZfaKN/MINT---New-Model-Process-(Model-Intake)?node-id=6093%3A119224&mode=dev

patrickseguraoddball:
I think this whole page was intended to be revamped

https://www.figma.com/file/M72DReE73SSXMN48WZfaKN/MINT---New-Model-Process-(Model-Intake)?node-id=6199%3A10400&mode=dev

patrickseguraoddball:
The roles should appear on the task list per Figma

https://www.figma.com/file/M72DReE73SSXMN48WZfaKN/MINT---New-Model-Process-(Model-Intake)?node-id=6199%3A10400&mode=dev

https://www.figma.com/file/M72DReE73SSXMN48WZfaKN/MINT---New-Model-Process-(Model-Intake)?node-id=6201%3A11454&mode=dev

patrickseguraoddball:
This should now be `Manage model team members`

Screenshot 2023-10-26 at 3 30 43 PM

https://www.figma.com/file/M72DReE73SSXMN48WZfaKN/MINT---New-Model-Process-(Model-Intake)?node-id=6204%3A12404&mode=dev

patrickseguraoddball (changes requested):
I realize we are still getting the non-model leads collaborators aren't sorted alphabetically by last name.  I think if we really want to get this functionality we need to do the both sorting outside of the table, and don't do any initial sort.

Screenshot 2023-11-01 at 1 05 43 PM

Additionally this sorting functionality should be replicated for both the task list and the readonly Teams page.  Can probably use the same sort util for all three

patrickseguraoddball (approved):
LG Thanks Gary!

PR #762, opened 2023-10-18
[EASI-3217] Allow multiple roles on frontend

garyjzhao:
# EASI-3217

## Changes and Description

- Update `AddCollaborator` to take an array of new roles 
    - using multiselect over single select
- Updated `collaborator.spec.js` e2e test
- Update various tests to take in `teamRoles: ['MODEL_LEADS']`

## How to test this change

1. Add new collaborator to a model plan
2. Ensure you can select multiple roles
3. After adding, check to see a comma separated list instead of a single role

## PR Author Review Checklist

## PR Reviewer Guidelines


patrickseguraoddball (changes requested):
After adding a collaborator with multiple roles, then going back in to the form to edit, I can remove all roles and submit successfully.  I don't think this is intended behavior to allow someone to not have a role

I assumed with had guards on the mutation as well.  So we might want to revisit the input of the mutation as well as the form validation

Screenshot 2023-10-19 at 8 56 21 AM

patrickseguraoddball:
This above issue also affects the guard we had in place to always have at least one model lead.  The mutation throws and error if there is no model lead but other roles selected. It doesn't throw an error when there is no model lead and no other roles selected

src/views/ModelPlan/Collaborators/AddCollaborator/index.tsx
```diff
@@ -125,7 +125,7 @@ const Collaborators = () => {
           input: {
             modelPlanID: modelID,
             userName: username!,
-            teamRole: teamRole!
+            teamRoles: teamRoles!
```
patrickseguraoddball, line 128: Previously we had this as non-nullable because the form element (select) didnt make it possible to choose a null value.  Now with multiselect that is indeed possible.  We dont want to have this non null assertion here unless we have fullproof validation or adjust the form type to more accurately describe this field

src/views/ModelPlan/Collaborators/table.tsx
```diff
@@ -75,7 +75,10 @@ const CollaboratorsTable = ({
       {
         Header: collaboratorsMiscT('table.actions'),
         Cell: ({ row }: any) => {
-          if (row.original.teamRole === 'MODEL_LEAD' && isLastLead) {
+          if (
+            row.original.teamRoles.includes(TeamRole.MODEL_LEAD) &&
+            isLastLead
+          ) {
```
garyjzhao, line 81: This code addresses removing the `Edit` and `Remove` links to the only `Model Lead` in the collaborator list

src/validations/modelPlanCollaborators.ts
```diff
@@ -6,6 +6,7 @@ const CollaboratorsValidationSchema = Yup.object().shape({
   }),
   teamRoles: Yup.array()
     .of(Yup.string())
+    .min(1, 'Requires at least one role')
     .required('Enter a role for this team member')
 });
```
garyjzhao, line 11: This code ensures that there is at least one role in the `teamRoles` array. I think this is the simplest FE method of addressing the issue you brought up, @patrickseguraoddball . Thoughts?

garyjzhao:
Failing `server_test` is not caused changes introduced in this PR. Waiting for #740 to push up fix

src/validations/modelPlanCollaborators.ts
patrickseguraoddball: Yep this is great thanks!  I think BE is also working on a restraint for this, but it's best to have it in both places

patrickseguraoddball (changes requested):
I think you should still be able to edit, just not remove if there is only one model lead.  There is no button to edit this person when there should be

Screenshot 2023-10-20 at 4 29 16 PM

PR #778, opened 2023-11-02
Discussion Tagging BE

StevenWadeOddball:
# EASI-3286

## Changes and Description

- Implemented HTML sanitization
   -  tags allowed based on previous work on EASI
   - span allowed with attributes needed for a mention
- Interface / Union type to returned a tagged entity (currently a user account or a possible solution)
- Plan Discussion Created emails take HTML in the body
- Possible Operational Solution Get By key added
   - There is the potential to turn this into a dataloader, but I left this out to avoid scope creep
   
   ### Tag Structure
   `<span data-type="mention" tag-type="USER_ACCOUNT" class="mention" data-id="SKZO" data-label=\"Alexander Stark">@Alexander Stark</span>`
   - `data-type` --> must be "mention"
   - `tag-type` --> `USER_ACCOUNT` or `POSSIBLE_SOLUTION`
   - `class` --> must be "mention"
   - `data-id` --> the key for the data you are tagging, either a username, or a possible operational solution key
   - `data-label` --> Not required by the backend, but sent by the front end library
   - `data-id-db` --> The primary key of the tagged entity in the database. If this is set from the frontend, the backend won't lookup the information or update the tag
   
   After being processed by the backend, the tag is updated to include the `data-id-db` (the pk from the database)
   `<span data-type="mention" tag-type="USER_ACCOUNT" class="mention" data-id="SKZO" data-id-db="f0884065-1b56-40f3-a65b-0c85204e06b7" data-label="Alexander Stark">@Alexander Stark</span>`
   
### High Level Flow
1. Anything with the `scalar` type `TaggedHTMLInput` get's sanitized, and the mentions are parsed using regex.
2. The resolvers get the database primary key for the tagged entity, and update the rawContent of the Tagged html to include `data-id-db`.
3. The updated content is saved to the database in the relevant table (currently only plan discussion and discussion_reply)
4. The HTMLMentions are converted to Tags which are saved to the database.

### Returning tagged entities
1. Using a Union type in GQL, and an interface in GO, a possible operational solution or a user account is conditionally returned based on the tags provided table name, field name, and relevant key.

## How to test this change
1. Create a model Plan
2. Create a Discussion using the Postman Collection
    a. Look at the created email in MailCatcher to see the formatting for the mentions class
3. Create a Discussion reply using the Postman Collection

I'd also like feedback on error handling when creating tags. I suggest creating some bad tags / perhaps with EUAs that can't be found with local lookup. Do you agree with the approach or have suggestions for improvements? 

## PR Author Review Checklist

## PR Reviewer Guidelines


pkg/graph/schema.graphql
```diff
@@ -421,7 +421,10 @@ union TaggedEntity =  UserAccount | PossibleOperationalSolution
 TaggedContent represents content that has a tag in it. It is composed of the raw tag text, as well as the array of possible tags
 """
 type TaggedHTML {
-  rawContent: String! #TODO: SW change this to an HTML type?
+  """
+  RawContent is HTML. It is sanitized on the backend
+  """
+  rawContent: String!
```
StevenWadeOddball, line 427: I did some exploratory work at making an HTML scalar, but it felt convoluted. Instead, TaggedHTMLInput is a scalar which sanitizes the html, and leaves the front end representation as a string.

pkg/graph/resolvers/plan_discussion.go
```diff
@@ -42,11 +44,27 @@ func CreatePlanDiscussion(
...
+		//TODO: do we need to stop execution here? Should we silently continue instead? Should we filter out any bad tags?
+		// return nil, err
+	}
+
+	discussion, err := store.PlanDiscussionCreate(logger, planDiscussion)
 	if err != nil {
 		return nil, err
 	}
 
+	//TODO: should we put this in a transaction? Likely, discussions should be saved even if tags arent'
```
StevenWadeOddball, line 61: I'd like some feedback on if you think this should be blocking or not. I felt it was better to write the discussion than not. Also, what do you think about continuing if we can't match a tag to a database entity up above?

pkg/email/plan_discussion_created.go
```diff
@@ -10,7 +12,7 @@ type PlanDiscussionCreatedBodyContent struct {
 	ClientAddress     string
 	DiscussionID      string
 	UserName          string
-	DiscussionContent string
+	DiscussionContent template.HTML
```
StevenWadeOddball, line 15: This allows the template to render the content as html instead of escaped

pkg/email/templates/shared_style.html
```diff
@@ -135,6 +135,9 @@
       line-height: normal;
       color: #71767A;
     }
+    .mention {
+      color: #005EA2;
```
StevenWadeOddball, line 139: if desired in the future, this could be expanded to make the mentions have additional behavior

pkg/graph/resolvers/plan_discussion.go
```diff
@@ -174,13 +196,30 @@ func CreateDiscussionReply(
 	if err != nil {
 		return nil, err
 	}
+	err = UpdateTaggedHTMLMentionsAndRawContent(ctx, store, &discussionReply.Content, getAccountInformation)
 
-	result, err := store.DiscussionReplyCreate(logger, discussionReply)
-	return result, err
+	if err != nil {
```
StevenWadeOddball, line 201: Same as above. What are your thoughts about tags failing to parse / update stopping the discussion / reply being created?

pkg/graph/resolvers/tag.go
```diff
@@ -0,0 +1,148 @@
...
+	case models.TagTypePossibleSolution:
+		return PossibleOperationalSolutionGetByID(logger, store, *EntityIntID)
+	default:
+		return nil, fmt.Errorf(" no tagged entity configured for table: %s", tagType)
+	}
+
+}
+
+// UpdateTaggedHTMLMentionsAndRawContent updates the tagged html with the correct entity ids, and updates the RAW HTMl with the new representation of the mentions
+func UpdateTaggedHTMLMentionsAndRawContent(ctx context.Context, store *storage.Store, tHTML *models.TaggedHTMLInput, getAccountInformation userhelpers.GetAccountInfoFunc) error {
```
StevenWadeOddball, line 76: I'd like some thoughts on this function please. I'd almost make it a method, but it requires the store.  If it was made a method, we could perhaps pass another function that abstracts getPossibleOperationalSolutionByKey store method.

The benefit to a method is that the function modifies the `TaggedHTMLInput`. I wonder if that would be cleaner? Or perhaps we could return the modified tagged html instead of modifying the instance?

pkg/graph/resolvers/tag.go
```diff
@@ -0,0 +1,148 @@
...
+		}
+
+		// Update the Tagged HTML Raw content by replacing the tags old value, with the new representation
+		newTotalRaw := strings.Replace(string(tHTML.RawContent), string(mention.RawHTML), string(newHTML), -1)
+		tHTML.RawContent = models.HTML(newTotalRaw)
+
+		mention.RawHTML = newHTML
+	}
+	if len(errs) > 0 {
+		return fmt.Errorf("issues encountered getting database ids for tagged entities. %v", errs) // We aren't wrapping these errors because this is an array
```
StevenWadeOddball, line 124: This method results in 1 bad mention not stopping execution of the rest of the mentions being analyzed.

pkg/graph/resolvers/tag.go
```diff
@@ -0,0 +1,148 @@
...
+	return nil
+}
+
+// TagCollectionCreate converts an array of mentions, and creates an array in the database for unique tags.
+func TagCollectionCreate(logger *zap.Logger, store *storage.Store, principal authentication.Principal,
+	taggedField string, taggedTable string, taggedContentID uuid.UUID, mentions []*models.HTMLMention) ([]*models.Tag, error) {
+
+	tags := models.TagArrayFromHTMLMentions(taggedField, taggedTable, taggedContentID, mentions)
+
+	uniqTags := lo.UniqBy(tags, func(tag *models.Tag) string {
```
StevenWadeOddball, line 135: I'd like opinions on this please. This prevents an entity being tagged in the same table twice. There might be multiple mentions referring to the entity, but they would only have one tag. I think this will be useful for in app notifications as they come up

pkg/models/html.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	htmlPackage "html"
+	"html/template"
+
+	"github.com/cmsgov/mint-app/pkg/sanitization"
+)
+
+// hTML represents html code. It is sanitized when unmarshaled from graphQL or when converted to hTML to only allow specific tags
+type hTML string
```
StevenWadeOddball, line 11: By not exporting the type, we can effectively control when the data is being sanitized.

pkg/models/html.go
```diff
@@ -0,0 +1,39 @@
...
+		return ""
+	}
+	return sanitization.InnerHTML(h.String())
+}
+
+func (h hTML) String() string {
+	// TODO: Remove this hotfix that was introduced as a result of entities being rendered as escaped HTML in non-rich-text views
+	// (Rich Text views handle this escaped data properly, so this hotfix is only needed until we implement rich text views across the board)
+	// Note: We only really need to do this here, instead of ALSO doing it in email code because the encoded HTML is handled cleanly by the html/template package
+	return htmlPackage.UnescapeString(string(h))
```
StevenWadeOddball, line 38: This is following the paradigm established in EASI.

pkg/models/plan_discussion.go
```diff
@@ -8,7 +8,7 @@ import (
 type PlanDiscussion struct {
 	baseStruct
 	modelPlanRelation
-	Content             string              `json:"content" db:"content"`
+	Content             TaggedHTMLInput     `json:"content" db:"content"`
```
StevenWadeOddball, line 11: This is set as TaggedHTMLInput instead of TaggedHTML, so that we can return a type in GQL instead of a scalar

pkg/storage/tagStore.go
```diff
@@ -0,0 +1,110 @@
...
+func (s *Store) TagCollectionCreate(logger *zap.Logger, tags []*models.Tag, createdBy uuid.UUID) ([]*models.Tag, error) {
+
+	retTags := []*models.Tag{}
+	stmt, sErr := s.db.PrepareNamed(tagCreateCollectionSQL)
+	if sErr != nil {
+		return nil, sErr
+	}
+	defer stmt.Close()
+	errs := []*error{}
+	mapSlice := []map[string]interface{}{}
```
StevenWadeOddball, line 57: This approach is similar to the serialization occurring in the data loader. It also means, it only creates tags if all tags are valid. One invalid tag means none of them will be saved.

pkg/storage/tagStore.go
```diff
@@ -0,0 +1,110 @@
...
+		return nil, err
+	}
+	return retTags, nil
+
+}
+
+// TagCollectionGetByContentIDAndField returns relevant tags for specific table and field
+func (s *Store) TagCollectionGetByContentIDAndField(_ *zap.Logger, taggedTable string, taggedField string, taggedContentID uuid.UUID) ([]*models.Tag, error) {
+
+	var tags []*models.Tag
```
StevenWadeOddball, line 90: This could also be a dataloader

StevenWadeOddball:
Note, failing frontend tests are addressed in the FE branch that will be moved to the feature branch.

pkg/models/html.go
ClayBenson94: I think per go standards this might actually just be `html` instead of `hTML`

pkg/email/plan_discussion_created.go
ClayBenson94: Can you add this as an inline comment on the field?

pkg/graph/resolvers/plan_discussion.go
```diff
@@ -174,13 +196,30 @@ func CreateDiscussionReply(
...
+		return reply, err
+	}
+	reply.Content.Tags = tags
+	return reply, err
 }
 
 // UpdateDiscussionReply implements resolver logic to update a Discussion reply object
 func UpdateDiscussionReply(logger *zap.Logger, id uuid.UUID, changes map[string]interface{}, principal authentication.Principal, store *storage.Store) (*models.DiscussionReply, error) {
+	// THIS IS NOT USED by the front end. If it is ever used, make sure to handle tags
```
ClayBenson94, line 222: Nit: Can you update these comments to be in the function comments using the `Deprecated:` prefix? That way they'll show up to anyone trying to call the functions

https://stackoverflow.com/a/36360323

pkg/graph/resolvers/plan_discussion.go
ClayBenson94: I think I'd rather the functionality of tagging be atomic. I think if we can't match up a tag to an entity in the DB (currently an account or a soln), that somewhat invalidates the content of the tag, and suceeding (silently) might lead to unexpected behavior

It's a bit more complicated in the case of creating the tag, I think -- since the plan discussion has to be created before the tag is created (so it has reference to the ID)

This is probably worth syncing on to see if we can come up with a way (maybe just for these 2 DB calls) to wrap in a transaction.

pkg/graph/resolvers/tag.go
ClayBenson94: I think this approach is generally fine! I think one of two things would help clarfify expectations of this function:

1. Add comments to the function definition that clarify that the `tHTML` input type _will_ be modified, and that that's the expectation of the function
2. (As you suggested) update the function to return `(tHTML, error)`, and make the code _not_ modify the `*tHTML` input (in this case, I'd suggest updating the comments as well just to be extra clear!)

pkg/graph/resolvers/tag.go
ClayBenson94: Per the comment I left in `plan_discussion.go` (around possibly making this function end early & not creating the discussion/tag entities) it's possible that we might not need this if we go with that approach (since analyzing other mentions/tags wouldn't do anything)

If we go that route, I'd update this function as well to just immediately return the first error we get

pkg/graph/resolvers/plan_discussion.go
ClayBenson94: See comment from above! (As well as in `tag.go`)

pkg/models/utilitites.go
ClayBenson94, line 1: Are `StructArrayToJSONArray` or `StructArrayToMapArray` used anywhere?

ClayBenson94 (changes requested):
I left a few specific comments on some spots, but the only other high-level feedback I have is around some of the naming of `Tag` and `Mention`

There's a few spots where we use each of those terms, and I'm not sure what denotes one vs. the other, or if they're meant to represent the same thing -- perhaps we should try and normalize using one over the other?

pkg/graph/resolvers/plan_discussion.go
StevenWadeOddball: That sounds good! I have a few ideas for this, I think perhaps a similar approach to how we wrap delete statements in a transaction would also be good!

pkg/models/html.go
StevenWadeOddball: thanks Clay, I updated the name.

pkg/models/utilitites.go
StevenWadeOddball: It initially was, ( I was going to call that method instead of manually doing this in `TagCollectionCreate`). I ended up converting the tags to  a map manually there because I had to iterate through all the tags anyways, so it saved an iteration of looping. 

I could see removing them from the code base, or I could see leaving them. I could use them in the TagCollectionCreate function to show their use, but it would be less efficient

migrations/V110__Add_Tag_Table.sql
```diff
@@ -0,0 +1,24 @@
...
+    entity_uuid UUID,
+    entity_intid int,
+    created_dts timestamp with time zone NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    modified_dts timestamp with time zone,
+    created_by uuid NOT NULL REFERENCES user_account(id),
+    modified_by uuid REFERENCES user_account(id)
+);
+
+
+ALTER TABLE tag
```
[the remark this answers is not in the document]
StevenWadeOddball: Good call, I added this as a SQL comment, so it will also be visible outside of the migration file

pkg/graph/resolvers/tag.go
StevenWadeOddball: Thanks, I updated the comment to make it even clearer that the tHTML and mention are updated in this function call.

I also extracted some of these functions in your suggstion Tom. I left some code part of the orchestrator where I thought it was more readable to do so. You'll also notice the update methods are combined, as you can't update the individual mention content before updating the parent.

StevenWadeOddball:
Thanks @ClayBenson94  and @OddTomBrooks, changes are in! I also changed naming per our previous conversation Clay. The scalar is now TaggedHTML, and the return type is TaggedContent. 

I also added some comments around Mention vs Tag to clarify what they represent, since they are distinct, but still both needed.

ClayBenson94 (approved):

PR #785, opened 2023-11-06
[EASI-3509] - Fixed null render on readonly

patrickseguraoddball:
# EASI-3509

## Changes and Description

- Fixed null render values for readonly

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/src/views/ModelPlan/ReadOnly/GeneralCharacteristics/index.tsx
+++ b/src/views/ModelPlan/ReadOnly/GeneralCharacteristics/index.tsx
@@ -140,7 +140,10 @@ const ReadOnlyGeneralCharacteristics = ({
           'isNewModel',
           <ReadOnlySection
             heading={generalCharacteristicsT('isNewModel.label')}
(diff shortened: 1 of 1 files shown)


ClayBenson94 (approved):
Tested locally - looks good!

The reason that this works (and doesn't render `''`) seems to be because the `ReadOnlySection` specifically handles falsy `copy` properties as using the `miscellaneous` "`na`" translation (which took me a sec, looking at this PR!)

Actually I think it was due to the template literal being passed a null value, and doesn't have a map for null.  So instead of returning null, it returns a string of the unmapped template literal - `model.options.null`

patrickseguraoddball:
> Tested locally - looks good!
> 
> The reason that this works (and doesn't render `''`) seems to be because the `ReadOnlySection` specifically handles falsy `copy` properties as using the `miscellaneous` "`na`" translation (which took me a sec, looking at this PR!)

Actually I think it was due to the template literal being passed a null value, and doesn't have a map for null.  So instead of returning null, it returns a string of the unmapped template literal - `model.options.null`

PR #829, opened 2023-11-29
[EASI-3664] Drop the existing_model column from plan_general_characteristics

OddTomBrooks:
# EASI-3664

## Changes and Description

- Add migration dropping existing_model column

## PR Author Review Checklist

## PR Reviewer Guidelines


OddTomBrooks:
Do not merge until model plan data is updated

StevenWadeOddball (changes requested):
Thanks Tom, the migration works good. 

I noticed when reviewing the general characteristics postman that the `PlanGeneralCharacteristicsChanges` type hasn't been updated.
`  existingModel: String` should not be part of it anymore. Can you please update?

Also, can you update the postman query so you can just click send and it works? I get an error when trying to run it because it isn't filled out. It would be ideal if we could open the collection and send the query.

ClayBenson94:
> Do not merge until model plan data is updated

@OddTomBrooks What remains to be updated here? Is there data in prod that needs to get updated for this, I assume?

OddTomBrooks:
> Thanks Tom, the migration works good.
> 
> I noticed when reviewing the general characteristics postman that the `PlanGeneralCharacteristicsChanges` type hasn't been updated. ` existingModel: String` should not be part of it anymore. Can you please update?
> 
> Also, can you update the postman query so you can just click send and it works? I get an error when trying to run it because it isn't filled out. It would be ideal if we could open the collection and send the query.

Requested changes are in except turnkey click-send on the plan general characteristics update query. As discussed async, this is a design change from the existing query structure which would cause a loss of structured information by deleting various variables on the input. Such a change is outside the scope of the ticket and is best discussed as whether it's a design direction we want to take and homogenize similar queries.

PR #840, opened 2023-12-06
Easi 3556/email selected solution pocs

StevenWadeOddball:
# EASI-3556

## Changes and Description

- Introduced filter views to the possible operational solution table
- Created a new email template for when a solution is selected
    - Created a new database query to get the data (otherwise there would be 9 db queries at a minimum)
    - The email only sends for solutions that aren't custom (`isOther = false`)
    - The email only sends if there are defined POCs for that solution

## How to test this change
1. To simply validate the template, run `scripts/dev email`
2. Compare with [Figma](https://www.figma.com/file/8MfXzXzmKWv8vBjTasUIkD/MINT---Notifications?type=design&node-id=2921-759&mode=design&t=mOhoasphhUyCQcU6-0)

Otherwise,
1. Bring up the app and make a model plan.
2. Make a solution before updating plan basics performance_period_starts (this is the model start date)
3. Look at the email that comes (model start date should be excluded from the email.

4. Update the date in model basics, and choose a new solution
5. The new email should have the model start date

Filter Views
1. Select a solution that has a pre-defined filter view.
2. Make sure the email link uses that views language, and links correctly in MINT (4innovation (4i) is a good example, it should link to IDDOC)
3. Do the same with a solution without a filter view (RMADA is an example)
4. Make sure the email only links to the read view, and the language is correct

## PR Author Review Checklist

## PR Reviewer Guidelines


patrickseguraoddball (changes requested):
Looks great!  Just a few minor style points

In the Figma, `Model Leads` should cover both singular and plural cases `Model Lead(s)`

Screenshot 2023-12-06 at 2 36 38 PM

https://www.figma.com/file/8MfXzXzmKWv8vBjTasUIkD/MINT---Notifications?node-id=2935%3A707&mode=dev

There should be a period after the email listed in the footer here 

Screenshot 2023-12-06 at 2 38 14 PM

https://www.figma.com/file/8MfXzXzmKWv8vBjTasUIkD/MINT---Notifications?node-id=2935%3A712&mode=dev

There could/should be a tad more spacing underneath the `View CBOSC details in MINT` link
Screenshot 2023-12-06 at 2 39 18 PM

StevenWadeOddball:
Thanks @patrickseguraoddball ! I corrected the punctuation, and the pluralization of model lead.  I took a look at the spacing after the view details link, I'm not sure the amount to add. In figma, it looks like it's configured as spacing between inside the square. I added a line break, but it looked like it was too much, so I left it out. Thoughts? I can add back if you like
image


I think we might want to check with product and ux about the intent of the solution status column. In some ways, I could see it making sense for it to always say `not_started`, but if that's the case, the status isn't important to display in the email. I think your solution would work good if needed, as you suggested in chat we could also handle subsequently on the backend, but it does pose some challenges as well.

patrickseguraoddball:
> Thanks @patrickseguraoddball ! I corrected the punctuation, and the pluralization of model lead. I took a look at the spacing after the view details link, I'm not sure the amount to add....

I think what you captured there in the screenshot looks correct!  Not a huge deal either way, thanks for the update!

patrickseguraoddball (approved):
LG, this works great!

PR #849, opened 2023-12-11
[EASI-3618] Update Precedence Rules

garyjzhao:
# EASI-3618

## Changes and Description

Backend Changes from @OddTomBrooks 
- Updated gql schema
- Created flyway migration adding new enum type and table columns
- Updated data model
- Updated Postman collection

Frontend Changes from @garyjzhao 
- Updated Precendence Rules question to now have a checkbox and additional detail text input
- Updated `precendenceRules` in `src/i18n/en-US/modelPlan/beneficiaries.ts` to incorporate new YesNo options
- Added new 'No additional information specified' copy to `miscellaneous.ts`
- Update tests and snaps

## How to test this change

BE:
- Bring servers up
- Verify migration succeeded
- Run plan beneficiaries update endpoint in Postman
- Verify success and data in response

FE:
- Navigate to Frequency questions in Beneficiaries
- See changes have been made to the question
- Ensure data is recorded in read only and csv

## PR Author Review Checklist

## PR Reviewer Guidelines


StevenWadeOddball (changes requested):
Thanks @garyjzhao and @OddTomBrooks . Requesting Changes based on feedback in progress on https://github.com/CMSgov/mint-app/pull/842#discussion_r1422910149.

OddTomBrooks:
> Thanks @garyjzhao and @OddTomBrooks . Requesting Changes based on feedback in progress on [#842 (comment)](https://github.com/CMSgov/mint-app/pull/842#discussion_r1422910149).

Requested changes are implemented!

src/types/translation.ts
```diff
@@ -284,7 +285,7 @@ export type TranslationBeneficiaries = {
   beneficiarySelectionFrequencyOther: TranslationFieldProperties;
   beneficiaryOverlap: TranslationFieldPropertiesWithOptions<OverlapType>;
   beneficiaryOverlapNote: TranslationFieldProperties;
-  precedenceRules: TranslationFieldProperties;
+  precedenceRules: TranslationFieldPropertiesWithOptions<YesNoType>;
   status: TranslationFieldPropertiesWithOptions<TaskStatus>;
```
patrickseguraoddball, line 289: Both of the `Please describe fields` require their own translation propery.  You can reference other examples in this file.  They shouldn't be translated under `miscellaneous` as they need to be able to rendered dynamically

src/views/ModelPlan/TaskList/Beneficiaries/Frequency/index.tsx
```diff
@@ -362,28 +369,75 @@ const Frequency = () => {
...
+                              </FieldErrorMsg>
+
+                              {getKeys(beneficiaryPrecedenceConfig.options).map(
+                                key => (
+                                  <Fragment key={key}>
+                                    <Field
+                                      as={CheckboxField}
+                                      id={`beneficiaries-precedence-rules-${key}`}
+                                      data-testid={`beneficiaries-precedence-rules-${key}`}
+                                      name={`precedence-rules-${key}`}
```
patrickseguraoddball, line 395: The value `name` here is the most important part of the Formik component.  It isn't merely a name, but the actually fileld name where the value is tied to the Formik library.  It should be `precedenceRules` that houses both values for yes and no

src/views/ModelPlan/TaskList/Beneficiaries/Frequency/index.tsx
```diff
@@ -362,28 +369,75 @@ const Frequency = () => {
...
+                                        if (e.target.checked) {
+                                          arrayHelpers.push(e.target.value);
+                                        } else {
+                                          const idx = values.precedenceRules.indexOf(
+                                            e.target.value as YesNoType
+                                          );
+                                          arrayHelpers.remove(idx);
+                                        }
+                                      }}
+                                    />
```
patrickseguraoddball, line 415: You shouldn't need the `FieldArray component, not the onChange handler here.  The CheckboxField should be able to handle this automatically.

src/views/ModelPlan/TaskList/Beneficiaries/Frequency/index.tsx
```diff
@@ -362,28 +369,75 @@ const Frequency = () => {
...
+                                          as={TextAreaField}
+                                          className="height-15"
+                                          error={flatErrors.precedenceRules}
+                                          id={`beneficiaries-precedence-rules-${key}-note`}
+                                          data-testid={`beneficiaries-precedence-rules-${key}-note`}
+                                          name={`precedenceRules${beneficiaryPrecedenceConfig.options[key]}`}
+                                        />
+                                      </div>
+                                    )}
+                                  </Fragment>
```
patrickseguraoddball, line 432: This component needs to be it's mapping in the translation file, just like the parent.  It shouldn't be run through the miscellaneous translation library

patrickseguraoddball (changes requested):
I think the form component needs a bit more attention, we can sync on these changes for anything that may be unclear.  I also want to sync on the intent of the translation mapping, and why it's important to include any user enter value as a way to make rendering dyncamic.  

There is missing the csv value here, and it doesn't show up in the csv
https://github.com/CMSgov/mint-app/blob/ec0c7db0f6ca3c727e1b346284c5eeef04d27ce2/src/utils/export/CsvData.ts#L181-L182

MINT.postman_collection.json
```diff
@@ -298,7 +298,7 @@
...
...DualElligibleDifferentNote\n        excludeCertainCharacteristics\n        excludeCertainCharacteristicsCriteria\n        excludeCertainCharacteristicsNote\n        numberPeopleImpacted\n        estimateConfidence\n        confidenceNote\n        beneficiarySelectionMethod\n        beneficiarySelectionOther\n        beneficiarySelectionNote\n        beneficiarySelectionFrequency\n        beneficiarySelectionFrequencyOther\n        beneficiarySelectionFrequencyNote\n        beneficiaryOverlap\n        beneficiaryOverlapNote\n        precedenceRules\n        precedenceRulesYes\n        precedenceRulesNo\n        precedenceRulesNote\n        createdBy\n        createdDts\n        modifiedBy\n        modifiedDts\n        readyForReviewBy\n        readyForReviewDts\n        readyForClearanceBy\n        readyForClearanceDts\n        status\n\n        createdByUserAccount \n        {\n            commonName\n        }   \n        modifiedByUserAccount\n        {\n            commonName\n        }     \n        readyForReviewByUserAccount \n        {\n            commonName\n        }\n        readyForClearanceByUserAccount \n        {\n            commonName\n        }        \n    }\n}",
```
StevenWadeOddball, line 301: Thanks for updating the return type. [Original Conversation](https://github.com/CMSgov/mint-app/pull/842#discussion_r1422890111) Can you also add on setting the values? It would be helpful to see that the values are set when you run the mutation.
```GQL
       precedenceRules: [YES,NO]
        precedenceRulesYes: "It is a yes"
        precedenceRulesNo: "It is a no"
        precedenceRulesNote: "Precedence notes go here"
```

migrations/V121__Update_Precedence_Rules.sql
```diff
@@ -0,0 +1,18 @@
+CREATE TYPE YES_NO_TYPE AS ENUM (
+  'YES',
+  'NO'
+  );
+
+ALTER TABLE plan_beneficiaries
+  ADD COLUMN precedence_rules_note ZERO_STRING;
+
+UPDATE plan_beneficiaries
+  SET precedence_rules_note = precedence_rules;
```
StevenWadeOddball, line 10: Thanks @OddTomBrooks ,

One potential issue with this, if the row wasn't ever updated, this update will error because of the auditing trigger. The `modified_by` and `modified_dts` fields should be updated to ensure that there isn't an error. In other migrations we use the system account UUID and current_timestamp for these fields

StevenWadeOddball (changes requested):
Thanks @OddTomBrooks, a couple of minor changes needed, but then this should be good to go from a backend perspective.

MINT.postman_collection.json
OddTomBrooks: Ah yes, I intended to migrate this talk to the open PR. Thanks for bringing this up.

We have a large variety of variables which do not have test values set. It is unclear to me which is the process we want standard, so I went with the majority of settings. Is there a reason we want values for these variables set but not others?

migrations/V121__Update_Precedence_Rules.sql
OddTomBrooks: Ahhh I didn't think of this! This worked in local testing because I had to set the modified_by / modified_dts field when I manually set the precedence_rules. Thanks for pointing this out, I'll go find a migration addressing this and copy its process. Will have to think about the testing process.

MINT.postman_collection.json
StevenWadeOddball: Thanks Tom, 

My thoughts here, it would be ideal to have values set for every field, but the postman scripts where largely first committed not filled out. It's not the biggest gain to do a tech debt ticket to fill out every field.

However, I think there is some value if you have to modify the postman script already to include the new values. This has the benefit of allowing a reviewer to verify that the new fields work, and have data filled out. Otherwise, we just have to do it every time we want to test.

 So I figure if we have a situation where we can easily add the new fields, it's ideal to improve it a little as we go. (And makes validating even easier)

OddTomBrooks: That makes sense. Since I understand now that we consider the fields without variables to be tech debt, I'll make a point in the future to leave the value assignments.

migrations/V121__Update_Precedence_Rules.sql
StevenWadeOddball: No worries! I sometimes do it with a couple of test cases with SQL. My DB management tool can save just some individual scripts for me, so I can run all the old migrations, then insert test data, then run the new one. It makes it a little less or a burden to test.

I'd probably suggest just having a record that isn't updated, (or perhaps add some filtering to the update statement, you don't really need to update records that don't have precedence rules, this will update everything currently). Still should probably set the modified properties to be safe, but the where clause would help also.

MINT.postman_collection.json
StevenWadeOddball: thanks Tom, I appreciate it. It seems like fields we are likely updating for our testing already, so it doesn't seem like we will have to do much more than we currently are.

StevenWadeOddball (dismissed):
Thanks for making these changes @OddTomBrooks , this looks good to me. Approving for the BE. 

This will need to be updated to address Patrick's concerns on the frontend still, but putting approval here to note BE should be good to go.

src/views/ModelPlan/TaskList/Beneficiaries/Frequency/index.tsx
```diff
@@ -362,28 +368,53 @@ const Frequency = () => {
...
+                                name="precedenceRules"
+                                label={beneficiaryPrecedenceConfig.options[key]}
+                                value={key}
+                                checked={values.precedenceRules.includes(key)}
+                              />
+
+                              {values.precedenceRules?.includes(key) && (
+                                <div className="margin-left-4">
+                                  <span>
+                                    {beneficiariesT('precedenceRulesYes.label')}
```
patrickseguraoddball, line 399: Ideally this would be decoupled from a single label, much like you did on like 407

```
{beneficiariesT('precedenceRules${beneficiaryPrecedenceConfig.options[key]}.label')}
```

src/views/ModelPlan/TaskList/Beneficiaries/Frequency/index.tsx
```diff
@@ -362,28 +368,53 @@ const Frequency = () => {
...
+                                  </span>
+                                  <Field
+                                    as={TextAreaField}
+                                    className="height-15"
+                                    error={flatErrors.precedenceRules}
```
patrickseguraoddball, line 404: This error isn't really needed, and honestly we overuse them throughout the app because we don't have much validation to even trigger errors.  If we did have validation here the error would need to be converted to a boolean - something like `!!flatErrors.precedenceRules`

patrickseguraoddball (changes requested):
Thanks for the changes.  I forgot to mention that the filter group mappings need to also be added to any relevant filter groups here - 

https://github.com/CMSgov/mint-app/blob/d66ced44d940a8ed29bd2207bcb0c612aea2c6df/src/views/ModelPlan/ReadOnly/_components/FilterView/BodyContent/_filterGroupMapping.ts#L453

patrickseguraoddball (approved):
LG

PR #854, opened 2023-12-14
Discussion reply email to discussion originator

StevenWadeOddball:
# EASI-3601

## Changes and Description

- New Email templates
   - new email added to email script
- New DB query to minimize db calls for reply emails

## How to test this change
1. Run the app
2. Make a discussion
3. Make a reply
   a. confirm the account you used to make the discussion is sent an email
4. Keep making replies
   a.  ensure that it always goes to the account that made the discussion
   b. ensure that the two most recent replies are displayed
   c. ensure that there is a count of all the remaining discussions.
   
The email can be validated by running scripts/dev email

## PR Author Review Checklist

## PR Reviewer Guidelines


cmd/test_email/main.go
```diff
@@ -25,6 +25,8 @@ func main() {
 	sendPlanDiscussionCreatedTestEmail(emailService, templateService, addressBook)
 	sendPlanDiscussionTaggedUserTestEmail(emailService, templateService, addressBook)
 	sendPlanDiscussionTaggedSolutionTestEmail(emailService, templateService, addressBook)
+	//DiscussionReply email
```
OddTomBrooks, line 28: Comment feels redundant

pkg/email/discussion_reply_created_originator.go
```diff
@@ -0,0 +1,76 @@
...
+
+	"github.com/google/uuid"
+	"github.com/samber/lo"
+
+	"github.com/cmsgov/mint-app/pkg/models"
+)
+
+// DiscussionReplyCreatedOriginatorSubject is the subject for hte email
+type DiscussionReplyCreatedOriginatorSubject struct {
+	UserName string // The full name of the user replying
```
OddTomBrooks, line 15: Consider changing variable name to `FullName` or other similar convention so it's more self-descriptive

pkg/email/discussion_reply_created_originator.go
```diff
@@ -0,0 +1,76 @@
...
+	ModelAbbreviation string
+	OriginatorName    string
+	OriginatorRole    string
+	Replies           []DiscussionReplyEmailContent // all relevant reply information
+}
+
+// DiscussionReplyEmailContent represents the replies for a discussion
+// These will be sorted from newest to oldest.
+type DiscussionReplyEmailContent struct {
+	UserName string // the full name of the user replying to email
```
OddTomBrooks, line 34: Same as above, consider `FullName` or similar

pkg/email/discussion_reply_created_originator.go
```diff
@@ -0,0 +1,76 @@
...
+		Content:  dreb.Content.RawContent.ToTemplate(),
+	}
+}
+
+// func (drebs *DiscussionRepliesEmailContentDB) ToDiscussionReplyEmailContent() []DiscussionReplyEmailContent {
+// 	replyEmails := lo.Map[*DiscussionRepliesEmailContentDB, DiscussionReplyEmailContent](drebs, func(reply *DiscussionReplyEmailContentDB, _ int) DiscussionReplyEmailContent {
+// 		return reply.ToDiscussionReplyEmailContent()
+// 	})
+// 	return replyEmails
+// }
```
OddTomBrooks, line 68: Is this code still relevant?

StevenWadeOddball: Thanks for catching! I'll remove it.

cmd/test_email/main.go
StevenWadeOddball: perhaps, I wanted to show that this is a section that we can add future discussion reply emails, vs an initial discussion email. I think it's probably extra verbose, but I'd prefer to keep it for this auxiliary package.

pkg/email/templates/discussion_reply_created_originator_body.html
```diff
@@ -0,0 +1,49 @@
...
+        <h2>Discussion</h2>
+        <br/>
+        <p><b>{{.OriginatorName}}</b></p>
+        <p class="subtitle"> {{.OriginatorRole}}</p>
+        <br/>
+        <p>{{.DiscussionContent}}</p>
+        <br/>
+
+        {{ $length := len .Replies }} {{ if gt $length 2 }}
+            <i class="subtitle">{{ len (slice (printf "%*s" $length "") 2) }} earlier {{if eq $length 3}} reply {{end}} {{if gt $length 3}} replies {{end}} not shown</i>
```
OddTomBrooks, line 19: This is not the end of the world, but it is possible for the reply chain to get very long and this potentially become inefficient to process through the template. It is a potential vulnerability, but would almost require an exploitative use of the system.

An alternative approach would be to share only up to K replies, as would be embedded in the email as well as a count of the replies in the thread. The expansion of complexity to address the potential vulnerability is likely not worthwhile, but for security we need to be conscious of these potential issues.

pkg/email/discussion_reply_created_originator.go
StevenWadeOddball: I agree in principal, but I leaned on the side of naming consistency. Other fields that hold full name data in other email templates call this data UserName as well. My compromise for this was to add the comment to make it clear this is users full name.

pkg/storage/SQL/email/discussion_reply_to_originator_details_get.sql
```diff
@@ -0,0 +1,21 @@
...
+    reply.id,
+    reply.discussion_id,
+    reply.content,
+    reply.user_role,
+    reply.user_role_description,
+    reply.is_assessment,
+    reply.created_dts
+    -- reply.created_by,
+    -- reply.modified_by,
+    -- reply.modified_dts
```
OddTomBrooks, line 14: Is this still relevant?

pkg/email/discussion_reply_created_originator.go
StevenWadeOddball: same as above.

pkg/storage/SQL/email/discussion_reply_to_originator_details_get.sql
StevenWadeOddball: Good call! thanks, I'll remove it

OddTomBrooks (approved):

pkg/email/templates/discussion_reply_created_originator_body.html
StevenWadeOddball: Thanks Tom, I think that is a good call out. I think I will note this as an improvement that can be updated for a future PR.

Something like 
1. pass a count of replies to the template
2. pass the two most recent replies to the template (instead of the whole array).

This would effectively stop the iteration

PR #860, opened 2023-12-18
NOREF/Discussion Reply email subject and template efficiency

StevenWadeOddball:
# [EASI-3601](https://jiraent.cms.gov/browse/EASI-3601)

## Changes and Description

- Update email subject to be based on model name
   - I discussed this with UX. I noticed while deploying that using the username of the user who made the reply results in emails being grouped oddly in gmail. To make them group more logically, the subject will be based on the name of the model plan.
- Update templating efficiency
    - in response to suggestion by @OddTomBrooks [here](https://github.com/CMSgov/mint-app/pull/854#discussion_r1428475833), I updated the template logic. We now explicitly pass the count of the number of replies, and only the first two actual replies.

## How to test this change
1. Run the app
2. Make a discussion
3. Make a reply
   a. confirm the account you used to make the discussion is sent an email
4. Keep making replies
   a.  ensure that it always goes to the account that made the discussion
   b. ensure that the two most recent replies are displayed
   c. ensure that there is a count of all the remaining discussions.

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/cmd/test_email/discussion_reply_emails.go
+++ b/cmd/test_email/discussion_reply_emails.go
@@ -38,6 +38,10 @@ func sendDiscussionReplyOriginatorEmail(
 	if err != nil {
 		return err
 	}
+	replyCount := len(replies)
+	if replyCount > 2 {
+		replies = replies[:2] // only retain the first two replies
+	}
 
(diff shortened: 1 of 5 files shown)


patrickseguraoddball (approved):
This works as expected, following the testing instructions!

PR #877, opened 2023-12-29
[NOREF] - Added util and test to correctly order readonly items

patrickseguraoddball:
# NOREF

## Changes and Description

Previously readonly enum values were not maintaining the correct order from Figma when rendered.  The BE does not persist order of multiselect.  This work add a util function to order the readonly values as they are defined in the translation file, rather than the default BE order

This work also defines another util function for readonly.  When values have multiple additional info/other values, we need to properly map those to their respective values.  For example `fundingSource` on `Payments` has 4 enum values.  3 of those 4 values all trigger additional fields with text values.  In the readonly, we need to match up those values to their conditional parent values, while in the correct order.

- Added util to correctly order readonly enum values
- Added util function to match up conditional fields with parent enum values
- Added unit testes

## How to test this change

These are merely util functions with no implementation yet.  Verify their functionality with the written unit tests

- Run the unit tests

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/src/views/ModelPlan/ReadOnly/_components/ReadOnlySection/index.test.tsx
+++ b/src/views/ModelPlan/ReadOnly/_components/ReadOnlySection/index.test.tsx
@@ -1,7 +1,13 @@
 import React from 'react';
 import { render, screen } from '@testing-library/react';
+import { FundingSource } from 'gql/gen/graphql';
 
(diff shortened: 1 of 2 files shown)


garyjzhao (approved):
LGTM

PR #916, opened 2024-01-23
Easi 3774/existing model link refactor

StevenWadeOddball:
# EASI-3774

### BE
- Add field name to Existing Model Link
   - update the update mutation to require field name
- Add parent level type ExistingModelLinks
   - add a resolver to return an array of names for a model plan and field name combination
      - implement this as a dataloader
- Move all current model plans to be in reference to the resemblesExistingModelWhich question for plan general characteristics
- Updated Unit Tests
- Updated Postman Collection

### FE

- Updated `GetAllGeneralCharacteristics.ts` , `GetGeneralCharacteristics.ts`, and `UpdateExistingModeLinks.ts` to have updated queries
- Updated ready only and task list general characteristic pages 

#### app code
1. Bring up the app
2. Use the updated postman collection to create a model plan
3. Use the collection to add links
4. Use the collection to query the names, and return the links
#### migration
1. Comment out the last migration, run this code to insert test data
``` sql
INSERT INTO "public"."user_account"("id", "username", "is_euaid", "common_name", "locale", "email", "given_name", "family_name", "zone_info", "has_logged_in") VALUES('0e892587-a1d2-44ef-8497-6091f6d1edf1', 'MINT', 'TRUE', 'MINT Doe', 'en_US', 'MINT@local.cms.gov', 'MINT', 'Doe', 'America/Los_Angeles', 'TRUE');


/* Make Model Plan */
INSERT INTO "public"."model_plan"("id", "model_name", "archived", "status", "created_dts", "created_by") VALUES('96148eb0-7768-467b-9253-114dc032d55d', 'Excellent Model', 'FALSE', 'PLAN_DRAFT', '2024-01-23 16:14:10.491352+00', '0e892587-a1d2-44ef-8497-6091f6d1edf1');
/* Make Link */
INSERT INTO "public"."existing_model_link"("id", "model_plan_id", "current_model_plan_id", "created_by", "created_dts") VALUES('96148eb0-7768-467b-9253-114dc032d55d', '96148eb0-7768-467b-9253-114dc032d55d', '96148eb0-7768-467b-9253-114dc032d55d', '0e892587-a1d2-44ef-8497-6091f6d1edf1', '2024-01-23 16:14:10.491352+00');

```
2. Run the new migration
3.  Ensure the link has the new correct field name value `GEN_CHAR_RESEMBLES_EXISTING_MODEL_WHICH`

### FE

1. Load up the app 
2. Work through the General Characteristics task list and make sure nothing breaks
3. Check Read Only General Characteristics as well

## PR Author Review Checklist

## PR Reviewer Guidelines


patrickseguraoddball (changes requested):
The `resemblesExistingModelWhich` field does not render in the CSV.  This is something we may want to sync on as the way `resemblesExistingModelWhich` introduces a new structure that currently isnt supported by the dynamic translation mapping.  I think we should implement this into our translation/type to be able to support nested fields.  Once this work is done we can render it in the CSV without hardcoding labels/values

`resemblesExistingModelWhich` doesn't exist in the translations and will probably need to replace `existingModelLinks` (Already somewhat of a unique case, but can replace/rework as a part of this)

@garyjzhao ping me when you have a chance and we can hop on a call to discuss a plan of action

patrickseguraoddball (changes requested):
I think we removed it, but `!allPlanTranslation?.[key]?.isModelLinks &&` is needed in the condition on line 162 of the `useFetchCSVData.tsx`

patrickseguraoddball (approved):
LG

pkg/models/model_plan.go
```diff
@@ -21,6 +21,7 @@ func NewModelPlan(createdBy uuid.UUID, modelName string) *ModelPlan {
 	}
 
 }
+func (m ModelPlan) isLinkedExistingModel() {}
```
OddTomBrooks, line 24: Was this method meant to be filled?

StevenWadeOddball: No, that is to represent the union type in GQL. It satisfies the interface

PR #920, opened 2024-01-26
[NOREF] Update GHA SHA targets

ClayBenson94:
# NOREF

## Changes and Description

Following docs from [here](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request), we can opt to use the merge commit for triggering workflows instead of the head commit.

Making this change means we're more easily able to catch issues that would result from a merge that Git thinks is clean/easy, but practically introduces errors (duplicate migration versions is a big one that Git doesn't catch)

## How to test this change

TODO: I'll link some workflow runs that target this branch from another PR!

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/.github/workflows/build_application_images.yml
+++ b/.github/workflows/build_application_images.yml
@@ -4,9 +4,9 @@ on:
   workflow_call:
 
 env:
-  GIT_HASH: ${{ github.event.pull_request.head.sha || github.sha }}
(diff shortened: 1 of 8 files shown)


Jdwoodson (approved):
LGTM

PR #922, opened 2024-01-27
EASI-3835: remove logstash/opensearch from local deployment

Jdwoodson:
# EASI-3835

## Changes and Description

- remove logstash/opensearch from local deployment via docker compose

## PR Author Review Checklist

## PR Reviewer Guidelines


ClayBenson94:
@Jdwoodson Converting this to draft. I think we should have the backend code removed on this as well. I'll work with the BE folks to get this done in this PR! (That way we can tag the commit in `main` before this goes in as the final "pre removal" commit)

.spelling
```diff
@@ -219,10 +219,6 @@ routes.go
 inlining
 Performant
 serializable
-OpenSearch
-Opensearch
```
ClayBenson94, line 223: Removing these lines is causing linters to fail due to these terms existing in ADRs. These shouldn't be removed for that reason!

ClayBenson94 (changes requested):
Also looks like the `pkg/querybuilderfilters` tests are failing due to missing references to auto-generated types no longer being present.

I think we should just delete the `querybuilderfactories`, `querybuilderfilters`, and `pkg/oddsearch` packages along with the other code here, since it all ties in with OpenSearch.

OddTomBrooks:
Since we have review on this I’m promoting this from draft and addressing feedback.

ClayBenson94 (approved):
Changes look good! @Jdwoodson, since you're the author, are you good to merge this and deploy it out?

PR #979, opened 2024-02-27
[Easi-4043] User notification preferences flag to array

StevenWadeOddball:
# EASI-4043

## Changes and Description

- Migrated user notification preferences to use an array instead of a single enum value
- introduced a serialization package to host helper methods to serialize this enum array (originally prototyped in `EASI-1982/generic_type_array_serialization`
   - We could have used a pq string array, but it seems important to be able to use receiver methods on this preference type. We will be using this value throughout the application.

## How to test this change
1. Use the updated postman collection to update user preferences.
2. Try to remove all preferences, and re-add preferences.
3. Verify the existing logic to not send an email if tagged in a discussion or a reply continues to function as expected.

Note: This code does not enforce that a preference not be duplicated in an entry, for example, IN_APP could be listed twice. The application will still handle this, considering the presence of the enum at least once to be enough to send a notification. 

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/MINT.postman_collection.json
+++ b/MINT.postman_collection.json
@@ -1,6 +1,6 @@
 {
 	"info": {
-		"_postman_id": "32329f70-986e-430d-8a0b-4509ca71dab1",
+		"_postman_id": "58f72d3a-ea92-4df4-8d3d-4627cbf7eb3b",
 		"name": "MINT",
 		"schema": "https://schema.getpostman.com/json/collection/v2.1.0/collection.json",
 		"_exporter_id": "20320964"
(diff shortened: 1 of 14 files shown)


patrickseguraoddball (dismissed):
LG!  Thanks Steven!

StevenWadeOddball:
Thanks @patrickseguraoddball ! This should be good to go, I had to change a unit test that I originally missed. I corrected it and expanded the test slightly now that we are using arrays.

patrickseguraoddball (approved):
Thanks!

PR #358, opened 2022-12-14
[EASI-2346] - Faktory Jobs

rileyanderson:
# EASI-2346

## Changes and Description

- Add faktory jobs for daily digest
 ```mermaid
graph TD;
    DailyDigestCronJob-->AnalyzedAuditBatchJob;
    AnalyzedAuditBatchJob-->AnalyzedAuditJob1;
    AnalyzedAuditBatchJob-->AnalyzedAuditJob2;
    AnalyzedAuditBatchJob-->AnalyzedAuditJob3;
    AnalyzedAuditBatchJob-->AnalyzedAuditJobX;
    AnalyzedAuditJob1-->AnalyzedAuditBatchJobSuccess;
    AnalyzedAuditJob2-->AnalyzedAuditBatchJobSuccess;
    AnalyzedAuditJob3-->AnalyzedAuditBatchJobSuccess;
    AnalyzedAuditJobX-->AnalyzedAuditBatchJobSuccess;
    AnalyzedAuditBatchJobSuccess-->DigestEmailBatchJob;
    DigestEmailBatchJob-->DigestEmailJob1;
    DigestEmailBatchJob-->DigestEmailJob2;
    DigestEmailBatchJob-->DigestEmailJob3;
    DigestEmailBatchJob-->DigestEmailJobX;
    DigestEmailJob1-->DigestEmailBatchJobSuccess;
    DigestEmailJob2-->DigestEmailBatchJobSuccess;
    DigestEmailJob3-->DigestEmailBatchJobSuccess;
    DigestEmailJobX-->DigestEmailBatchJobSuccess;
```
- Add `Worker` field to `Server`  struct so the worker can get setup in `routes` with config it needs there then start where the server starts. I was seeing some weird behavior starting a go routine in `routes.go`. Plus I think it makes more sense to pull it out and start it where we start the server.
- Minor updates to existing code.
- Renamed files in worker to try and keep it more consistent/cleaner. Basically one file per job or batch/job combination.

### Cron Config
- TOML config files for cron jobs are in  `/.faktory` folder. We may want to move the prod and other envs to the infra repo in the future. In deployed environments this file must be loaded into `/etc/faktory/conf.d` on startup. See: [Ent Cron](https://github.com/contribsys/faktory/wiki/Ent-Cron)

### Environment Variables
**NOTE:** these have been updated in all deployed environments with `FAKTORY_PROCESS_JOBS=false`
[...]

PR #989, opened 2024-03-05
Task/easi 3949 daily digest notification be

StevenWadeOddball:
# EASI-3949

## Changes and Description

- New Activity Meta Data type for digest complete
- Refactor Factory to move some resolver logic from worker to resolver
- Daily Digest Seed data
- New data loaders

### To see the new GraphQl Activity
1. Bring up the application
2. Seed the database
   a. Note, this will create two daily digest emails in mail catcher, and will create notifications for users  `MINT` and `BTAL`
3. Run the updated postman script to get UserNotifications for the current user
4. Assert that that new type is returned. 

### To validate the factory job works correctly
1. Assert all unit tests pass, and modifications didn't interrupt any intentions of the original tests.

2. Run Factory locally, and assert digest jobs complete as expected
3. Follow the instructions from the [original factory  job PR](https://github.com/CMSgov/mint-app/pull/358). 
   a. Note, seeding the database now also analyzes audits. If testing with seed data, make sure to delete the analyzed audits, activities, and notifications from the DB after seeding.
   b. Alternatively, just create some model plans and update the data manually.
   c. Alternatively, comment out the call to seeder.CreateAnalyzedAuditData() in `cmd/dbseed/main.go` before seeding.
   
4. Note, to actually get the job to run, you need to have audit change data from the day before. The simplest way to simulate that is just to update all the change data, and set the modified date to yesterday using the SQL script below
```sql
update audit.change 
SET modified_dts =  CURRENT_TIMESTAMP - INTERVAL '1 day';

```
I also recommend updating the cron.toml config for faktory as suggested in the original PR. `  schedule = "*/1 * * * *"` will make the job run every minute instead of every 5 minutes. 

## PR Author Review Checklist

## PR Reviewer Guidelines


garyjzhao:
I tested the GraphQL section and it looks good.

pkg/storage/truncate.go
```diff
@@ -49,14 +50,31 @@ func (s *Store) TruncateAllTablesDANGEROUS(logger *zap.Logger) error {
 
 func removeNonSystemAccounts(s *Store) error {
 
-	script := `DELETE FROM user_account
-    WHERE username NOT IN
-    `
+	// script := `DELETE FROM user_account
+	// WHERE username NOT IN
+	// `
```
OddTomBrooks, line 55: Is there a reason to keep this?

pkg/worker/worker.go
```diff
@@ -58,6 +58,15 @@ func (w *Worker) Work() {
...
 	mgr.Register("AggregatedDigestEmailJob", w.AggregatedDigestEmailJob)
 
+	/**********************
+	* //Future Enhancement
+	* Consider providing workers with dataloaders, and potentially a shared context. As these run separate go routines for each worker,
+	***********************
+	*dataLoaders := loaders.NewDataLoaders(w.Store)
+	*ctx := loaders.CTXWithLoaders(context.Background(), dataLoaders)
+	*err := mgr.RunWithContext(ctx)
+	******************************/
```
OddTomBrooks, line 68: This is a great suggestion, but I wonder if this is the best way to capture these future investments. I've seen many ancient future TODO's in code bases forgotten to the annals of time.

pkg/storage/loaders/analyzed_audit_loader.go
```diff
@@ -0,0 +1,113 @@
...
+
+	audits, err := storage.AnalyzedAuditGetByModelPlanIDsAndDateLoader(loaders.DataReader.Store, jsonParams)
+	if err != nil {
+		return []*dataloader.Result{{Data: nil, Error: err}}
+	}
+
+	//First Key is Date, second is modelPlanID
+	auditsByID := map[string]map[uuid.UUID]*models.AnalyzedAudit{}
+	//MAP by date first
+	// then MAP by model_plan_id
```
OddTomBrooks, line 40: I wonder if this can be clarified using type aliasing. This is purely optional, but I think clarifying maps like this (especially nested maps) in our codebase will help keep it maintainable.

Here is some [untested] pseudo-code to illustrate what I mean
`// Define type aliases to improve readability
type ModelPlanID uuid.UUID
type AuditDate string
type AnalyzedAuditMap map[ModelPlanID]*models.AnalyzedAudit

// AuditsByDateAndID provides a clear description of what the map contains:
// A mapping from dates to another map, which maps model plan IDs to analyzed audits.
type AuditsByDateAndID map[AuditDate]AnalyzedAuditMap`

StevenWadeOddball: Interesting idea! I tried it out a little in the codebase, and I decided not to implement it, simply because it requires a good deal of additional type casting. (Eg, you have to cast each `UUID` to a `ModelPlanID`. As such, I left it as the underlying types. 

I'm comfortable with this compromise as this map is scoped to the function, and the keys are documented in comments.

pkg/worker/worker.go
StevenWadeOddball: Thanks, it's true we could make a tech debt ticket for this. However, there is not a large benefit at this point of time. (I started to implement this here, but it is too much scope creep.)

Making this note seemed to a good middle ground to me. This way if there is another ticket where this task could be in scope, there is a note in the codebase. If we need to expand our use of factory in the future, we might consider it more.

pkg/storage/truncate.go
StevenWadeOddball: thanks! Removed.

OddTomBrooks (approved):

PR #990, opened 2024-03-06
[EASI-3944] Notifications - On Discussion Replied Notification

OddTomBrooks:
# EASI-3944

## Changes and Description

* Created new notification meta structure
* Updated and generated GQL
* Updated plan discussion routing to gate sending reply emails based on preferences
* Updated plan discussion to send reply notifications based on preferences

## How to test this change

* Start the application
* Using Postman:
  * Create a default model plan
  * Create a default discussion
  * Create a default discussion reply
  * Query notifications
* Assert postman results are as expected
* Assert unit tests pass as expected

## PR Author Review Checklist

## PR Reviewer Guidelines


pkg/notifications/new_discussion_replied_meta.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/google/uuid"
+
+	"github.com/cmsgov/mint-app/pkg/models"
+	"github.com/cmsgov/mint-app/pkg/sqlutils"
+)
+
+// ActivityNewDiscussionRepliedCreate creates an activity for when a Discussion is replied to.
+func ActivityNewDiscussionRepliedCreate(ctx context.Context, np sqlutils.NamedPreparer, actorID uuid.UUID, discussionID uuid.UUID, replyID uuid.UUID, discussionContent models.TaggedHTML, getPreferencesFunc GetUserNotificationPreferencesFunc) (*models.Activity, error) {
+
+	activity := models.NewNewDiscussionRepliedActivity(actorID, discussionID, replyID, discussionContent.RawContent.String())
```
StevenWadeOddball, line 15: I think this should actually be the discussion reply content, not the discussion content.

pkg/notifications/new_discussion_replied_meta.go
```diff
@@ -0,0 +1,23 @@
...
+
+	retActivity, actErr := activityCreate(ctx, np, activity)
+	if actErr != nil {
+		return nil, actErr
+	}
+
```
StevenWadeOddball, line 21: This method should also be responsible for creating the notification as well. I think it should go to the created user of the discussion.

pkg/notifications/new_discussion_replied_meta_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+import (
+	"context"
+
+	"github.com/google/uuid"
+
+	"github.com/cmsgov/mint-app/pkg/models"
+)
+
+func (suite *NotificationsSuite) TestActivityNewDiscussionRepliedCreate() {
```
StevenWadeOddball, line 11: This is a good start, you probably could do without any tags in this test, the other activity type really deals with the tagged notification. I'd expect assertions that
1. The user that made the discussion gets a notification for the reply, and then assert the content matches as expected.

pkg/models/new_discussion_replied_meta.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"database/sql/driver"
+	"encoding/json"
+	"errors"
+
+	"github.com/google/uuid"
+)
+
+// NewDiscussionRepliedActivityMeta represents the notification data that is relevant to a new Discussion Reply
+type NewDiscussionRepliedActivityMeta struct {
```
StevenWadeOddball, line 12: This looks pretty good! I'd do a final pass looking at [Figma](https://www.figma.com/file/8MfXzXzmKWv8vBjTasUIkD/MINT---Notifications?type=design&node-id=3472-13198&mode=design&t=q13knZsma5QK1gdV-0) and include any other fields that might be needed. At A glance, I think we might need model_plan_ID to be stored, as well as a resolver to modelPlan.

I added this onto Gary's [PR](https://github.com/CMSgov/mint-app/pull/988) for the reply notification type, I didn't realize it was needed, but it's important for rendering.

pkg/graph/resolvers/plan_discussion.go
```diff
@@ -454,63 +454,73 @@ func CreateDiscussionReply(
...
-				modelPlan,
-				replyUser,
-			)
-			if errReplyEmail != nil {
-				logger.Error("error sending tagged in plan discussion reply emails to tagged users and teams",
-					zap.String("discussionID", discussion.ID.String()),
-					zap.Error(errReplyEmail))
-			}
+		// Early return if
+		if !pref.NewDiscussionReply.SendEmail() {
```
StevenWadeOddball, line 470: This is going to make some issues, the tag emails have their own preferences, so this specific preference shouldn't block those emails. (The tag emails are based on the preference of each individual who is tagged, while is this should be based on the preference of the user who created the discussion.

pkg/graph/resolvers/plan_discussion.go
```diff
@@ -454,63 +454,73 @@ func CreateDiscussionReply(
...
-
-			replyUser := principal.Account()
-			commonName := replyUser.CommonName
 
-			if err != nil {
-				logger.Error("error sending discussion reply emails. Unable to retrieve modelPlan",
-					zap.String("replyID", reply.ID.String()),
-					zap.Error(err))
-			}
+		pref, err := UserNotificationPreferencesGetByUserID(ctx, principal.Account().ID)
```
StevenWadeOddball, line 464: I think this is getting preferences for the wrong user, it should be getting the preferences of the user who created the discussion, not the preferences of the user who made the reply. The discussion creator should get the notification

pkg/graph/resolvers/plan_discussion.go
```diff
@@ -454,63 +454,73 @@ func CreateDiscussionReply(
...
+		if err != nil {
+			return nil, fmt.Errorf("unable to get user notification preference, Notification not created %w", err)
+		}
```
StevenWadeOddball, line 467: Somewhere in this flow, the notification method needs to get called still.  `notifications.ActivityNewDiscussionRepliedCreate`

the actor should be the user who made the reply, but the notification should go to the user who created the discussion.

StevenWadeOddball (commented):
Thanks @OddTomBrooks, this looks like you are on the right track!

I added some more notes as you continue to finalize the implementation.

The other thing not called out is just updating postman as well.

StevenWadeOddball:
Review is in progress still. @OddTomBrooks, can you target the feature branch here please? This shouldn't merge to main until the frontend is built out.

pkg/graph/resolvers/plan_discussion.go
```diff
@@ -454,13 +454,24 @@ func CreateDiscussionReply(
 		if err != nil {
 			return reply, err
 		}
+
 		// Create Activity and notifications in the DB
-		_, notificationErr := notifications.ActivityTaggedInDiscussionReplyCreate(ctx, tx, principal.Account().ID, discussion.ModelPlanID, discussion.ID, reply.ID, reply.Content, loaders.UserNotificationPreferencesGetByUserID)
+		_, notificationErr := notifications.ActivityTaggedInDiscussionReplyCreate(ctx, tx, discussion.CreatedBy, discussion.ModelPlanID, discussion.ID, reply.ID, reply.Content, loaders.UserNotificationPreferencesGetByUserID)
```
StevenWadeOddball, line 459: This method call was updated incorrectly, `principal.Account().ID` is the actor, you switched it to `discussion.CreatedBy`. Can you please revert the change you made to this line?

pkg/graph/resolvers/activity.resolvers.go
```diff
@@ -60,5 +84,16 @@ func (r *Resolver) TaggedInPlanDiscussionActivityMeta() generated.TaggedInPlanDi
...
 type taggedInDiscussionReplyActivityMetaResolver struct{ *Resolver }
 type taggedInPlanDiscussionActivityMetaResolver struct{ *Resolver }
+
+// !!! WARNING !!!
+// The code below was going to be deleted when updating resolvers. It has been copied here so you have
+// one last chance to move it out of harms way if you want. There are two reasons this happens:
+//   - When renaming or deleting a resolver the old code will be put in here. You can safely delete
+//     it when you're done.
+//   - You have helper methods in this file. Move them out to keep these resolver files clean.
+func (r *newDiscussionRepliedActivityMetaResolver) ModelPlanID(ctx context.Context, obj *models.NewDiscussionRepliedActivityMeta) (uuid.UUID, error) {
```
StevenWadeOddball, line 97: Can you delete this please?

pkg/notifications/notifications_test.go
```diff
@@ -27,3 +29,14 @@ func TestNotificationsSuite(t *testing.T) {
 
 	suite.Run(t, rs)
 }
+
+func (suite *NotificationsSuite) deserializeActivityMetadata(testActivity *models.Activity) models.ActivityMetaData {
```
StevenWadeOddball, line 33: Optional: Could you add a comment to this test function? That would be helpful if the hope is to use this in other functions where the same logic is being tested.

StevenWadeOddball (changes requested):
Thanks @OddTomBrooks, this is almost there!

I functionally tested the change with postman. I confirmed that a notification is created as expected, and that the email respects the user preference to only send an email when it is activated.

OddTomBrooks:
> Thanks @OddTomBrooks, this is almost there!
> 
> I functionally tested the change with postman. I confirmed that a notification is created as expected, and that the email respects the user preference to only send an email when it is activated.

Thank you for the thorough review. All of your requests are in and ready for re-review, a request has also been sent.

StevenWadeOddball (approved):
Thanks for making the changes @OddTomBrooks, this looks good to go.

PR #1022, opened 2024-03-26
[EASI-4017] GQL Modularization - Miscellaneous

OddTomBrooks:
# EASI-4017

## Changes and Description

- A wide variety of types have been extracted from the main gql schema into corresponding modules
- A dummy endpoint was introduced to the main schema in order to define the Mutation type at a top-level
  - This endpoint is intentionally not implemented. As it should never be used, an error is expected.

## How to test this change

- Smoke test the application, particularly focusing on areas where modules have been refactored

## PR Author Review Checklist

## PR Reviewer Guidelines


pkg/graph/schema/schema.graphql
```diff
@@ -1,1243 +1,14 @@
...
-
 """
 Query definition for the schema
 """
 type Query {
-  planDocument(id: UUID!): PlanDocument!
-  @hasAnyRole(roles: [MINT_USER, MINT_MAC])
-  existingModelCollection: [ExistingModel!]!
-  @hasAnyRole(roles: [MINT_USER, MINT_MAC])
   searchOktaUsers(searchTerm: String!): [UserInfo!]!
```
StevenWadeOddball, line 5: This feels like it could / should live in the user_info.graphql file. Thoughts?

pkg/graph/schema/schema.graphql
```diff
@@ -1,1243 +1,14 @@
...
-  existingModelLink(id: UUID!): ExistingModelLink!
-  @hasAnyRole(roles:[MINT_USER, MINT_MAC])
-  mostRecentDiscussionRoleSelection: DiscussionRoleSelection
-  @hasAnyRole(roles: [MINT_USER, MINT_MAC])
 }
 
 """
 Mutations definition for the schema
 """
 type Mutation {
```
StevenWadeOddball, line 12: I tested this, the dummy mutation isn't necesary. This works, I tested it out locally. See this [GitHubIssue](https://github.com/99designs/gqlgen/issues/630#issuecomment-473735885)
```suggestion
type Mutation
```

This way, we don't have a mutation that panics as expected behavior

pkg/graph/schema/types/operational_solution.graphql
```diff
@@ -0,0 +1,104 @@
...
+
+extend type Query {
+  operationalSolutions(operationalNeedID: UUID!, includeNotNeeded: Boolean! = false): [OperationalSolution!]!
+  @hasAnyRole(roles: [MINT_USER, MINT_MAC])
+
+  operationalSolution(id: UUID!): OperationalSolution!
+  @hasAnyRole(roles: [MINT_USER, MINT_MAC])
+}
+
+extend type Mutation {
```
StevenWadeOddball, line 95: I only see mutations for subtasks here, but the add and update mutations for an operational solution are missing from this file.

Update, it looks like they are currently in the possible operational solution file.

StevenWadeOddball (changes requested):
Thanks Tom,

Looks good! Just a couple of changes requested. I smoke test the application and it functioned as expected.

StevenWadeOddball (approved):

PR #1030, opened 2024-04-02
[EASI-4159] Graphql codegen translation schema

patrickseguraoddball:
# EASI-4159

## Changes and Description

- Extracted base translations field properties and added them to schema for type generation
- Added translation schemas for each task list section
- Mapped generated base type with extra FE types
- Merged generated task list section schema with FE form translation schema
- Added `TranslationDataType` in favor of string literal
- Added lint ignore for schemas that are not used by BE

## How to test this change

There should be no UI changes.  App should behave as normal

## PR Author Review Checklist

## PR Reviewer Guidelines


pkg/graph/schema/types/translation.graphql
```diff
@@ -0,0 +1,62 @@
+"""
+Base typea that represents FE translation structure
```
StevenWadeOddball, line 2: Is this a block comment for the file?  that comment structure attaches as a description in the GQL Schema. 
(in this case to the TranslationDataType). Could you change the page comment to use `#` at the start of every line instead?

if you update the type comments to use `"""`, they will be exposed in the GQL introspection query

pkg/graph/schema/types/translation.graphql
```diff
@@ -0,0 +1,62 @@
...
+enum TranslationDataType {
+  STRING
+  NUMBER
+  BOOLEAN
+  DATE
+  ENUM
+  OBJECT
+}
+
+# Represents the FORM type of the translation field
```
StevenWadeOddball, line 16:
```suggest
` ` ` (replace without spaces)
 Represents the FORM type of the translation field
` ` `
```
image

pkg/graph/schema/types/translation.graphql
```diff
@@ -0,0 +1,62 @@
...
+  dbField: String!
+  label: String!
+  readonlyLabel: String
+  sublabel: String
+  multiSelectLabel: String
+  isArray: Boolean
+  dataType: TranslationDataType!
+  formType: TranslationFormType!
+  isOtherType: Boolean # Is a question a followup to another that doesn't designate it's own readonly question/line,
+  otherParentField: String # gql field name for the parent question for fields that represent Other, Please specify, etc.  Used in change history to render parent question for context
```
StevenWadeOddball, line 43: Should this say the db_field name for the parent question? Or is that just dependent on the data export?

Also, if you use the ` ` ` comment style on the line before this field, it is exported in the introspection query

StevenWadeOddball:
This looks great Patrick! I just added some notes for adding comments to the file, but doesn't change functionality. I think this will work great to run with!

pkg/graph/schema/types/translation.graphql
patrickseguraoddball: Gotcha, thanks I have no idea how comments work here.  I just copied from the other related file haha

pkg/graph/schema/types/translation.graphql
patrickseguraoddball: The export converts it to the db field.  In the frontend, it is a gqlfield

patrickseguraoddball:
> This looks great Patrick! I just added some notes for adding comments to the file, but doesn't change functionality. I think this will work great to run with!

Thanks I'll update these comments :)

pkg/graph/schema/types/translation.graphql
StevenWadeOddball: That works! I'm going to manually implement the type on the backend, so I can specify it's a db field name on the go type

patrickseguraoddball: I'm also open to changing it to the db field, as currently the FE isn't using this in any way yet.  If it's just the BE use we can maybe convert that data

StevenWadeOddball: Thanks Patrick! I'm good with trying any way to start it out. We could also just take out the GQL specific language, and just leave it to say field name. That way, it's true for FE and BE use cases?

PR #1039, opened 2024-04-05
[NOREF] Update translation schema - parent child relation

patrickseguraoddball:
# NOREF

## Changes and Description

- Added schema types to support parent/child question relations

## How to test this change

Ensure that the mapped types a reflected on the FE types

## PR Author Review Checklist

## PR Reviewer Guidelines


pkg/graph/schema/types/model_plan_translation.graphql
```diff
@@ -0,0 +1,13 @@
+
+"""
+Represents model plan base translation data
+"""
+type ModelPlanTranslation {
+  modelName: TranslationField! @goTag(key: "db", value: "model_name")
+  previousName: TranslationField! @goTag(key: "db", value: "previous_name")
```
StevenWadeOddball, line 7: previousName and nameHistory aren't actually fields in the database, they are GQL resolvers, I don't think it hurts to leave them though. I wonder if we should remove the @goTag directives for these fields to be clear about that?

pkg/graph/schema/types/translation.graphql
```diff
@@ -79,3 +79,109 @@ type TranslationFieldWithOptions {
...
+  """
+  Is a question a followup to another that doesn't designate it's own readonly question/line
+  """
+  isOtherType: Boolean
+  """
+  Field name for the parent question for fields that represent Other, Please specify, etc.  Used in change history to render parent question for context
+  """
+  otherParentField: String
+  options: Map!
+  parentRelation: Map!
```
StevenWadeOddball, line 156: Does this need to be a map because of the closure? Since it's 1:1, could it be the Field Name? 

We could maybe also expose the idea of an interface if it's 1:1, and is any type of translation field? I'm good to run with this for this PR and re-visit later!

StevenWadeOddball (approved):

pkg/graph/schema/types/model_plan_translation.graphql
patrickseguraoddball: Hmm yeah, does that affect being able to translate the data at all? If not, we can remove them

pkg/graph/schema/types/translation.graphql
patrickseguraoddball: The closure represents a Map basically.  If you look at the export it evaluates to the entire parent type.  May would be better to talk through

pkg/graph/schema/types/model_plan_translation.graphql
StevenWadeOddball: It wouldn't affect it in this instance, because we are programmatically accessing by DB fields. (And since the fields don't exist, we'd never hit those data points.

pkg/graph/schema/types/translation.graphql
StevenWadeOddball: Sounds good! I'd be happy to sync on this. I think that could potentially be ok to not know the exact parent type, just say it is a translation type (which we could define as a union in the schema). I made essentially this in the Backend, I just didn't expose the type in GQL previously.

pkg/graph/schema/types/translation.graphql
```diff
@@ -79,3 +79,111 @@ type TranslationFieldWithOptions {
...
+  """
+  Is a question a followup to another that doesn't designate it's own readonly question/line
+  """
+  isOtherType: Boolean
+  """
+  Field name for the parent question for fields that represent Other, Please specify, etc.  Used in change history to render parent question for context
+  """
+  otherParentField: String
+  options: Map!
+  childRelation: ChildTranslation!
```
StevenWadeOddball, line 132: For all of the child fields, doesn't this represent more than one entry? Parents are one parent per child, but Parents could have many children correct? This field might need to go back to being a map (or an array)

pkg/graph/schema/types/translation.graphql
```diff
@@ -79,3 +79,111 @@ type TranslationFieldWithOptions {
...
+  """
+  Is a question a followup to another that doesn't designate it's own readonly question/line
+  """
+  isOtherType: Boolean
+  """
+  Field name for the parent question for fields that represent Other, Please specify, etc.  Used in change history to render parent question for context
+  """
+  otherParentField: String
+  options: Map!
+  parentRelation: TranslationFieldWithOptionsAndChildren!
```
StevenWadeOddball, line 158: Should the parent field be a union type instead? Or if the child is a `TranslationFieldWithOptionsAndParent`, can you safely assume that the parent always is `TranslationFieldWithOptionsAndChildren`?

I had similar questions about the other parent implementations for the related questions.

StevenWadeOddball (commented):
Thanks Patrick,

This is looking good! For most of these concerns, I'm comfortable merging to feature and testing this out more with the backend and seeing how it functions. We can continue iterating as needed when integrating with the backend as well.

pkg/graph/schema/types/translation.graphql
patrickseguraoddball: I don't think so, but I'm not sure I fully understand your question.  I was shaping these based on the FE, and we don't really define any parent as a union.  They all have distinct types

PR #1053, opened 2024-04-16
[EASI-4169] Mark Primary Point-Of-Contact

OddTomBrooks:
# EASI-4169

## Changes and Description

- GQL
- SQL
- Migration
- Resolvers
- Store methods and corresponding SQL
- Unit Test
- Updated Postman endpoints
- Added primary contact resolver extension to possible operational solution\

## How to test this change

- Restart containers
- Check possible operational solution contacts
  - Assert assigned primary contacts correspond with frontend cards
- Postman endpoints
  - Get all possible operational solutions
    - Assert results contain primary contacts and that they are proper
  - Set a primary operational solution contact

## PR Author Review Checklist

## PR Reviewer Guidelines


ClayBenson94:
@OddTomBrooks Is there a reason to merge this into a `NOREF` instead of into `main`? If not, might make sense to just rebase onto `main` unless you've got a plan for something else!

migrations/V153__Add_Primary_Point_Of_Contact_Configuration.sql
```diff
@@ -0,0 +1,54 @@
+ALTER TABLE possible_operational_solution_contact
+  ADD COLUMN is_primary BOOLEAN DEFAULT NULL;
```
ClayBenson94, line 2: I think it might be good to make this `NOT NULL` and just have it be explicitly `true` or `false`, since I don't think we have a reason to differentiate `NULL` and `false`.

pkg/graph/schema/types/possible_operational_solution_contact.graphql
```diff
@@ -9,6 +9,7 @@ type PossibleOperationalSolutionContact {
   email: String!
   isTeam: Boolean!
   role: String
+  isPrimary: Boolean
```
ClayBenson94, line 12: Per my previous comment (in the migration), I think we can make this not nullable

```suggestion
  isPrimary: Boolean!
```

pkg/graph/schema/types/possible_operational_solution.graphql
```diff
@@ -4,6 +4,7 @@ type PossibleOperationalSolution {
   key: OperationalSolutionKey!
   treatAsOther: Boolean!
   pointsOfContact: [PossibleOperationalSolutionContact!]!
+  primaryContact: PossibleOperationalSolutionContact
```
ClayBenson94, line 7: I'd say we should either:
- Ensure this resolver uses a dataloader
- Remove this resolver and leave it up to the frontend to query `pointsOfContact` where `isPrimary: true` to fetch primaries

This will currently fire off a SQL query to fetch the primary contact for each (less than ideal, but not a dealbreaker I'd say)

pkg/models/possible_operational_solution_contact.go
```diff
@@ -12,10 +12,11 @@ type PossibleOperationalSolutionContact struct {
...
 
-	Name   string  `db:"name" json:"name"`
-	Email  string  `db:"email" json:"email"`
-	IsTeam bool    `db:"isTeam" json:"is_team"`
-	Role   *string `db:"role" json:"role"`
+	Name      string  `db:"name" json:"name"`
+	Email     string  `db:"email" json:"email"`
+	IsTeam    bool    `db:"isTeam" json:"is_team"`
+	Role      *string `db:"role" json:"role"`
+	IsPrimary *bool   `db:"is_primary" json:"isPrimary"`
```
ClayBenson94, line 19: (Following other comments, this can probably change types!)
```suggestion
	IsPrimary bool   `db:"is_primary" json:"isPrimary"`
```

migrations/V153__Add_Primary_Point_Of_Contact_Configuration.sql
```diff
@@ -0,0 +1,54 @@
+ALTER TABLE possible_operational_solution_contact
+  ADD COLUMN is_primary BOOLEAN DEFAULT NULL;
+
+WITH PrimaryContacts AS (
+  SELECT poc.id AS contact_id
+  FROM possible_operational_solution_contact poc
+         JOIN possible_operational_solution pos ON poc.possible_operational_solution_id = pos.id
+  WHERE (pos.sol_key, poc.name) IN (
```
ClayBenson94, line 8: Does this migration aim to add a primary for each solution? It looks like they're all covered here, which is great news!

On that note: I'd initially written in the ticket that a possible operational solution may have 0 primaries, but if this migration successfully covers all our bases, it might make sense to force a constraint to have `exactly 1` primary per solution ID! It would not only help validate that this migration sets exactly 1 per solution, but also future migrations won't be able to set an invalid state where we have no primary POCs!

ClayBenson94 (changes requested):
A few minor suggestions, and one larger change that we could make now that I'm reviewing a bit deeper!

pkg/graph/schema/types/possible_operational_solution.graphql
OddTomBrooks: The general desired paradigm we've been migrating towards is to include resolvers in this fashion. I considered this as being a dataloader, but it seems the efficiency gains for a dataloader are more about batch loading data. Wouldn't we need to fire a SQL query regardless of which solution we use?

Another consideration, if the frontend doesn't need the primaryContact field can't they simply omit it from the GQL query?

OddTomBrooks:
> @OddTomBrooks Is there a reason to merge this into a `NOREF` instead of into `main`? If not, might make sense to just rebase onto `main` unless you've got a plan for something else!

I'm not sure what the FE implementation considerations are. I don't see a problem with rebasing to main, just being considerate in case FE has needs before it hits main.

migrations/V153__Add_Primary_Point_Of_Contact_Configuration.sql
OddTomBrooks: The goal was to achieve POC parity with FE primary contact displays. At inception we weren't sure if this actually had complete coverage, but ultimately it seems to have worked out. I lost a bit of work, previously I achieved a similar constraint using a partial index. If it's the same for you I'll reintroduce this logic as it seems the appropriate mechanism.

pkg/graph/schema/types/possible_operational_solution.graphql
ClayBenson94: I'm fine to leave it, and the frontend can use whichever option makes the most sense for them! In any case, the query made is generally simple & quick!

migrations/V153__Add_Primary_Point_Of_Contact_Configuration.sql
ClayBenson94: Yeah, I think that'd be good! Not sure about how partial indexes work vs. just table/row constraints, but I think something that forces exactly 1 `true` per solution ID would be good here!

ClayBenson94:
> > @OddTomBrooks Is there a reason to merge this into a `NOREF` instead of into `main`? If not, might make sense to just rebase onto `main` unless you've got a plan for something else!
> 
> I'm not sure what the FE implementation considerations are. I don't see a problem with rebasing to main, just being considerate in case FE has needs before it hits main.

On that note, I might loop in @garyjzhao to do some small FE changes on this, since it does seem (in its current state) to change the order of returns, though I expect the FE changes to be minor!

patrickseguraoddball:
Working on these failing unit tests

patrickseguraoddball:
Verified that this branch's poc's line up with what's in dev

patrickseguraoddball (approved):
LG

PR #1101, opened 2024-05-06
[EASI-4238] Added latestModifiedDts field to operational need resolver

OddTomBrooks:
# EASI-4238

## Changes and Description

- Added lastModifiedDts for tracking usage which maps against various tables' change

## PR Author Review Checklist

## PR Reviewer Guidelines


StevenWadeOddball (changes requested):
Thanks @OddTomBrooks,

I haven't pulled this to test locally yet. I have concerns about the architecture.  Each model plan has at least 32 operational needs. Effectively, this means anytime this is endpoint is called, it will reach out to the database at least 32 times. If it is called for multiple model plans, the number of db queries grows  (n+1).  

I think a `dataloader` is warranted for this use case to ensure their is only one query to the database.

 Also, I wanted to check is there a reason this is on the operational need and not on the model plan itself? From the JIRA ticket, it looks like the ask is for the most recent date for the entire operational solution and implementation section, not an individual operational need. That could accomplished similarly with a query that takes a model plan id. A `dataloader` would still be warranted with this approach, but it would already cut down from >= 32 queries to 1 (for each model plan).

OddTomBrooks:
> Thanks @OddTomBrooks,
> 
> I haven't pulled this to test locally yet. I have concerns about the architecture. Each model plan has at least 32 operational needs. Effectively, this means anytime this is endpoint is called, it will reach out to the database at least 32 times. If it is called for multiple model plans, the number of db queries grows (n+1).
> 
> I think a `dataloader` is warranted for this use case to ensure their is only one query to the database.
> 
> Also, I wanted to check is there a reason this is on the operational need and not on the model plan itself? From the JIRA ticket, it looks like the ask is for the most recent date for the entire operational solution and implementation section, not an individual operational need. That could accomplished similarly with a query that takes a model plan id. A `dataloader` would still be warranted with this approach, but it would already cut down from >= 32 queries to 1 (for each model plan).

This was added to the operational needs resolver as from the [Slack conversation](https://oddball.slack.com/archives/C059QQA3HK5/p1713898541209749) this seems to be the requested structure. A significant degree of complexity can be resolved by simply omitting the lastModifiedDts field from queries except where necessary, which to my understanding of GraphQL will occlude the unnecessary field resolution.

I will investigate performance and architectural considerations.

StevenWadeOddball:
> This was added to the operational needs resolver as from the [Slack conversation](https://oddball.slack.com/archives/C059QQA3HK5/p1713898541209749) this seems to be the requested structure. A significant degree of complexity can be resolved by simply omitting the lastModifiedDts field from queries except where necessary, which to my understanding of GraphQL will occlude the unnecessary field resolution.
> 
> I will investigate performance and architectural considerations.

Thanks @OddTomBrooks for taking a look. My take from that slack conversation was that it could be on the model plan or the need, so I wanted to see if you considered the model_plan_id approach. 

And that's definitely true that the query will only be called when the field is requested. However, we can't control when this is called. Someone could query a model plan collection, and get the needs for every model plan.

ClayBenson94:
I think in any case having this property on the model plan / task list itself (rather than on each need) means the FE won't have to query many needs and find the largest, but rather have the backend make a single query across the entire model plan and return a single date. I think that makes things a bit less complex for both the backend and the frontend!

pkg/storage/model_planStore.go
```diff
@@ -69,6 +73,48 @@ func (s *Store) ModelPlanGetByModelPlanIDLOADER(_ *zap.Logger, paramTableJSON st
 	return planSlice, nil
 }
 
+// ModelPlanTrackingDateGetByModelPlanIDLOADER returns the most recent tracking dates for a series of model plans
+func (s *Store) ModelPlanTrackingDateGetByModelPlanIDLOADER(logger *zap.Logger, paramTableJSON string) (
```
StevenWadeOddball, line 77: maybe this name should reference the task list section name?

pkg/storage/model_planStore.go
```diff
@@ -69,6 +73,48 @@ func (s *Store) ModelPlanGetByModelPlanIDLOADER(_ *zap.Logger, paramTableJSON st
...
+		logger.Error("Failed to prepare SQL statement", zap.Error(err))
+		return nil, err
+	}
+	defer stmt.Close()
+
+	arg := map[string]interface{}{
+		"paramTableJSON": paramTableJSON,
+	}
+
+	rows, err := stmt.Queryx(arg)
```
StevenWadeOddball, line 94: Nice, I've not seen queryx used before. Is this so you only iterate through results one? You could also declare a type in this function, and use select.

StevenWadeOddball (commented):
Thanks for implementing a dataloader Tom, I didn't have time to finish the review today, but I will look first thing on Monday morning.  This looks pretty good at a high level! I'm going to continue Monday with testing, and some attention to the dataloader batch function. commenting now so you can see the review in progress.

pkg/storage/model_planStore.go
OddTomBrooks: I made a few attempts at that to keep the code style homogenous, reasons I ultimately went with this:

1. Deserialization was failing and creating an empty / nil array, time value for figuring out that issue wasn't worth it
2. Creating a two-field model which is unexposed felt very edge-case to me. It seems to me like manual iteration to build an associative container in this case is at least as appropriate if not slightly more so. Maybe a similar argument is that sometimes Tuples are more appropriate than rolling out new objects for utilitarian tasks.
3. I suspect it's a bit more efficient since the only unmarshalling is to direct known data types rather than any sort of tag-mapping. Not that I think this is really a significant concern either way.

pkg/graph/schema/types/model_plan.graphql
```diff
@@ -68,6 +68,7 @@ type ModelPlan {
   prepareForClearance: PrepareForClearance!
   nameHistory(sort: SortDirection! = DESC): [String!]!
   operationalNeeds: [OperationalNeed!]!
+  lastModifiedDts: Time!
```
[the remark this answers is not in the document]
OddTomBrooks: Maybe. I have been over this name a few times. Ultimately I kept it simple and expected some discussion. Maybe a healthy middle-ground would be `opSolutionLastModifiedDts`?

pkg/storage/model_planStore.go
OddTomBrooks: I'll focus discussion around this on the graphql comment and update this method's name to correspond.

pkg/graph/schema/types/model_plan.graphql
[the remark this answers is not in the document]
StevenWadeOddball: Yeah! I think that would be pretty good. Thanks Tom!

pkg/storage/SQL/model_plan/get_tracking_date_by_id_LOADER.sql
```diff
@@ -0,0 +1,63 @@
...
+    FROM plan_document_solution_link
+    INNER JOIN operational_solution ON operational_solution.id = plan_document_solution_link.solution_id
+    INNER JOIN operational_need ON operational_need.id = operational_solution.operational_need_id
+    WHERE operational_need.model_plan_id IN (SELECT id FROM QUERIED_IDS)
+    GROUP BY operational_need.model_plan_id
+
+    UNION ALL
+
+    SELECT
+        plan_document.model_plan_id,
```
StevenWadeOddball, line 52: I think this is almost correct, but not quite. 

I think we only care about document dates that correspond to a linked document. Maybe it would make sense to add to the select above it instead? (Just add one more join to the document from the document solution link, and find the greatest link or linked document date?)

pkg/storage/loaders/model_plan_tracking_date_loader.go
```diff
@@ -0,0 +1,50 @@
...
+		logger.Error("issue converting keys for data loader in Model Plan Tracking Date", zap.Error(*err))
+		return nil
+	}
+	marshaledParams, err := arrayCK.ToJSONArray()
+	if err != nil {
+		logger.Error("issue converting keys to JSON for data loader in Model Plan Tracking Date", zap.Error(*err))
+		return nil
+	}
+
+	trackingDates, _ := dr.Store.ModelPlanTrackingDateGetByModelPlanIDLOADER(logger, marshaledParams)
```
StevenWadeOddball, line 29: Could you handle this error here please?

pkg/graph/resolvers/model_plan.resolvers.go
```diff
@@ -113,6 +114,11 @@ func (r *modelPlanResolver) OperationalNeeds(ctx context.Context, obj *models.Mo
 	return OperationalNeedCollectionGetByModelPlanIDLOADER(ctx, obj.ID)
 }
 
+// LastModifiedDts is the resolver for the lastModifiedDts field.
+func (r *modelPlanResolver) LastModifiedDts(ctx context.Context, obj *models.ModelPlan) (*time.Time, error) {
```
StevenWadeOddball, line 118: Can you please make some tests for this? Since the logic lives in SQL, it would be good to have some assertions that it returns the dates as expected.

pkg/storage/loaders/model_plan_tracking_date_loader.go
StevenWadeOddball, line 1: Overall, this looks really good! Thanks Tom.

I've learned a couple of things recently that have come up in other applications (we haven't necessarily seen the issue in MINT yet). That being said, there remains some tech debt to be refactored in other data loaders.

1. A Dataloader needs an error for each key provided. If not, the data loader will return another irrelevant error
`The batch function supplied did not return an array of responses`. Effectively, this means iterating over the length of the key array, and making an error for each key. Can you do that where you are returning nil please? (The returned store error would be appropriate as well)

pkg/storage/loaders/model_plan_tracking_date_loader.go
```diff
@@ -0,0 +1,50 @@
...
+	"fmt"
+
+	"github.com/graph-gophers/dataloader"
+	"go.uber.org/zap"
+
+	"github.com/cmsgov/mint-app/pkg/appcontext"
+)
+
+// GetModelPlanTrackingDateByModelPlanID uses a DataLoader to aggregate a SQL call and return the Model Plan Tracking Date in one query
+func (loaders *DataLoaders) GetModelPlanTrackingDateByModelPlanID(ctx context.Context, keys dataloader.Keys) []*dataloader.Result {
```
StevenWadeOddball, line 14: I tried a different paradigm for the notification preferences loader. (`pkg/storage/loaders/user_notification_preferences_loader.go`). 

Effectively, I don't export the batch function, and I also create a single exported method which handles all the logic for loading the data loader (what you now have living in resolvers. I like that it keeps all responsibility for data loaders in a single place. What do you think? This comment is non-blocking, but if you agree with the paradigm, feel free to update as well!

pkg/graph/resolvers/model_plan.go
```diff
@@ -247,6 +248,24 @@ func ModelPlanGetByIDLOADER(ctx context.Context, id uuid.UUID) (*models.ModelPla
...
+	key.Args["id"] = id
+
+	thunk := planTrackingDateLoader.Loader.Load(ctx, key)
+	result, err := thunk()
+
+	if err != nil {
+		return nil, err
+	}
+
+	return result.(*time.Time), nil
```
StevenWadeOddball, line 266: We do this in a couple places, but this I've recently seen that this can make a panic. Could you please do the type assertion a line before to validate that the result before returning?

Something like
```suggestion
	timeRes, isTime := result.(*time.Time)
	if !isTime {
		return nil, fmt.Errorf("the result was not *time.Time as expected for the solution section tracking date, it was %T", result)
	}
	return timeRes, nil
```

pkg/storage/model_planStore.go
StevenWadeOddball: Thanks Tom! That makes sense to me and this seems like a great approach! Thanks for explaining your approach here

StevenWadeOddball (changes requested):
Thanks Tom, overall this looks pretty great!

The data loader is returning results as expected. I have one concern about it not returning the appropriate time for documents.

Other feedback is about testing, and some error handling for dataloaders.

I tested the current functionality and it works as expected (except for the document case mentioned above).

pkg/storage/loaders/model_plan_tracking_date_loader.go
OddTomBrooks: Yes. It's worth considering that we currently do not capture or handle the error from model_plan_loader.go:29:

	`plans, _ := dr.Store.ModelPlanGetByModelPlanIDLOADER(logger, marshaledParams)`
	
We may benefit from eventual static analysis to catch other such unhandled error situations.

I previously added error handling, but removed it due a capturing error. It seems for some reason we pass error pointers in prior utility methods, which is not idiomatic to Go. Part of tech investment should likely be to restructure these methods to return a simple error struct. I will work around this in the moment by creating an `err2` value.

StevenWadeOddball: I think you are right, I forget the reason those helper methods returned a pointer, but I do agree it probably should just be the value instead of a pointer.

StevenWadeOddball (approved):
🚀 Thanks @OddTomBrooks! The changes look good to go

PR #1105, opened 2024-05-07
[EASI-2857] Cypress Test Updates and Upgrades

garyjzhao:
# EASI-2857

## Changes and Description

- Updated cypress to `13.9.0`
- Add Beneficiaries cypress test
- Addressed flakeiness in Notification test
- Changed cypress's `global-hook.js` from `beforeEach` to `before`
  - `before()` runs once before all your code.
  - `beforeEach()` runs before each of your code blocks.
- Streamlined the Notification by combining assertions and etc

## How to test this change

1. `yarn install`
2. run cypress test to ensure no failures
3. 🎉 

## PR Author Review Checklist

## PR Reviewer Guidelines


patrickseguraoddball (changes requested):
LG Gary!  I wanted to request changes just to improve on the performance of the notifications test.  There seems to be a lot of unncessary cleaning and seeding of the db (which takes a lot of time).  I think we can run this on a single set of data (no need to clean and reseed).  It might need a few adjustments (cy.logout before each re login, etc).  I think it would increase the performance a lot and make more sense with the same set of data

patrickseguraoddball (approved):
Nice work with the improvements and session implementation

PR #1106, opened 2024-05-07
[EASI-4279] - New record types

patrickseguraoddball:
# EASI-4279

## Changes and Description

- Added support for date render
- Added support for follow up parent question or note -  (`referenceLabel`)
- Added UUID type
- Updated unit tests

## How to test this change

Verify that date, other types, and notes render according to figma

## PR Author Review Checklist

## PR Reviewer Guidelines


StevenWadeOddball (approved):
Thanks Patrick, this will be good!

I think the table_reference will be good for foreign keys. I wonder if we will want to type that as an enum? (if there already is a table name enum, we could potentially use that instead). 

Also, I think the translation field  graphql type needs to be updated for the new tableReference field. I can do that when I work on the backend piece of this though, as I will have to update the backend translation type as well.)

PR #1115, opened 2024-05-14
[Easi 4345] Generic translated audit meta data, discussion reply meta data

StevenWadeOddball:
# EASI-4345

## Changes and Description

- Added a generic audit meta data type 
   -  All meta data to be null in the DB
   - Add a meta data type to the table
- Fill out the meta data for a discussion reply

NOTE: This meta data work will continue for additional types. PR is being created now to unblock front end work. 

## How to test this change
1. Make a discussion
2. Make a reply
3. Translate the audits
4. Verify the data is as expected.

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/migrations/V155__Translated_Audit.sql
+++ b/migrations/V155__Translated_Audit.sql
@@ -24,14 +24,24 @@ CREATE TABLE translated_audit (
     primary_key UUID NOT NULL,
     action DATABASE_OPERATION NOT NULL, 
 
-    meta_data_type TRANSLATED_AUDIT_META_DATA_TYPE NOT NULL, -- This could be whatever
-    meta_data JSONB NOT NULL, -- This could be whatever
+    meta_data_type TRANSLATED_AUDIT_META_DATA_TYPE,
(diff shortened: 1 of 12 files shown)


patrickseguraoddball (commented):
I think the `relation` in it's current state doesn't give us anything that `tableName` doesn't give us?  Not sure, but it doesn't seem like its needed at this point.  Also if we do need more specificity that `tableName` then we can probably make `relation` an enum for the type of relation.  Sorry if this doesn't make sense, maybe we can sync/async on it

patrickseguraoddball:
Also do with think `relationContent` will always be a part of the metadata?  Can it be a part of the base type/don't need the additional union type?

patrickseguraoddball:
Doesn't block at all, thanks for the feedback!  Yep good to go with this.

patrickseguraoddball (approved):

StevenWadeOddball:
Thanks Patrick! I'll plan on doing some refactoring on this coming up and merge it for now!

PR #1116, opened 2024-05-15
[EASI-4333] Discussion, Documents, CRTDL Change history

patrickseguraoddball:
# EASI-4333

## Changes and Description

- Added support for Discussion, Documents, CRTDL
  - Some outstanding work on CRTDL and Discussions pending metadata updates
- Fixed capitalization on status (only actually rendered in CH)
- Updated query for metadata updates
  - Added typeguard to identify data type 
- Added more comments

## Notes

- Documents WILL NOT match Figma, and instead will mimic CRTDL's (had conversation with Natasha, but will still need to verify these changes once she returns)
- No meta data for discussion reply count
- No meta data for CRTDL id_number on update

## How to test this change

- For each audit type, create/upload, reply, updated, delete
- Verify matches Figma
  - See notes above

## PR Author Review Checklist

## PR Reviewer Guidelines


StevenWadeOddball (approved):
Thanks Patrick, 

This looks good! I think this is about as close at it can be until some more BE changes are implemented.

I think we might need to update the translation for contains sensitive cost information. It looks like it is rendered in change history as a switch on visibility (Restricted or All)
https://www.figma.com/design/5Ux715qVJHg38EbJKavMR4/MINT---Change-History?node-id=570-129695&t=AM2xNwi1eCVCw5FP-0
image

PR #1117, opened 2024-05-15
[EASI-4302, EASI-4300] Sticky Filter Bar and Scroll to Top on Read Only Pages

garyjzhao:
# EASI-4302, EASI-4300

## Changes and Description

- Adjusted css to make sure sticky filter bar does not hide things behind it
- Created new `scrollToTheFilterBanner` function and passed it down to `SideNav` component
- Add `height-viewpoint` to each of the read only `PageLoading` components to address scroll jitter

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/src/components/shared/SectionWrapper/index.tsx
+++ b/src/components/shared/SectionWrapper/index.tsx
@@ -9,6 +9,7 @@ type SectionWrapperProps = {
   border?: boolean;
   borderBottom?: boolean;
   borderTop?: boolean;
+  id?: string;
 };
 
 const SectionWrapper = ({
@@ -16,7 +17,8 @@ const SectionWrapper = ({
   children,
   border,
   borderBottom,
-  borderTop
+  borderTop,
+  id
 }: SectionWrapperProps) => {
   const classNames = classnames(
     'mint-section',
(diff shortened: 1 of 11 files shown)


patrickseguraoddball (changes requested):
The scroll to the top does not seem to be working as expected.  When scrolled all the way to the bottom of Payments, then click Ops Eval and Learning, the page does not scroll to the top. 

Also was the intent of the behavior to scroll to the bottom of the banner or the top of the page?  My instinct would be to the top of the page, but when I click a section (not scrolled to the bottom) it scroll down to the bottom of the banner.  I can sync and demo this behavior if needed!

garyjzhao:
Re: the scrolling

I brought up the functionality with Zoe, since Natasha is OOO, and she said that it would go right below the summary box. 

Re: function being passed down as a prop

Good call, I think I was trying something that never panned out and hadnt revert the work

patrickseguraoddball (commented):
This is kinda a nitpick bit it's noticeable to me.  When clicking a section after scrolling down, there is a slight discrepancy with the side nav not being completely at the top.  It still scrolls for about 20 pixels after it snaps to the top.  I can demo this if you want, kinda hard to explain.  We can also maybe run it by UX to see if it's an issue

PR #1120, opened 2024-05-16
[EASI-4294] Subtasks heading, column order, and Op Solutions in unfiltered PDF export

garyjzhao:
# EASI-4294

## Changes and Description

- Updated translation file `Subtask` -> `Subtasks`
- Updated `ReadOnlyComponents` function to have optional arg
- Updated the `hiddenColumns` array to not exclude subtasks
- Removed repeated snapshot test in the`SharedExport` 

## PR Author Review Checklist

## PR Reviewer Guidelines


src/views/ModelPlan/ReadOnly/OperationalNeeds/index.tsx
```diff
@@ -51,7 +57,7 @@ const ReadOnlyOperationalNeeds = ({ modelID }: { modelID: string }) => {
         modelID={modelID}
         type="needs"
         readOnly
-        hiddenColumns={['Subtasks']}
+        isExportingPDF={isExportingPDF}
```
patrickseguraoddball, line 60: There is a little prop drilling here that could maybe leverage a context we have for exporting pdf.  You can get the state of the export from this code.  Not huge deal just a suggestion.  I think there should already be the context wrapper - PrintPDFWrapper - around all of readonly

`const { isPrintPDF } = useContext(PrintPDFContext);`

garyjzhao: Great suggestion! I didn't know we had something like that already existing.

patrickseguraoddball (approved):
LG

PR #1122, opened 2024-05-17
[Easi 4356] CH ignore foreign key on audit, and only translate fields with translations

StevenWadeOddball:
# EASI-4356

## Changes and Description

- Removed aggregate model plan translation functionality (since this isn't how the job is working)
- Translate Field returns nil if a translation wasn't present. This field won't get saved to the database as a translated field
    - this will still be retained in the raw audit
- Foreign Key field is explicitly ignored on insert or delete.

We will hold off on review of this for the weekend. We are deciding if we should write all translated audits, but make it filterable. If so, some of these changes will be reverted

- If we do that, we will want to add another field to show if a field was actually translated or not, or if it was transferred as the raw value.
## PR Author Review Checklist

## PR Reviewer Guidelines


patrickseguraoddball (approved):
Works as expected, nice!  I think we still need to audit what is now not being translated to ensure it's correct.  I'll create a ticket to make sure we have translations for things we normally dont need translated on the FE

StevenWadeOddball:
> Works as expected, nice! I think we still need to audit what is now not being translated to ensure it's correct. I'll create a ticket to make sure we have translations for things we normally dont need translated on the FE

Thanks Patrick! There is also this ticket in the backlog to add a unit test to make sure no field get forgotten [EASI-4354](https://jiraent.cms.gov/browse/EASI-4354)

PR #1126, opened 2024-05-22
[EASI-4334] Linking table translations

StevenWadeOddball:
# EASI-4334

## Changes and Description

- Add change type Unchanged
    - addresses an edge case when a value changes from null to empty array.
    - the change type is categorized as unchanged, and the field is skipped when writing to the db.
- Update audit config to allow explicitly inserting a relational field even if it is already captured a primary or foreign key
   - updated document solution link to explicitly insert solution id.
- Translations for Document Solution Link
  - foreign key translations
- Translations for Existing Model Link table
  - foreign key translations 

## How to test this change
1. Make document solution links
2. Make Existing Model Links
3. Translate the tables.
4. Ensure they look correct.

## PR Author Review Checklist

## PR Reviewer Guidelines


patrickseguraoddball (approved):
Works as expected.  It looks like I have all the info needed to finish these off, nice!

PR #1127, opened 2024-05-22
[EASI-4359] - Pagination and op needs

patrickseguraoddball:
# EASI-4359

## Changes and Description

This work covers the addition of pagination to CH, as well as temporary renders for custom need create/update, solution create/update, and subtasks, and  document link (not yet implemented on BE).  

All of the audit changes for these types are tentative, but wanted to get in the base of this work before continuing

- Pagination
- Added CH link on readview
  - Added router state to manage text and back links
- Audit renders for custom need create/update, solution create/update, and subtasks, and  document link

## How to test this change

- Seed DB
- Verify pagination works
- Verify documents and subtasks behave as CRTDL does
- Verify custom need and solutions render with collapsible data beneath
  - Custom need create will collapse, solution added will not - only update

## PR Author Review Checklist

## PR Reviewer Guidelines


StevenWadeOddball (approved):
Thanks Patrick! This looks good! Some non-blocking observations below.

I was wondering about adding custom solutions. Should more information be shown? The audit is translating point of contact information as well when created, but it isn't rendered. 

I was also wondering for subtask, should the subtask name be displayed in the fields as well as the parent card? I think it only shows status in FIGMA.
image
image.

I also realized, subtasks will need meta data added (at least for update) to show the subtask title. (but that doesn't affect this PR)

I also noticed an issue when updating a custom solution. The nameOther field should populate in this form, but I think it is trying to use the name field. 
image

Because of that, I got an error when trying to update the custom solution. I expect this wasn't caused in this branch, but I wanted to call it out just in case!

PR #1129, opened 2024-05-23
[EASI-EASI-4312] Search, sort, date grouping

patrickseguraoddball:
# EASI-4312, EASI-4313, EASI-4315

## Changes and Description

Search is currently set to search on translated `tableName`, `actor`, `date` (formatted as it renders), and each of the translated field - `fieldNameTranslated`, `newTranslated`, `oldTranslated`, `referenceLabel`

- Added search component
- Added sort
- Added date grouping
- Added util and test for logic
- Added metadata to query for discussion reply
- Added comments here and there
- Added solution subtasks as a collapsible audit

Search, sort, and date grouping, and pagination should all work in conjunction with each other.  Ex: entering a query does not reset sort, they are cummulative - except for pagination, will be responsive on any other changes

## How to test this change

Test out search, sort and date grouping
Verify works as expected

## PR Author Review Checklist

## PR Reviewer Guidelines


patrickseguraoddball:
> Thanks @patrickseguraoddball , this looks good.
> 
> Would it be a big lift to have sort and search persist on page reload? If it's not very difficult, that could be a nice addition! I confirmed that sort and search and pagination don't interfere with each other at all.
> 
> The only thing I noticed is that the discussion reply isn't rendering the Discussion content like I'd expect from FIGMA. image

Thanks! yeah I've addressed the repy, as the metadata had gotten changed.  It's in my current branch.  As far as persisting all the search sort features, that wasn't in scope, and can probably use it's own ticket/not small.  I had thought about it maybe v2 or if I have time

PR #1135, opened 2024-05-28
[Easi 4354]/ch no missing field unit test

StevenWadeOddball:
# EASI-4354

## Changes and Description

- Add a new unit test utility function `assertTranslationStructCoverage` to iterate through all fields on a struct, and assert that a translation exists for all fields except excluded
- Call the new unit test for each translation

## How to test this change
1. Look through all the changes
2. Verify the newly added translations
3. Verify the fields being excluded make sense to be excluded

If you want to test the behavior that will happen for a new field being added, add a new field to a struct (like model plan). Run the test and assert that is fails do to their being a new field with the db tag, but no translation.

## PR Author Review Checklist

## PR Reviewer Guidelines


mappings/translation/plan_general_characteristics.json
```diff
@@ -24,6 +24,24 @@
...
+  },
+  "currentModelPlanID": {
+    "gqlField": "currentModelPlanID",
+    "goField": "CurrentModelPlanID",
+    "dbField": "current_model_plan_id",
+    "label": "What model is this a track of?",
+    "dataType": "UUID",
+    "formType": "TEXT",
+    "tableReference": "model_plan"
+  },
```
patrickseguraoddball, line 44: I think the merge overwrote the property "hideFromReadonly"

patrickseguraoddball (changes requested):

mappings/translation/plan_general_characteristics.json
StevenWadeOddball: Thanks for catching that Patrick! I'll update it

patrickseguraoddball: It was there, was looking in the export, where it's not needed, my bad!

StevenWadeOddball: No problem at all :)

patrickseguraoddball (approved):
LG thanks!

PR #1141, opened 2024-05-31
[EASI-4306] [EASI-4308] New Model Plan Notification Updates

OddTomBrooks:
# EASI-4306, EASI-4308

## Changes and Description

- Added New Model Created notification
  - GQL
  - SQL
  - Notification Type / Wiring
  - Postman collection updates

## How to test this change

- Bring up service containers
- Add relevant user notification preferences for active user
- Create a new 

## PR Author Review Checklist

## PR Reviewer Guidelines


migrations/V158__Add_New_Model_Plan_Notification.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TYPE ACTIVITY_TYPE
+  ADD VALUE 'NEW_MODEL_PLAN' AFTER 'MODEL_PLAN_SHARED';
+
```
StevenWadeOddball, line 3: It would be nice to have a comment on the column (the rest of the columns in this table have one, it makes the column documented in the migration as well as later in whatever db management tool you are using. Example below for the `daily_digest_complete` column
image


```suggest
COMMENT ON COLUMN user_notification_preferences.new_model_plan IS 'Notification preference for when a model plan is created';

```

pkg/models/user_account_notification_preferences.go
```diff
@@ -0,0 +1,28 @@
+package models
+
+import (
+	"github.com/google/uuid"
+	"github.com/samber/lo"
+)
+
+type UserAccountNotificationPreferences struct {
```
StevenWadeOddball, line 8: This was a little confusing to me to know where the data is coming from and what it represents.Have you considered embedding the whole user account object and add the other property? (That way any struct drift is covered here as well. The only thing I don't love about that approach is that UserAccount comes from another package.

I think adding comments to the type and the functions would also help with the clarity of where the data comes from and what the function is doing.

```suggest
type UserAccountNotificationPreferences struct {
	authentication.UserAccount
	NewModelPlanNotification UserNotificationPreferenceFlags `db:"new_model_plan"`
}
```

pkg/models/user_notification_preferences.go
```diff
@@ -43,6 +45,11 @@ func DefaultUserNotificationPreferencesFlags() []UserNotificationPreferenceFlag
 	return []UserNotificationPreferenceFlag{UserNotificationPreferenceInApp, UserNotificationPreferenceEmail}
 }
 
+// EmptyUserNotificationPreferencesFlags returns empty Preferences flags for any user, defaulting to all turned on.
```
StevenWadeOddball, line 48:
```suggest
// EmptyUserNotificationPreferencesFlags returns empty Preferences flags for any user, defaulting to all turned off.
```

pkg/notifications/new_model_plan_meta.go
StevenWadeOddball, line 1: This looks good, can you add some unit tests please?

pkg/storage/user_accountStore.go
```diff
@@ -98,6 +103,30 @@ func (s *Store) UserAccountGetByIDLOADER(
...
 
+// UserAccountNotificationPreferencesNewModelPlan gets the notification
+// preferences for a new model plan for a slice of users who have at least one
+// enabled preference
+func (s *Store) UserAccountNotificationPreferencesNewModelPlan(np sqlutils.NamedPreparer) (
+	[]*models.UserAccountNotificationPreferences,
+	error,
+) {
+
+	var userNotificationPreferences []*models.UserAccountNotificationPreferences
```
StevenWadeOddball, line 114: No need to refactor this unless you want to, but there are a couple examples of a new paradigm that abstracts away some boiler plate work for named statements (`sqlutils.SelectProcedure`). See  `user_notification_preferencesStore.go` for examples.

pkg/graph/resolvers/model_plan.go
```diff
@@ -158,6 +180,27 @@ func ModelPlanCreate(
 				), zap.Error(sendEmailErr))
 			}
 		}()
+
+		for _, emailPref := range newModelPlanEmailPrefs {
```
StevenWadeOddball, line 184: Does this have to iterate through each user because you need the id for the unsubscribe link? Otherwise you could maybe do a BCC list or something.

StevenWadeOddball (commented):
Thanks Tom, this is looking good. I added a couple of notes and thoughts. The relational structure of activity to notification looks right to me.

pkg/graph/resolvers/model_plan.go
OddTomBrooks: I have tried this both ways. In the end, whether by multiple recipients or BCC, the recipient email list is delivered to all recipients. At best, this is inconsistent UX and has a less professional presentation. At worst, there may be some privacy and informational transaction concerns. If we get into multiple thousands of recipients that may be cause for some performance concern, but for the foreseeable future we do not have significant enough scale for this to be a primary concern. At that scale we would need to employ a dispatch service like SES anyway.

In conclusion, for the purposes of consistency with other emails and privacy I will maintain this as individually sent emails rather than batching them.

pkg/models/user_account_notification_preferences.go
OddTomBrooks: Let's talk about this tomorrow. I'd rather not invest much more time on this query, because it's downstream of some tech debt I've discussed with Clay. I'd like to replace this query entirely with something more generic by transposing the notifications table and using an ENUM field to identify notification types rather than columns. That is an intermediate / significant lift, so it needs to be handled separately to this ticket.

pkg/storage/user_accountStore.go
OddTomBrooks: I intended to get to this on Monday, but now facing time constraints and am trying to get backend work for another feature out this week. Though in the same vein as another comment, I suspect this function will be deprecated in favor of a more generic solution after some tech debt is addressed.

pkg/models/user_account_notification_preferences.go
ClayBenson94: Chatted about this with Tom on a call -- I think we're going to keep this as-is and look to maybe have a discussion going forward about how we pull Notification Preferences from the DB. I think it'd be nice to have / use a single `struct` type that can handle most of our cases (has all flags, has user account info), but didn't want to hold the PR just for that.

Maybe we can chat as a group sometime this week and run through the use case / design of this so we can all align on an approach!

pkg/graph/resolvers/model_plan.go
OddTomBrooks: I revisited this. Clay and I did some email testing, and are confident about the behavior of a BCC approach despite the administrative view of Mailcatcher. Suggestion integrated. 👍

StevenWadeOddball: Awesome, thanks for exploring that.

ClayBenson94:
> I think thelink to the model plans should send to the readonly. doublechecking on this.

I think that's right. I'll find the edit!

pkg/email/templates/model_plan_created_body.html
```diff
@@ -6,4 +6,12 @@ <h3>{{.ModelName}}</h3>
 <p>
   <a href="{{.ClientAddress}}/models/{{.ModelID}}/task-list">View this Model Plan in MINT</a>
```
ClayBenson94, line 7:
```suggestion
  <a href="{{.ClientAddress}}/models/{{.ModelID}}/read-only">View this Model Plan in MINT</a>
```

ClayBenson94 (changes requested):
I think per @patrickseguraoddball's finding on the email this is all that's left!

patrickseguraoddball (approved):
LG

patrickseguraoddball (approved):
Bold works!

PR #1159, opened 2024-06-11
[EASI-4427] Preserve question order

patrickseguraoddball:
# EASI-4427

## Changes and Description

- Added `order` to translation schema 
  - Designates the order of the question in the form.  Uses integer as page and question order uses hundreths place.  Ex: 1.01, 1.02, 2.01, 2.02.  Uses integer as page and question order uses hundreths place.  Ex: 1.01, 1.02, 2.01, 2.02
- Added mappings to all translations
- Ran translation export

## How to test this change

Verify all the questions have `order` properties on them (we assign order to all, even if they are not rendered)

## PR Author Review Checklist

## PR Reviewer Guidelines


StevenWadeOddball (approved):
Thanks Patrick! This looks great!

Like we discussed, I'll plan on writing some unit tests to assert that order is populated on the backend instead of manually reviewing here. I'll also verify the order typing on the backend, we might need to change the type to a decimal compatible type for go, but I think that should happen in the BE branch if it's actually needed.

PR #1163, opened 2024-06-12
[NOREF] Updating golangci config

ClayBenson94:
## Changes and Description

- Same change as https://github.com/CMSgov/easi-app/pull/2633
- Enable `stylecheck` linter, as some rules were phased out of `staticcheck` and put into `stylecheck`
- Fixed all findings
- Modified exclusion for package comments to use new `stylecheck ST1000` rule instead of `package-comments` (to maintain previous functionaltiy)

## How to test this change

If using VSCode, you can edit your settings to enable `golangci-lint` as the linter of choice for go files -- this will enable findings to appear within the editor / problems view so you can address issues before having to run the lint script! This configuration will automatically use the `.golangci-lint` file in the root of the project!
image

Additionally -- you can just edit an exported function/var/type to not have a comment, run `scripts/dev lint`, and ensure that the linter catches it!

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -2,10 +2,12 @@ linters-settings:
   gosec:
     # Currently excluding rule G602 due to issue slowing down execution
     # https://github.com/securego/gosec/issues/1010
+    # If we ever re-enable this rule, make sure https://github.com/securego/gosec/issues/1005 is also fixed
     excludes:
       - "G602"
   govet:
-    check-shadowing: true
+    enable:
+      - shadow
     settings:
       printf:
         funcs:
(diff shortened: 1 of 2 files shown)


OddTomBrooks (approved):

StevenWadeOddball (changes requested):
Thanks Clay,

Interestingly, this didn't catch a new exported function added in Model Plan that didn't have a comment. When we paired the other day it worked though 🤔. 

Besides that, I just found a couple of config changes!

StevenWadeOddball (approved):
Thanks Clay!

Thanks for pointing out, I was testing for the absence of a comment on an exported function, which this doesn't check, only that the comment is the right format. This looks great!

PR #1167, opened 2024-06-13
[EASI-4309] Front End work for the Dates Changed Notification

garyjzhao:
# EASI-4309

## Changes and Description

- Add `DatesChanged` component
- Update `IndividualNotification` component to accommodate the `DatesChanged` component
- Add Dates Changed Notification settings in the `Settings` page
- Updated `GetNotification` and `GetNotificationSettings` gql queries
- Add new copy to translation files
- Updated Cypress test to account for this new notification type

## How to test this change

1. Enable email and in-app Dates Changed notifications in the Notification settings page
2. Verify that preferences persist
3. Update dates for any model plan
4. Return to Notification page and verify that a new "Dates Changed" notification popped up
5. Hit the dropdown and verify the dates changed matches and the design matches figma
6. Unsubscribe via email link `http://localhost:3005/notifications/settings?unsubscribe_email=DATES_CHANGED` 
7. Verify that unsubscribing actually works
8. Verify that tests pass

## PR Author Review Checklist

## PR Reviewer Guidelines


patrickseguraoddball (changes requested):
I had one scenario where the `datesChangedNotificationType` was set to null even though the checkbox were checked.  This resulted in no notifcations. These were the steps I was able to reproduce.  

1. Check all notifications for both in-app and email
2. Create a model plan
3. Usubscribe via email link
4. Once success banner renders, recheck the same box for `When a new Model Plan is created`
5. Save
6. Mutation passed a value of `null` for `datesChangedNotificationType`
7. Return to setting page, all notifications are checked but the query returns `null` for `datesChangedNotificationType`

patrickseguraoddball:
Additionally when the link is clicked via email, there's a brief flicker of the error banner alert, before the success banner renders.  We should look into this behavior.

src/views/Notifications/Settings/index.tsx
```diff
@@ -85,6 +96,18 @@ const NotificationSettings = () => {
...
+    // if datesChangedNotificationType is not changed by user, but datesChanged is changed, then do the following logic
+    if (!changes.datesChangedNotificationType) {
+      if (changes.datesChanged?.length) {
+        // If Dates Changed notification is subscribed, then manually set datesChangedNotificationType to ALL_MODELS
+        changes.datesChangedNotificationType =
+          DatesChangedNotificationType.ALL_MODELS;
+      } else {
+        // If Dates Changed notification is unsubscribed, set datesChangedNotificationType to null
+        changes.datesChangedNotificationType = null;
+      }
```
patrickseguraoddball, line 108: `datesChangedNotificationType` would already be null, no need to set it again here

src/views/Notifications/Settings/index.tsx
```diff
@@ -85,6 +96,18 @@ const NotificationSettings = () => {
...
     };
 
+    // if datesChangedNotificationType is not changed by user, but datesChanged is changed, then do the following logic
+    if (!changes.datesChangedNotificationType) {
+      if (changes.datesChanged?.length) {
+        // If Dates Changed notification is subscribed, then manually set datesChangedNotificationType to ALL_MODELS
+        changes.datesChangedNotificationType =
+          DatesChangedNotificationType.ALL_MODELS;
+      } else {
+        // If Dates Changed notification is unsubscribed, set datesChangedNotificationType to null
```
patrickseguraoddball, line 106: I think this is part of the culprit.  When the page is first hit by the email link, `datesChangedNotificationType` is not null, the mutation fires and it get set to null.  Then once rechecked, it appears to not be dirty anymore, hence it explicitly sets it to null

patrickseguraoddball (changes requested):

src/i18n/en-US/notifications.ts
```diff
@@ -103,12 +124,12 @@ const notifications = {
...
-      dateChanges: 'When model dates change',
+      datesChanged: 'When model dates change',
       whichModel: 'Which models?',
       whichModelTypes: {
-        all: 'All models',
-        following: 'Models I’m following',
-        mine: 'My Model Plans'
+        ALL_MODELS: 'All models',
+        FOLLOWED_MODELS: 'Models I’m following',
+        MY_MODELS: 'My Model Plans'
```
patrickseguraoddball, line 132: Not sure why these are capitalized differently?  Is it being keyed on by a value somewhere?

src/views/Notifications/Settings/index.tsx
```diff
@@ -290,7 +366,36 @@ const NotificationSettings = () => {
...
+                Object.keys(obj1)
+                  .filter(key =>
+                    Object.prototype.hasOwnProperty.call(obj2, key)
+                  )
+                  .reduce((acc, key) => {
+                    acc[key] = obj1[key];
+                    return acc;
+                  }, {} as Obj);
+
+              const basicNotificationKeys = getMatchingKeys(
```
patrickseguraoddball, line 395: I think the naming of this is misleading, this represents basic notification values which are being used in the form to represent the checkboxes

patrickseguraoddball (changes requested):

src/views/Notifications/Settings/index.tsx
```diff
@@ -85,6 +96,13 @@ const NotificationSettings = () => {
       ...dirtyInputs
     };
 
+    // if datesChangedNotificationType is not changed by user, but datesChanged is changed, then do the following logic
+    if (!changes.datesChangedNotificationType && changes.datesChanged?.length) {
+      // If Dates Changed notification is subscribed, then manually set datesChangedNotificationType to ALL_MODELS
+      changes.datesChangedNotificationType =
```
patrickseguraoddball, line 102: There's a few instances where consistency with capitalization in comments is off.  Not huge just a nitpick

src/i18n/en-US/notifications.ts
garyjzhao: the `whichModelTypes` are in all caps to mimic the enums, as it will be sent along with the payload to the update mutation

src/views/Notifications/Settings/index.tsx
```diff
@@ -290,7 +366,36 @@ const NotificationSettings = () => {
...
+               */
+              const getMatchingKeys = (obj1: Obj, obj2: Obj): Obj =>
+                Object.keys(obj1)
+                  .filter(key =>
+                    Object.prototype.hasOwnProperty.call(obj2, key)
+                  )
+                  .reduce((acc, key) => {
+                    acc[key] = obj1[key];
+                    return acc;
+                  }, {} as Obj);
```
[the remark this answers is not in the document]
garyjzhao: Great idea. I wanted to do something like this when I was initially working on this, but couldn't come up with it. Implemented in the latest update

patrickseguraoddball (approved):
LG Nice!

PR #1182, opened 2024-06-20
[NOREF] Mutation handler improvement

patrickseguraoddball:
# NOREF

## Changes and Description

- Early return on `useHandleMutation` if there are no dirty changes.  This prevents CH from marking as task list section as `In progress` if no changes have been made when entering the form

## How to test this change

Verify mutation doesn't get called on task list sections (Basics and GenChar are the only pages that have complex mutations and may come back to in another ticket)

## PR Author Review Checklist

## PR Reviewer Guidelines

### DIFF
--- a/src/hooks/useHandleMutation.ts
+++ b/src/hooks/useHandleMutation.ts
@@ -75,6 +75,16 @@ function useHandleMutation<TData = any, TVariables = OperationVariables>(
           formikRef?.current?.values
         );
 
+        // If no changes, don't call mutation
+        if (Object.keys(changes).length === 0) {
+          unblock();
+          history.push({
+            pathname: destination.pathname,
+            state: destination.state
+          });
+          return false;
+        }
+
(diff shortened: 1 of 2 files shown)


StevenWadeOddball (commented):
Thanks Patrick!

I confirmed that Model basics still calls update with non-dirty changes (for basics and model plan). Gen Characteristics still does as well.

Plan Beneficiaries also updates for number of people impacted, I think the form instantiates the answer to 0, instead of denoting that the answer wasn't set. The other pages for the form worked as expected.

Participants and Providers also does it for expected number of participants. I suspect it is something to do with this slider component?
image

Ops Eval and Learning  and Payments works as expected!

StevenWadeOddball (approved):
that sounds good to me @patrickseguraoddball! I think this should be good to go ahead and merge with that in mind, and other changes can be be handled separately as desired.

PR #1192, opened 2024-06-25
[EASI-4446] Get Model Plan and Operational Solution Data by Operational Solution Key

OddTomBrooks:
# EASI-4446

## Changes and Description

- GQL
- SQL
- Models
  - Raw model for deserialization of SQL query
  - Added full constructors for Operational Solution and Model Plan for deserialization purposes
- Updated Postman collection
- Unit tests

## How to test this change

For your convenience, a variety of tests has been added as unit tests.
For manual testing, please use the Postman query `Get Model Plans By Operational Solution Key` and structure data however you see fit.

## PR Author Review Checklist

## PR Reviewer Guidelines
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -


MINT.postman_collection.json
```diff
@@ -2338,42 +2359,100 @@
...
+									"// for (let i =0; i< notifications.length; i++){",
+									"//     subtaskID = notifications[i].id",
+									"//     pm.collectionVariables.set(\"operationalSolutionSubtaskID\"+i, subtaskID);",
+									"// }"
+								],
+								"type": "text/javascript",
+								"packages": {}
+							}
+						}
+					],
```
OddTomBrooks, line 2384: User view customization changes shouldn't be here, cleaning it up

pkg/models/model_plan_and_operational_solution_raw.go
```diff
@@ -0,0 +1,39 @@
+package models
+
+import (
+	"time"
+
+	"github.com/google/uuid"
+)
+
+// ModelPlanAndOperationalSolutionRaw represents the combined data from the SQL query with unique field names
+type ModelPlanAndOperationalSolutionRaw struct {
```
OddTomBrooks, line 10: This structure exists to disambiguate fields with shared names between model plan and operational solution, such as `id` or `created_by`.

pkg/storage/model_planStore.go
```diff
@@ -361,3 +366,88 @@ func (s *Store) ModelPlanDeleteByID(logger *zap.Logger, id uuid.UUID) (sql.Resul
...
+			rawResult.ModelPlanID,
+			rawResult.ModelName,
+			rawResult.Abbreviation,
+			rawResult.Archived,
+			models.ModelStatus(rawResult.ModelPlanStatus),
+			rawResult.ModelPlanCreatedBy,
+			rawResult.ModelPlanCreatedDts,
+			rawResult.ModelPlanModifiedBy,
+			rawResult.ModelPlanModifiedDts,
+		)
```
OddTomBrooks, line 444: Here we decompose our raw intermediate object into well-known structures.

pkg/storage/model_planStore.go
```diff
@@ -361,3 +366,88 @@ func (s *Store) ModelPlanDeleteByID(logger *zap.Logger, id uuid.UUID) (sql.Resul
...
+	if err != nil {
+		return nil, err
+	}
+	defer stmt.Close()
+
+	arg := map[string]interface{}{
+		"operational_solution_key": opSolKey,
+	}
+
+	rows, err := stmt.Queryx(arg)
```
OddTomBrooks, line 385: Usage of Queryx and row-based iteration here is a matter of preference. I probably could have deserialized directly into a slice then decomposed the slice. Both ways work. Theoretically there could be some micro-considerations on mem stack performance, but realistically the number of models and opSols we'll see is very unlikely to be cause for concern here.

pkg/storage/SQL/model_plan/get_by_operational_solution_key.sql
```diff
@@ -0,0 +1,39 @@
+SELECT
```
OddTomBrooks, line 1: I have wondered if this is better implemented as a DataLoader implementation. The cost/benefit proposition on what will typically be small result sets seems dubious, but I am interested in hearing other perspectives on this. In the meantime, I built this out as a quick and simple(r) solution which can be refactored if it makes sense to do so.

pkg/storage/model_planStore.go
ClayBenson94: > deserialized directly into a slice then decomposed the slice

Can you elaborate on this? I feel like with the way the `ModelPlanAndOperationalSolution` struct is designed we could just use a `sqlx.Select()` method to populate the slice of structs automatically (without any need for iterating over the rows, or the `models.New*Full()` methods)

ClayBenson94 (changes requested):

pkg/storage/model_planStore.go
OddTomBrooks: The constructors are a workaround for package scope issues I was running into with inline construction.

Connected to discussion on deserialization, unless we can find a way to scan embedded structs in a way which disambiguates fields we will still need to use *Raw as an intermediary struct to capture query results regardless of sqlx method used. Using Queryx rather than Select lets us hold less data in memory during the Raw-to-embedded conversion process. That memory performance is extremely unlikely to matter. In theory we can use Select to scan a slice, but unless we can find a solution for scanning embedded structs we have to do a loop to convert types either way. There is something to be said for code consistency in using Select, on the other hand this is what Queryx is for. I can see arguments for either usage. My preference leans a bit to Queryx and since the value proposition is low to invest in a Select query (without a solution for embedded struct scanning) I left it as-is pending significantly strong feelings from the team.

pkg/graph/schema/types/operational_solution_and_model_plan.graphql
```diff
@@ -0,0 +1,19 @@
...
+  statusPlannedActiveOrEnded: StatusPlannedActiveOrEnded!
+}
+
+type ModelPlanAndOperationalSolution {
+  operationalSolution: OperationalSolution!
+  modelPlan: ModelPlan!
+}
+
+extend type Query {
+  modelPlansByOperationalSolutionKey(operationalSolutionKey: OperationalSolutionKey!): [ModelPlanAndOperationalSolution]!
```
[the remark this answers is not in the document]
OddTomBrooks: Oops. Good catch.

pkg/models/model_plan_and_operational_solution_raw.go
OddTomBrooks: On the one hand, I agree 100% with your concerns about redundancy in field definitions. I spent about half a day looking for a solution to this before falling back on the current implementation. Multiple tag definitions per your suggestion is an interesting idea, looking into this to see if it works.

Conversely, since we explicitly define our field selections in SQL rather than using wildcards it could be argued that the additional code brittleness is a mild extension of an existing architectural issue. That said, I’m wondering if we should consider wildcard usage and fall back on unit tests to assert deserialization success.

pkg/storage/SQL/model_plan/get_by_operational_solution_key.sql
OddTomBrooks: Thank you for elaborating on your consideration for DataLoader usage. Based on that criteria, I agree 💯.

pkg/storage/model_planStore.go
ClayBenson94: > unless we can find a solution for scanning embedded structs we have to do a loop to convert types either way

I see what you mean! I think `sqlx.Select` would only really simplify the `rows.StructScan(&rawResult)` call (which is, under the hood, what those sqlx struct methods call as well)

I think we should probably have a sync today to brainstorm some alternative approaches to this to see if there's anything we can do to ease the amount of wiring we have to do to get these results turned into concrete types.

(Part of me is tempted to create a new type to represent this joined data and use that instead of _actual_ ModelPlan and OperationalSolution types)

pkg/storage/operational_solutionStore.go
```diff
@@ -72,6 +77,27 @@ func (s *Store) OperationalSolutionGetByID(_ *zap.Logger, id uuid.UUID) (*models
...
+		"paramTableJSON": paramTableJSON,
+	}
+
+	opSols, err := sqlutils.SelectProcedure[models.OperationalSolution](
+		s.db,
+		sqlqueries.OperationalSolution.GetByIDLOADER,
+		arg,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("issue selecting operational solution by id with the data loader, %w", err)
```
ClayBenson94, line 95: This might be a good case for using the *zap.Logger passed in here -- perhaps something like this! (Though this commit won't work directly)

```suggestion
		errorMessage := "error selecting operational solution by id"
		logger.Error(errorMessage, zap.Error(err))
		return nil, errors.Wrap(err, errorMessage)
```

This allows us to see the trace of the error through the logs, but also retains the error messages context through the call chain!

pkg/graph/schema/types/operational_solution_and_model_plan.graphql
```diff
@@ -0,0 +1,22 @@
...
+  ACTIVE
+  ENDED
+}
+
+extend type ModelPlan {
+  statusPlannedActiveOrEnded: StatusPlannedActiveOrEnded!
+}
+
+type ModelPlanAndOperationalSolution {
+  # TODO -- we can uncomment these if we want?
```
ClayBenson94, line 12: Can you either uncomment this / remove the comment?

Sorry -- I definitely threw this in last minute before handing the branch over!

pkg/models/model_plan.go
```diff
@@ -19,8 +23,36 @@ func NewModelPlan(createdBy uuid.UUID, modelName string) *ModelPlan {
 		Archived:   false,
 		Status:     ModelStatusPlanDraft,
 	}
+}
 
+// NewModelPlanFull returns a new model plan with all fields set
+func NewModelPlanFull(
```
ClayBenson94, line 29: I think we can get rid of this method now that it's not used in the storage methods anymore!

pkg/models/operational_solution.go
```diff
@@ -37,6 +37,54 @@ func NewOperationalSolution(createdBy uuid.UUID, operationalNeedID uuid.UUID) *O
 	}
 }
 
+func NewOperationalSolutionFull(
```
ClayBenson94, line 40: Similar to NewModelPlanFull -- I think we can get rid of this method now that it's not used in the storage methods anymore!

ClayBenson94 (changes requested):
Thanks for the quick updates, Tom!

Just a couple small improvements / cleanups, but other than that I think this is all set!

ClayBenson94 (approved):

PR #1199, opened 2024-06-27
[Easi-4472]/Change History backend cleanup part 1

StevenWadeOddball:
# EASI-4472

## Description

This PR is starting to address a lot of notes that were left behind to come back to for Change History

1. SQL column comments
    a. Translated Audit
    b. Translated Audit Field
    c. Translated Audit Queue
2. Translated Audit Queue create sql removed. More clarity around the queue item
3. Translated Audit Field
     a. make TranslationDataType and TranslationForm Type not nullable. 
4. Translated Audit Field Meta Data removed
     a. This was not actually being used anywhere. If we need to, we can add it again at later time
5. Introduce table name typing in locations where it was missed
6. Remove unnecessary comments
7. Remove unused code
## How to test this change

This changes should not affect functionality. Please smoke test the application and make sure there is not affected functionality. If so, this should be fine code wise.

Please also look at the stylistic refactoring. If there is any comments removed that you disagree with, please voice your concerns. I can address them here, or in subsequent cleanup PR.

- 

## PR Author Checklist

## PR Reviewer Guidelines


mappings/translation_test_utilities.go
```diff
@@ -187,12 +187,9 @@ func assertTFieldBase(t *testing.T, field reflect.StructField, base models.Trans
 	assertStringPointerNilOrNotEmpty(t, base.SubLabel, "SubLabel", field)
 	assertStringPointerNilOrNotEmpty(t, base.MultiSelectLabel, "MultiSelectLabel", field)
 
-	// assert.NotZero(t, base.IsArray) // not zero doesn't work for bool because false is zero
-
-	assert.NotZero(base.DataType)
+	assert.NotZero(base.DataType, "issue for field %s. Value: %s", field.Name)
```
ClayBenson94, line 190: Should this be using `NotZerof` if you want to use formatted strings?

StevenWadeOddball: yes! Good call, thanks, I'll update these method calls.

patrickseguraoddball (approved):
As far a smoke test this works as expected/no change in functionality :)

PR #1210, opened 2024-07-02
Easi 4484/ch backend cleanup v2

StevenWadeOddball:
# EASI-4484
Address remaining TODO notes from the change history branch.
## Description

- Generically provide mechanism to skip translating table by table ID.
- Update some todos to Future Enhancement notes that were not needed for deployment
- Stop fetching the model plan directly for the audit, since the data was not need.
- Merge in Main (note, most of these file changes are actually just from merging in main)
- Simplify Seed command to only have one translate command
- Unit test to validate that db tags are not duplicated on a translation.

## How to test this change
1. Functionality should remain unchanged.
2. Please do a smoke test on change history, and also see if there are any addressed comments you don't agree with.

## PR Author Checklist

## PR Reviewer Guidelines


patrickseguraoddball (commented):
Smoke test was successful, works as expected!

PR #1212, opened 2024-07-03
[NOREF] Added escapeValue false for i18n

patrickseguraoddball:
# NOREF

## Description

MINT was having instances where special HTML character were getting escaped by react-i18n library via interpolation values, as well as React (natively parsing all JSX values as plain text/not computed by the browser - we do NOT have `dangerouslysetinnerhtml` set in the application).  This was leading to special characters being escaped and also being rendered as plain text - ex: `Patrick&#39;s`

https://oddball.slack.com/archives/C059N01AYGM/p1719943391656779

## How to test this change

- Verify user input characters are rendered as human readable throughout the app.  The are handled as plaint text natively by React

## PR Author Checklist

## PR Reviewer Guidelines

### DIFF
--- a/src/i18n.ts
+++ b/src/i18n.ts
@@ -16,7 +16,7 @@ i18n
     fallbackLng: 'en-US',
     resources,
     interpolation: {
-      escapeValue: false // react already safes from xss => https://www.i18next.com/translation-function/interpolation#unescape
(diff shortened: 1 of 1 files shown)


StevenWadeOddball (approved):
I tested this by creating a model plan with an `'`, and I verified that the text renders correctly! Thanks @patrickseguraoddball

PR #1219, opened 2024-07-09
[EASI-4504] Improve local user accounts

StevenWadeOddball:
# EASI-4504

## Description

Before this PR, all test user data followed the pattern user_name + "Doe". This was not the best experience for demos etc. This PR aimed to expose other test data so it would be more meaningful for demos and local development.

- Replace the mock user data call by instead wiring up the local Okta API client.
- getMockUserData replaced with a dictionary variable.
- Additional mock user data added to include values seen in FIGMA as well as other additional test names

## How to test this change

1.  bring up the app
2. log in with local login as any user that has test data in `pkg/local/okta_api.go`.
3. Verify that  a User account is made that matches the data provided in the local files
4. Log out
5. Log in with an EUAID that isn't in the list, verify that you can still log in with the old UserName + "Doe" paradigm.

Collaborator
1. Search for users in the list
2. Add as a collaborator

Discussions
1. Tag users from the local dev data in a discussion. 
2. Verify it looks correct

Tests
1. Verify that all the tests continue to function as expected. 

## PR Author Checklist

## PR Reviewer Guidelines


pkg/graph/resolvers/resolver_test_utilities.go
```diff
@@ -25,15 +26,16 @@ import (
...
+	DBConfig      storage.DBConfig
+	LDClient      *ld.LDClient
+	Logger        *zap.Logger
+	UserInfo      *models.UserInfo
+	Store         *storage.Store
+	S3Client      *upload.S3Client
+	PubSub        *pubsub.ServicePubSub
+	Principal     *authentication.ApplicationPrincipal
+	Context       context.Context
+	FetchUserInfo func(context.Context, string) (*models.UserInfo, error)
```
ClayBenson94, line 38: Would it be easier to just put the entire Okta client here? We do similar stuff with `Store`, and `S3Client`

pkg/graph/resolvers/resolver_test.go
```diff
@@ -28,11 +30,32 @@ func (suite *ResolverSuite) SetupTest() {
 	err := suite.testConfigs.Store.TruncateAllTablesDANGEROUS(suite.testConfigs.Logger)
 
 	//GET USER ACCOUNT EACH TIME!
-	princ := getTestPrincipal(suite.testConfigs.Store, suite.testConfigs.UserInfo.Username)
+	princ := suite.GetTestPrincipal(suite.testConfigs.Store, suite.testConfigs.UserInfo.Username)
 	suite.testConfigs.Principal = princ
 	assert.NoError(suite.T(), err)
 }
 
+func (suite *ResolverSuite) GetTestPrincipal(store *storage.Store, userName string) *authentication.ApplicationPrincipal {
```
ClayBenson94, line 38:
```suggestion
func (suite *ResolverSuite) getTestPrincipal(store *storage.Store, userName string) *authentication.ApplicationPrincipal {
```

Thinking it might make sense to not have this method exported -- does anything outside the resolver suite need it?

pkg/graph/resolvers/resolver_test_utilities.go
StevenWadeOddball: Yeah! I think that's a good idea, I'll try it out.

pkg/graph/resolvers/resolver_test.go
StevenWadeOddball: You are right, nothing should use it outside the suite, so that makes sense. I'll make the change

 Interestingly, if you reference the suite in the same package, but not in a test file, you get a GO-Static check failure.

pkg/graph/resolvers/resolver_test_utilities.go
StevenWadeOddball: I updated it! Largely throughout the test code, a stubFetchUserInfo method is being called, but I think that is ok for those cases being tested.

ClayBenson94: Nice. Might just be me, but I like seeing `OktaClient oktaapi.Client` and just knowing that we have those methods available in the test client, rather than having to read/understand `FetchUserInfo func(context.Context, string) (*models.UserInfo, error)` every time 😂

StevenWadeOddball: 💯! Thanks for suggesting it! I think that will be an improvement to have that ability as well!

patrickseguraoddball (approved):
LG thanks!

PR #1221, opened 2024-07-11
[EASi-4511] Convert empty strings to nil for translations

StevenWadeOddball:
# EASI-4511

## Description

It was noted that certain conditions can create a translation error when a field tries to write an empty string to the old, new or translated fields. This is not allowed as the fields are `ZERO_STRINGS` (requiring non empty strings or nil).

To address situations where an answer changes from nil to "", we now convert "" to nil. If the answer is effectively unchanged, the field is not saved as a translated field.

This happened effectively when a new CR was made with an empty note field.

## How to test this change

- Create a model plan
- Create a plan CR, enter text in the note field, but then delete all the text
- translate the audit changes.
- verify that there is an entry for the new CR, and also verify that the note field does not render on the front end.

## PR Author Checklist

## PR Reviewer Guidelines

### DIFF
--- a/.github/workflows/change_history_dev_deploy.yml
+++ b/.github/workflows/change_history_dev_deploy.yml
@@ -12,9 +12,7 @@ jobs:
   run_tests:
     uses: ./.github/workflows/run_tests.yml
     secrets: inherit
-    with:
-      continue_on_error: true
-      
+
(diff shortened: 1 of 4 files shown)


patrickseguraoddball (approved):
Works as expected!

PR #1226, opened 2024-07-15
[EASi-4520]/Change history job logging

StevenWadeOddball:
# EASI-4520

## Description

The jobs for change history currently will fail if there is a duplicate entry attempting to translate an audit. This might happen in lower environments based on the timing of the `postgres` database being cleaned, and a mismatch with the faktory reddis instance. This isn't actually an issue though, so we should ideally not fail the job. 

This PR accomplishes that by typing that error, and not failing if so. This also begins to expand error handling for postgres data, so that `errors.As` can be used to determine specific error types. You will note that sqlutils get and select now process all errors. This can be expanded to handle new error types as desired in the future.

Also, the translated audit job is now a sugared logger with the worker JID, internal translated_audit_queue_id, and the audit.change id. 

## How to test this change
1. Run the app with Faktory running locally (
    a. set envrc.local variable values
          i.  APP_ENV=local
          ii. FAKTORY_PROCESS_JOBS=TRUE
         
 2. Seed the database.
 3. Once the audits have been translated, manually update the data in the database to try to reprocess all these already processed jobs
 ```sql
UPDATE translated_audit_queue
SET
 status = 'NEW',
 modified_by = '00000001-0001-0001-0001-000000000001',
 modified_dts = CURRENT_TIMESTAMP

RETURNING *
 ```
  

4. Ensure that the duplicate processing doesn't cause the job to fail.
5. You can also inspect the Docker Container to look at the logging messages.

Unit tests have been updated to also replicate this condition for the translated audit job.

## PR Author Checklist

## PR Reviewer Guidelines

### DIFF
--- a/pkg/sqlqueries/SQL/translated_audit_queue/update.sql
+++ b/pkg/sqlqueries/SQL/translated_audit_queue/update.sql
@@ -8,7 +8,7 @@ SET
     created_by= :created_by,
     created_dts= :created_dts,
     modified_by= :modified_by,
-    modified_dts= :modified_dts
+    modified_dts= CURRENT_TIMESTAMP
 WHERE id =  :id
 RETURNING
 id,
--- a/pkg/sqlutils/errors.go
+++ b/pkg/sqlutils/errors.go
@@ -1,0 +1,75 @@
+package sqlutils
+
+import (
+	"fmt"
+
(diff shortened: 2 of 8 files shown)


patrickseguraoddball (approved):
All the test criteria passed!

PR #1230, opened 2024-07-17
[EASI-4394] Models by Solution homepage

patrickseguraoddball:
# EASI-4394

## Description

- Updated `GlobalTableFilter` to not have decoupled state with query.  State setter and state are now coupled.  This fixes issues with discrepancies between text in search box and state of global filter.
- Component for banner with button group for toggling status
  - Statuses groups should match comment per Figma (planned is anything not active or ended, etc)
- Card component for rendering models of solutions
- Parent/wrapper component to house all individual components
- Mobile/Tablet responsive
  - Should render select dropdown on tablet or when too many solutions
- Unit tests

## How to test this change

Note: When searching with query and no results, query box and results will persist across status buttons, even though search is normally not available on statuses that have less than 4 results.  This is to persists the search while notifying the user the search input is still active.

- Seed db
- Configure homepage settings to include models by solution
- Add a few solutions
- Add more than 4 models to a single solution
- Verify search and pagination appear and work correctly (pagination appears at more than 3)
- Test mobile responsiveness

## PR Author Checklist

## PR Reviewer Guidelines


StevenWadeOddball (approved):
Thanks @patrickseguraoddball

I did a functional test and verified that the homepage functions as expected. Search appears when at least 5 models are present, and I verified that pagination also appears with 4 models.

The select drop down also appears as expected at mobile widths. 🚀

PR #1233, opened 2024-07-17
[EASI-4532]/customizable homepage status update

StevenWadeOddball:
# EASI-4532

## Description

The purpose of this ticket is to make it so that paused and cancel doesn't show up as planned in the models by operational solution view
- Seed Data now seeds multiple model plans that use the 4innovation solution
- Seed data now seeds a default view for the `MINT` user
- FE uses the status from the backend to organize model_plans_by_solution
- Exhaustive BE unit tests on expected status

## How to test this change

- Bring up the app
- Seed the database.
- Log in as the mint user
- Look at the models by operational solution. 
- Verified that paused and cancelled don't show up as PLANNED

## PR Author Checklist

## PR Reviewer Guidelines


cypress/e2e/homepageSettings.js
```diff
@@ -1,6 +1,6 @@
 describe('Homepage Settings', () => {
   beforeEach(() => {
-    cy.localLogin({ name: 'MINT' });
+    cy.localLogin({ name: 'ANON' });
```
ClayBenson94, line 3: Was there something in this change that made this necessary?

pkg/graph/resolvers/operational_solution_and_model_plan.go
```diff
@@ -14,3 +14,18 @@ func ModelPlansByOperationalSolutionKey(
 ) ([]*models.ModelPlanAndOperationalSolution, error) {
 	return store.ModelPlanGetByOperationalSolutionKey(logger, operationalSolutionKey)
 }
+
+// ModelPlanByOperationalSolutionStatusPlannedActiveOrEnded converts a Model status to a ModelBySolutionStatus
+func ModelPlanByOperationalSolutionStatusPlannedActiveOrEnded(modelPlanStatus models.ModelStatus) models.ModelBySolutionStatus {
```
ClayBenson94, line 19: Can you rename this function to match the new name in the GQL schema?

src/views/HomeNew/components/ModelsBySolutions/index.test.tsx
```diff
@@ -89,6 +94,7 @@ const mocks = [
             modelPlan: {
               id: '4fc87324-dbb0-4867-8e4d-5a20a76c8ae2',
               modelName: 'Plan with Basics',
+              modelBySolutionStatus: ModelBySolutionStatus.ENDED,
```
ClayBenson94, line 97: Should we add a case for `ModelBySolutionStatus.OTHER` here?

ClayBenson94 (changes requested):

cypress/e2e/homepageSettings.js
StevenWadeOddball: Yes, the seed data for MINT adds view customization for the MINT user. This cypress test only expects that the user doesn't have any customizations. `ANON` doesn't have any seed data associated.

pkg/graph/resolvers/operational_solution_and_model_plan.go
StevenWadeOddball: Good call! Thanks!

src/views/HomeNew/components/ModelsBySolutions/index.test.tsx
StevenWadeOddball: Good call, I'll add it.

StevenWadeOddball: Added in [50aa733](https://github.com/CMSgov/mint-app/pull/1233/commits/50aa733351ba687742b771e9fac551ed3c085ae8)

pkg/graph/resolvers/operational_solution_and_model_plan.go
StevenWadeOddball: [23ee4d0](https://github.com/CMSgov/mint-app/pull/1233/commits/23ee4d08764826ad992fe8ae9f297dd668bd48a5)

cypress/e2e/homepageSettings.js
ClayBenson94: Ah! Might be worth just a quick `// Using ANON since that user has no customized views` in case anyone ever adds data and sees that test break! But thanks for the explanation!

ClayBenson94 (approved):

PR #1250, opened 2024-07-25
[EASI-4536] (Backend) Model Plan Anticipated Statuses

OddTomBrooks:
# EASI-4536

## Description

A functional resolver, AnticipatedStatuses, has been created on ModelPlan to determine anticipated statuses based on the following conditions.

- “Complete ICIP” date in anticipated timeline has passed but status is not "ICIP completed" or a future status
- “Clearance start date” date in anticipated timeline has passed but status is not any of the clearance statuses or a future status
- “Clearance end date” date in anticipated timeline has passed but status is not "Cleared" or a future status
- “Announce model” date in anticipated timeline has passed but status is not "Announced" or a future status
- “Performance start date” date in anticipated timeline has passed but status is not "Active" or a future status
- “Performance end date” date in anticipated timeline has passed but status is not "Ended"
- For all of the above, modal should only show if status is NOT "Paused" or "Canceled"

## How to test this change

- Create a model plan
- Set dates according to what you'd like to test in Plan Basics
- Use a Postman query to get the corresponding ModelPlan
- Assert that you are provided with the appropriate corresponding anticipated statuses

> [!NOTE]
> A date on the current day is considered to have passed. I could see reasoning going either way on this. I left the default behavior because it seems a bit more intuitively appropriate, besides being mathematically correct.

## PR Author Checklist

## PR Reviewer Guidelines

### DIFF
--- a/MINT.postman_collection.json
+++ b/MINT.postman_collection.json
@@ -39,7 +39,8 @@
 									"pm.collectionVariables.set(\"operationalNeedID\", firstOpNeed);",
 									"pm.collectionVariables.set(\"userID\", userID);"
 								],
-								"type": "text/javascript"
+								"type": "text/javascript",
+								"packages": {}
 							}
 						}
 					],
@@ -49,7 +50,7 @@
 						"body": {
 							"mode": "graphql",
 							"graphql": {
-								"query": "mutation createModelPlan{\n  \n  createModelPlan(modelName:\"Excellent Model\" ){\n    id\n    modelName\n    createdByUserAccount{\n        givenName\n        id\n    }\n    operationalNeeds {\n\n        id\n        modelPlanID\n        name\n        key\n        nameOther\n        needed\n        section    \n    }  \n    basics{\n        id        \n    }\n    generalCharacteristics {\n        id\n    }\n    participantsAndProviders {\n        id\n    }\n    beneficiaries {\n        id\n    }\n    opsEvalAndLearning {\n        id\n    }\n    collaborators {\n      id\n    }\n    payments{\n        id\n    }\n  }\n}\n  ",
+								"query": "mutation createModelPlan{\n  \n  createModelPlan(modelName:\"Excellent Model\" ){\n    id\n    modelName\n    createdByUserAccount{\n        givenName\n        id\n    }\n    operationalNeeds {\n        id\n        modelPlanID\n        name\n        key\n        nameOther\n        needed\n        section    \n    }  \n    basics{\n        id        \n    }\n    generalCharacteristics {\n        id\n    }\n    participantsAndProviders {\n        id\n    }\n    beneficiaries {\n        id\n    }\n    opsEvalAndLearning {\n        id\n    }\n    collaborators {\n      id\n    }\n    payments{\n        id\n    }\n  }\n}\n  ",
 								"variables": ""
 							}
 						},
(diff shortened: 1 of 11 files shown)


MINT.postman_collection.json
```diff
@@ -144,7 +145,7 @@
 						"body": {
 							"mode": "graphql",
 							"graphql": {
-								"query": "query modelPlanCollection {\n    modelPlanCollection(filter: WITH_CR_TDLS) {\n        id\n        modelName\n        abbreviation\n        archived\n        createdBy\n        createdDts\n        modifiedBy\n        modifiedDts\n        status\n        isFavorite\n        isCollaborator\n        lastModifiedDts\n    }\n}",
```
OddTomBrooks, line 147: lastModifiedDts should be opSolutionLastModifiedDts, I corrected this query since I was modifying it anyway according to boyscout rules

pkg/graph/resolvers/status_evaluation_strategies.go
```diff
@@ -0,0 +1,143 @@
...
+
+import (
+	"time"
+
+	"github.com/cmsgov/mint-app/pkg/shared/utilitytime"
+
+	"github.com/cmsgov/mint-app/pkg/models"
+)
+
+// GetAllStatusEvaluationStrategies returns all status evaluation strategies
```
ClayBenson94, line 11: Can you update/add comment on this function that explains the criticality/importance of the order of the strategies? (Mainly that the first matching strategy to evaluate to true is what gets used, so that's why they're in the order they're in)

pkg/graph/resolvers/status_evaluation_strategies.go
```diff
@@ -0,0 +1,143 @@
...
+
+// Evaluate returns the chronologically earliest clearance status
+// (models.ModelStatusInternalCmmiClearance) if the model plan status is not
+// "Plan Complete" and the clearance start date has passed
+func (s *ClearanceStartStrategy) Evaluate(
+	modelPlanStatus models.ModelStatus,
+	planBasics *models.PlanBasics,
+) ([]models.ModelStatus, bool) {
+	if models.GetModelStatusChronologicalIndex(modelPlanStatus) <
+		models.GetModelStatusChronologicalIndex(models.ModelStatusInternalCmmiClearance) &&
```
ClayBenson94, line 60: I might just wanna double check this one against the AC, since it's one that is a bit more complex than the others:
> “Clearance start date” date in anticipated timeline has passed but status is not any of the clearance statuses or a future status

The `but status is not any of the clearance statuses` part makes me think we should check the _latest_ (chronologically) clearance status, since `ModelStatusInternalCmmiClearance: 3`, but a`ModelStatusOmbAsrfClearance: 6`

I only suggest it because right now if your `ClearanceStarts` date has passed, but you're in `ModelStatusCmsClearance`, it'll still suggest an update (even though you're already in a clearance status)

Maybe worth another chat with Natalie to see if we're interpreting the A.C. correctly on that one!

pkg/models/model_plan.go
```diff
@@ -92,6 +92,33 @@ var ModelStatusHumanized = map[ModelStatus]string{
 	ModelStatusCanceled:              "Canceled",
 }
 
+var modelStatusChronologicalIndex = map[ModelStatus]int{
```
ClayBenson94, line 95: Might wanna add one more comment here to this map as well -- just to clarify what it means, and if there's any specific gotchas (like why `Paused` and `Canceled` are at the end, just so people reading this for the first time / coming back to it can get a quick grip on what's being done!)

ClayBenson94 (changes requested):
This looks good! Just might want a couple extra comments in a few spots, and then one other question around the Clearance strategy, but other than that I think this is good!

pkg/graph/resolvers/status_evaluation_strategies.go
ClayBenson94: Upon second review (thanks @OddTomBrooks for the heads up!) I think this is behaving as intended as-is!

To clarify: We only want to suggest the 4 clearance statuses if you're not in ANY clearance status, and this statement (`models.GetModelStatusChronologicalIndex(modelPlanStatus) < models.GetModelStatusChronologicalIndex(models.ModelStatusInternalCmmiClearance`) will do just that!

ClayBenson94: Since `ModelStatusInternalCmmiClearance` is the lowest clearance status, chronologically speaking.

ClayBenson94 (approved):

PR #1251, opened 2024-07-25
[MINT-2650]/model by solution bug fix

StevenWadeOddball:
# MINT-2650

## Description
Duplicate model plans were being returned to the front end for models that use a solution multiple times. To fix this the type needed to be updated.

As part of the work, a data loader for the operational solution by key was needed. Discussing with the front end team, it was determined we should return the possible operational solution instead of multiple solution instances.

Note, this dataloader has deviated somewhat from previous data loader implementations. This is to facilitate the eventual migration of the dataloaders to a different library.

## How to test this change
1. Bring up MINT.
2. Make a model plan
3. For multiple model plans, select the same solution for multiple needs.
4. For your homepage customization, ensure you only see one instance of that model plan 

(you can also achieve this by logging in as MINT, seeding the database, and updating a model plan that is already using 4 innovation to use it again).

## PR Author Checklist

## PR Reviewer Guidelines


ClayBenson94 (approved):

ClayBenson94:
Thanks Steven! 

My understanding on this makes is: if a model plan used the same operational solution multiple times, the join (Previously) would return multiple copies of the plan -- but now since it's joining on _possible_ solutions (which there's _never_ duplicates of) that isn't a problem

Is that right?

StevenWadeOddball:
> Thanks Steven!
> 
> My understanding on this makes is: if a model plan used the same operational solution multiple times, the join (Previously) would return multiple copies of the plan -- but now since it's joining on _possible_ solutions (which there's _never_ duplicates of) that isn't a problem
> 
> Is that right?

Thanks @ClayBenson94 ! That's 💯 % correct. We could have taken another approach and returned a list of solutions etc, but the solution data is not actually being used on the frontend at all. The group by clause no results in no duplicate models. 

This seemed a truer way to handle it than, returning only a representative solution.

PR #1266, opened 2024-07-31
[MINT-2998] Status route mutation handler

patrickseguraoddball:
# EASI-2998

## Description

- Added mutation handler to parse query param from url/email link and attempt to update status from task list
  - Error states for mutation and if status is already set
  - Success state
  - Clears param after attempt
 - Added default status on `<Status />` component tp take a default status for clearance status from email
- Added success message on next page/task list
- Added cypress test

## How to test this change

- Navigate to task list of a model plan
- Added query param `?model-status=ACTIVE` or any `ModelStatus` enum value.  This is case sensitive.  I decided to leave it case sensitive to stay true to enum value
- Verify status gets updated and alert renders
- Add a random string, verify error
- Add the existing status, verify specific error

## PR Author Checklist

## PR Reviewer Guidelines

### DIFF
--- a/src/views/ModelPlan/Status/index.tsx
+++ b/src/views/ModelPlan/Status/index.tsx
@@ -14,7 +14,6 @@ import {
 } from '@trussworks/react-uswds';
 import { ErrorMessage, Field, Form, Formik, FormikProps } from 'formik';
 import { ModelStatus, useUpdateModelPlanMutation } from 'gql/gen/graphql';
-import * as Yup from 'yup';
 
 import MainContent from 'components/MainContent';
 import PageHeading from 'components/PageHeading';
@@ -38,9 +37,6 @@ const Status = () => {
 
   const history = useHistory();
   const formikRef = useRef<FormikProps<StatusFormProps>>(null);
-  const validationSchema = Yup.object().shape({
-    status: Yup.string().required('Enter a role for this team member')
-  });
 
   const { status } = useContext(ModelInfoContext);
 
@@ -102,10 +98,6 @@ const Status = () => {
             initialValues={initialValues}
             enableReinitialize
             onSubmit={handleFormSubmit}
-            validationSchema={validationSchema}
-            validateOnBlur={false}
-            validateOnChange={false}
-            validateOnMount={false}
             innerRef={formikRef}
           >
             {(formikProps: FormikProps<StatusFormProps>) => {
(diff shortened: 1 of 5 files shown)


src/views/ModelPlan/TaskList/index.tsx
```diff
@@ -213,6 +231,77 @@ const TaskList = () => {
...
+              message: t('statusUpdateError', {
+                status: translatedStatus
+              }),
+              status: 'error'
+            });
+            params.delete('model-status');
+            history.replace({
+              search: params.toString()
+            });
+          });
```
garyjzhao, line 277: getting inaccurate error when i enter `?model-status=CANCEL`

image

garyjzhao (changes requested):
Found one error. see comment

src/views/ModelPlan/Status/index.tsx
```diff
@@ -29,44 +29,59 @@ type StatusFormProps = {
...
+        })
+        .catch(errors => {
+          formikRef?.current?.setErrors(errors);
+        });
+    }
   };
 
   const initialValues: StatusFormProps = {
-    status: status ?? undefined
+    status: modelStatus ?? status ?? undefined
```
garyjzhao, line 84: This does not seem right. What are you trying to do here?

patrickseguraoddball: Setting the default to the query param, then the query return, then string

src/views/ModelPlan/TaskList/index.tsx
patrickseguraoddball: CANCEL isn't a status

patrickseguraoddball: We don't have translations for CANCEL and nor would we really ever have that passed

garyjzhao: oh right. i was testing a status that does not exist.
Nevertheless, the error banner says `there was an error updating the status to .` <--

patrickseguraoddball: Right that will never happen, it also hasn't been mocked.  That text always runs through i18n translation.  I can ask Natasha but I dont thin kthis scenario would ever happen as it's coming from email link.  I'll see if I can get a general error from her.

patrickseguraoddball: @garyjzhao Just ended up adding the raw value if attempting to add something other than a set status

PR #1272, opened 2024-08-01
[MINT-3007] notification to model leads faktory boilerplate

StevenWadeOddball:
# MINT-3007

## Description

The goal of this PR is to provide the boiilerplate code for a new cron and batch job in faktory. This will be run once a day to suggest if a model plan should have an email sent to suggest an email be sent. It is expected that this code will be updated as the business logic around suggestions evolves.

Current assumptions:
1. There will be query that returns model plans that should be checked if a status update check should be run.
2. The ModelStatusUpdateJob will be updated with the business logic to actually evaluate and notify users.

Unit tests in this package assert that the parameters are provided as anticipated, and that the job can complete without error. Other unit tests should be made in the package which defined business logic

> [!NOTE]
> The jobs.md file can be updated to show the new work, but I think it's best to hold off until this is merged with the business logic in case the workflow changes.
## How to test this change

1.  Assert Unit Tests Pass
2. If desired, run `FAKTORY` locally (set FAKTORY_PROCESS_JOBS=TRUE)
3. visit http://localhost:7420/loops to assert that the new Model Status Update Cron Job is configured locally 
image
> [!NOTE]
> The actual cron job will have to be configured per environment. Ideally, this should run at the same time as the current daily digest job
## PR Author Checklist

## PR Reviewer Guidelines

### DIFF
--- a/pkg/worker/model_status_update_batch_job.go
+++ b/pkg/worker/model_status_update_batch_job.go
@@ -17,7 +17,7 @@ var modelStatusUpdateJobMaxRetry = 2
 // It will batch all child jobs, and when complete it will fire a callback
 func (w *Worker) ModelStatusUpdateBatchJob(ctx context.Context, args ...interface{}) error {
 	helper := faktory_worker.HelperFor(ctx)
-	sugaredLogger := w.Logger.With(zap.Any("JID", helper.Jid()), zap.Any("BID", helper.Bid()), zap.Any(appSectionKey, faktoryLoggingSection))
+	sugaredLogger := w.Logger.With(zap.Any("JID", helper.Jid()), zap.Any(appSectionKey, faktoryLoggingSection))
 	sugaredLogger.Info("Getting collection of model plans that require status checking")
 
 	// TODO: Implement the logic to return the models to check? Or do we check every model plan?
@@ -36,7 +36,7 @@ func CreateModelStatusUpdateBatch(logger *zap.Logger, w *Worker, cl *faktory.Cli
 	batch.Description = "Check if model status should be updated"
 	batch.Success = faktory.NewJob(modelStatusUpdateBatchJobSuccessName)
 	batch.Success.Queue = criticalQueue
-	// TODO: verify this overides the parent BID
+
 	sugaredLogger := logger.With(zap.Any("BID", batch.Bid))
 
 	err := batch.Jobs(func() error {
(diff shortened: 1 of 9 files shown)


pkg/worker/model_status_update_batch_job.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+	return nil
+
+}
+
+func CreateModelStatusJobInBatch(logger *zap.Logger, w *Worker, batch *faktory.Batch, plan *models.ModelPlan) error {
+	sugaredLogger := logger.With(zap.Any("modelPlanID", plan.ID))
+	sugaredLogger.Info("creating job for model status update.")
+	job := faktory.NewJob(modelStatusUpdateJobName, plan.ID)
+	//TODO: verify, do we want to just use the critical queue? Or should we generate another queue for this? Or perhaps choose a different existing queue?
```
OddTomBrooks, line 65: I'm not sure about this either. @ClayBenson94 what do you think?

ClayBenson94: Do you have a list of default queues for this? I'd say I'd be tempted to reserve the critical queue pretty much for Change History (since it needs to run every minute)

If there's a queue just lower than critical (high?) then I think that'd suffice for this kind of stuff, until we figure out any other categories/priorities that make sense!

OddTomBrooks: It'd help me if we had a clear definition of what the priorities are intended to handle. I generally agree this feels like a standard task, not a critical one.

StevenWadeOddball: So! Currently using the critical queue 1. Cron Jobs (Analyze Audit Batch, Translate Audit Batch, This method)  2. Analyzed Audit Jobs all trigger on the critical queue.

The `auditTranslateQueue` is configured for the individual jobs in a Translate Audit Batch, (so it doesn't grab priority from cron jobs).

I do agree we probably need some consensus around queue priorities. This is the current queue priorities
```
	// pull jobs from these queues, in this order of precedence
	mgr.ProcessStrictPriorityQueues(criticalQueue, defaultQueue, auditTranslateQueue, emailQueue)
```

I'd suggest
1. critical queue --> Jobs triggered by cron Jobs (batch jobs?)
2. default --> ? (maybe standard jobs?)
3. auditTranslateQueue --> just audit translations
4. emailQueue --> just email jobs?

perhaps we can do a quick staybehind after standup to get consensus on this?

OddTomBrooks (approved):

pkg/worker/model_status_update_batch_job.go
StevenWadeOddball: Thank you both, as we discussed, we will leave this for now. I made a ticket to audit queues for translations (MINT-3036)

PR #1275, opened 2024-08-02
[MINT-3025] Removed status mutation handler from tasklist

patrickseguraoddball:
# MINT-3025

## Description

Status  status mutation handler was no longer needed on task list to automatically change status from email links.  Status page should already be set to handle default status selections with the appropriate query parmater

- Removed status mutation handler from tasklist

## How to test this change

1.  Verify no statuses get updated on task list with supplied query parameter
2. Verify and status set on the status form page via `?model-status={status}` query parameter, that that status is set as the default

## PR Author Checklist

## PR Reviewer Guidelines

### DIFF
--- a/src/views/ModelPlan/Status/index.tsx
+++ b/src/views/ModelPlan/Status/index.tsx
@@ -168,7 +168,10 @@ const Status = () => {
                     <div className="margin-top-6 margin-bottom-3">
                       <Button
                         type="submit"
-                        disabled={!dirty && !modelStatus}
+                        disabled={
+                          (!dirty && !modelStatus) ||
+                          (!dirty && modelStatus === status)
+                        }
                         className=""
                         onClick={() => setErrors({})}
                       >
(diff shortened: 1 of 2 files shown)


ClayBenson94 (dismissed):
Changes work great locally!

One minor `nit` which I'll leave up to you if it's worth the effort: Taking a model plan that's already in a certain status and navigating to the /status route with the matching `?model-plan` still seems to highlight the `Update Status` button, even though no update would be needed

This slightly contrasts the behavior seen when just clicking the statuses manually (but like I said... pretty minor!)
Kapture 2024-08-05 at 09 03 50

Kapture 2024-08-05 at 09 02 08

patrickseguraoddball:
> Changes work great locally!
> 
> One minor `nit` which I'll leave up to you if it's worth the effort: Taking a model plan that's already in a certain status and navigating to the /status route with the matching `?model-plan` still seems to highlight the `Update Status` button, even though no update would be needed
> 
> This slightly contrasts the behavior seen when just clicking the statuses manually (but like I said... pretty minor!) !

@ClayBenson94  Thanks for catching this, I'll get it updated!

garyjzhao (approved):
Working as expected. 
Model status no longer changes outside of the `Update status` page. 
Updating the model in the status page will update the dropdown

PR #1281, opened 2024-08-05
[NOREF]change history job logging enhancements

StevenWadeOddball:
# NOREF

## Description
- Add utility methods to decorate faktory jobs with common fields.
- increased calls to logger.error where it makes sense, so we can trace the errors in SPLUNK
- add standard consts to access faktory fields.
- Unit tests to assert zap.With functionality (namely, that you can duplicate fields, so we should be careful when decorating a logger)

## How to test this change

1.  Look at the logging statements
2. Ensure unit tests pass

## PR Author Checklist

## PR Reviewer Guidelines


pkg/translatedaudit/translated_audit.go
```diff
@@ -50,14 +50,19 @@ func TranslateAudit(
...
+		err = fmt.Errorf("issue translating audit. %w", err)
+		logger.Error(err.Error(), zap.Error(err))
+		return nil, err
 	}
 
 	retTranslatedChanges, err := saveTranslatedAuditAndFields(store, translatedAuditWithFields)
 	if err != nil {
-		return nil, fmt.Errorf("issue saving translated audit. %w", err)
+		err = fmt.Errorf("issue saving translated audit. %w", err)
+		logger.Error(err.Error(), zap.Error(err))
```
ClayBenson94, line 64: I think there's pros and cons to wrapping chained errors with `fmt.Errorf` that I'd like to discuss here!

### Wrapping with `fmt.Errorf`
- Any single `err` object has the context of many errors beneath it, meaning a single log statement can log the error and you can see an effective error chain
  - This seems primarily useful in errors that we want to wrap, but _not_ log (since we want to include contextual information in the error, and probably are expecting some caller to log out the `err` object)

### Logging plain text in `logger.Error` calls
- Means that a single `err` object only contains the "triggering" error, and not contextual information of each function that it passed through / wraps it.
  - This seems like a simpler approach, and relies on the fact that we call `logger` statements multiple times throughout the call chain to allow us to view debug information.

I can't say I have a _strong_ preference one way or the other, as I think there's pros/cons both ways, but something about not having to wrap errors with `fmt.Errorf` every time, and instead just being able to do something like below speaks to a simple / repeatable approach:

```suggestion
		logger.Error("issue saving translated audit", zap.Error(err))
```

This way, any logs up the call chain aren't going to log this same message multiple times, but instead we can just search by a common `traceID` to see the unique list of errors (and where they got logged)

pkg/translatedaudit/translated_audit_job_utitilities.go
```diff
@@ -26,7 +27,7 @@ func TranslateAuditJobByID(ctx context.Context, store *storage.Store, logger *za
 
 	queueEntry, err = TranslatedAuditQueueUpdate(store, logger, queueEntry, constants.GetSystemAccountUUID())
 	if err != nil {
-		logger.Warn(err.Error(), zap.Error(err))
+		logger.Error(err.Error(), zap.Error(err))
```
ClayBenson94, line 30: Similar point to above -- I think including context-appropriate (and human readable) error text for our first parameter to `logger` calls is important!
```suggestion
		logger.Error("error updating translated audit queue during audit job", zap.Error(err))
```

pkg/worker/job_utilities.go
```diff
@@ -4,12 +4,77 @@ import (
...
+
+// TranslatedAuditQueueIDZapField provides the zap field for a TranslatedAuditQueueID
+func TranslatedAuditQueueIDZapField(translatedAuditQueueID uuid.UUID) zapcore.Field {
+	return zap.Any(translatedAuditQueueIDKey, translatedAuditQueueID)
+}
+
+// auditChangeIDZapField provides the zap field for an audit change id
+func auditChangeIDZapField(changeID interface{}) zapcore.Field {
+	return zap.Any(auditChangeKey, changeID)
+}
```
ClayBenson94, line 57: I like the idea behind these functions (presumably to allow for consistent search/query in Splunk), but it does seem to result in some really long / difficult to read logger statements (when I'd prefer to try and keep them simple)

```go
logger.With(TranslatedAuditQueueIDZapField(queueObj.ID), auditChangeIDZapField(queueObj.ChangeID), auditQueueAttemptsZapField(queueObj.Attempts))
```

Especially since some of these field names aren't reference many time (e.g. `auditQueueAttemptsZapField`), it feels like it's adding a fair bit of bloat compared to just referencing infrequently used keys in plain text:

```go
logger.With(
    zap.String("queue_id", queueObj.ID),
    zap.Any("change_id", queueObj.ChangeID,
    zap.Int("queue_attempts", queueObj.Attempts)
)
```

Could also just be worth simplifying the approach by introducing a `package logfields` and exporting commonly used keys there:


```go
logger.With(
    zap.String(logfields.UserID, userID), // or even logfields.UserID(userID)
    zap.Any("change_id", queueObj.ChangeID,
    zap.Int("queue_attempts", queueObj.Attempts)
)
```

ClayBenson94 (changes requested):
Left a few areas that I think we could continue to discuss!

I also could maybe spend some more time with the new faktory-specific logger helpers (`loggerWithFaktoryFieldsWithoutBatchID`, `loggerWithFaktoryFields`, and `loggerWithFaktoryStandardFields`) to better understand their use cases, as I am finding myself a bit lost when it comes to using `loggerWithFaktoryFieldsWithoutBatchID` and then later adding the `logger.With(batchID)` compared to when we just call `loggerWithFaktoryFields` and include the batch ID immediately

pkg/worker/job_utilities.go
StevenWadeOddball: I think a package could be a good idea, I think that I've introduced some keys in another package, and having a consistent shareable location would be more ideal.

I'll move them to their own package. I think I'd like to try the constructor functions there like `logfields.UserID(userID)`. I think spacing the with statement will help, as well as name spacing the log fields for context.

```go
logger.With(
      TranslatedAuditQueueIDZapField(queueObj.ID), 
      auditChangeIDZapField(queueObj.ChangeID), 
      auditQueueAttemptsZapField(queueObj.Attempts),
)
```

pkg/translatedaudit/translated_audit.go
StevenWadeOddball: That's fair, I think things like `sql` errors, or anything else the app should be contextually aware of are still wise to wrap.   

I do like the simplicity of passing the unwrapped error up the chain, and logging at each spot, but I'm a little hesitant about giving up the context you gain by wrapping at each spot. You can still get the context through logs, but the error is not interpreted throughout the code. 🤔 

I largely did this because I wanted to reuse the error message, while still logging and wrapping the message. That just might not be needed 🤔 per your point. But if we wanted to do both, we could have a  helper wrapper

```
func WrapAndLogError(logger *zap.Logger, errMessage string, err error) error {
	logger.Error(errMessage, zap.Error(err))
	return fmt.Errorf(errMessage+" err: %w", err)
}
```
so the code would simplify to 
```suggestion
		return nil, WrapAndLogError(logger, "issue translating audit.", err)
```

I'm not sure how important the additional wrapping actually is. This approach would still result in duplicate messages, but the initial error would still be the base error, without any wrapping. What do you think?

ClayBenson94: I think it might be good to try and simulate / see what these errors look like when they're wrapped multiple levels deep - my main concern is that we'll have really long and complex `err` messages in Splunk that will be difficult to parse.

I think the main benefit of wrapping errors would be if we plan to do anything with the error other than log it, but since we're taking the approach of logging freely within each function/level of the code, I think it sort of removes the benefit of this.

If we were wrapping errors and handing them up the chain for some middleware to log, I'd be all for it, but I think it might cause some extra bloat in our logs that might make them harder to parse through.

@StevenWadeOddball perhaps generating an error in one of these cases and outputting the logs would be a good way to get an idea of what they'll end up looking like?

StevenWadeOddball: I think that's a good call. I made this on the go playground to simulate.
https://go.dev/play/p/YO8hzohAhvP. I think we can revisit wrapping if we need to, but logging and passing the message along seems good, especially since we have a trace_id in the request. I think that it would be worthwhile to add a trace id to faktory jobs too. We have a job id, but that is shared across retries. A trace_id could be used to distinguish between instances of job runs.

ClayBenson94: That's incredibly helpful to see! That final example's `error` property is what I was thinking would be somewhat difficult to parse
```json
{
  "level": "error",
  "msg": "Model plan calculate status failed",
  "error": "Unable to save model plan err: the get statement failed err: pq: unable to parse parameters err: this is an error"
}
```

I'm good to re-visit wrapping as a concept at a later date too! (Wrapping can also be useful when attempting to determine typed errors as well, so maybe we could look at it if/when we look at `package apperrors`!

ClayBenson94 (approved):

ClayBenson94 (approved):

PR #1283, opened 2024-08-06
[MINT-2637] Handle sending email with no recipients

ClayBenson94:
# MINT-2637

## Description

> [!important]
> I think MINT-2637 calls out specifically handling the "Model Plan Created" notification being the source of this, but per my testing instructions, this could actually happen in a number of different cases, so I opted to handle (more generically) _any_ case where we attempt to send an email with no recipients.

- Updates go_simple_mail_service.go to un-export the `SendEmail` function (in favor of leaving `Send` exported instead, since that was the only one that was used outside the package anyways)
- Updates `sendEmail` to handle checking the length of the recipient list before sending.
  - I opted for this approach since it felt a bit cleaner than checking the [returned error's](https://github.com/xhit/go-simple-mail/blob/3161d31a9a885a33b09566f408732f39624dc4ec/email.go#L785-L787) text

This should help reduce (or completely eliminate) these kinds of errors:
image

## How to test this change

1.  Start the app with `scripts/dev up`
2. Open your notification preferences and ensure you're _not_ subscribed to the notifications that send when model plans have their dates changed.
3. Modify the `Plan Basics` of a new model plan
4. Ensure that there are no errors in the logs related to emails!

## PR Author Checklist

## PR Reviewer Guidelines

### DIFF
--- a/pkg/shared/oddmail/go_simple_mail_service.go
+++ b/pkg/shared/oddmail/go_simple_mail_service.go
@@ -98,11 +98,11 @@ func (g GoSimpleMailService) Send(
 		return err
 	}
 
-	return g.SendEmail(email)
+	return g.sendEmail(email)
 }
 
-// SendEmail is a GoSimpleMail specific method allowing for dispatching an email using a mail.Email object
-func (g GoSimpleMailService) SendEmail(email *mail.Email) error {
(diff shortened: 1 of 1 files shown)


OddTomBrooks (approved):
This is a very straightfoward and appropriate solution to the problem. Well done.

PR #1284, opened 2024-08-06
[MINT-1905] Notification to model leads to update status

OddTomBrooks:
# MINT-1905

## Description

This PR contains the combined functionality between myself and @StevenWadeOddball to send automated model phase suggestion emails on a daily Faktory job with some rule constraints.

## How to test this change

1. Run unit tests and validate their effectiveness
2. Execute local Faktory jobs based on 5 minute crons to assert successful email dispact to MailCatcher based on model lead emails.

## PR Author Checklist

## PR Reviewer Guidelines


pkg/graph/resolvers/model_plan_phase.go
StevenWadeOddball, line 1: Can you add comments to these new functions please?

StevenWadeOddball (commented):
Thanks Tom, this is looking good! I'm not completely done with the review, but I wanted to give you some feedback you could work on while I was reviewing still. I verified that the job runs as expected, and I got a suggestion for the Enhancing Oncology Model when I seeded data

pkg/graph/resolvers/model_plan_phase.go
```diff
@@ -0,0 +1,245 @@
...
+		phaseSuggestion := strategy.Evaluate(modelStatus, planBasics)
+		if nil != phaseSuggestion {
+			return phaseSuggestion, nil
+		}
+	}
+
+	return nil, nil
+}
+
+func ShouldSendEmailForPhaseSuggestion(
```
StevenWadeOddball, line 58: Can you add some comments around the function and the logic please?

pkg/graph/resolvers/model_plan_phase.go
```diff
@@ -0,0 +1,245 @@
...
+	currentPhaseSuggestion *model.PhaseSuggestion,
+	previousSuggestedPhase *models.ModelPhase,
+) bool {
+	if currentPhaseSuggestion == nil {
```
StevenWadeOddball, line 62: What does currentPhase Suggestion == nil mean? Does that mean that it is in the suggested phase already? Could you perhaps adds some comments to clarify that here and in the EvaluateSuggestedStatus function?

pkg/graph/resolvers/model_plan_phase_test.go
```diff
@@ -0,0 +1,179 @@
...
+		collaboratorInput,
+		suite.testConfigs.Principal,
+		false,
+		userhelpers.GetUserInfoAccountInfoWrapperFunc(suite.stubFetchUserInfo),
+		false,
+	)
+	suite.NoError(err)
+
+	//Asset that making a collaborator also creates an account
+	account, uAccountErr := storage.UserAccountGetByUsername(suite.testConfigs.Store, collaboratorInput.UserName)
```
StevenWadeOddball, line 50: Is this needed to be tested here? I think this might be more appropriate in the plan collaborator resolver test than here.

pkg/graph/resolvers/model_plan_phase_test.go
```diff
@@ -0,0 +1,179 @@
...
+		EXPECT().
+		Send(
+			gomock.Any(),
+			gomock.Eq([]string{expectedEmail}),
+			gomock.Any(),
+			gomock.Eq(expectedSubject),
+			gomock.Any(),
+			gomock.Eq(expectedBody),
+		).
+		AnyTimes()
```
StevenWadeOddball, line 80: Should this instead assert a specific number of times we expect an email to be sent. Any asserts it is called 0 or more times, so I don't think it's providing the test we want here.

pkg/graph/resolvers/model_plan_phase_test.go
```diff
@@ -0,0 +1,179 @@
...
+	mockEmailService := oddmail.NewMockEmailService(mockController)
+	mockEmailTemplateService := email.NewMockTemplateService(mockController)
+
+	planName := "Plan For Milestones"
+	plan := suite.createModelPlan(planName)
+
+	collaboratorInput := &model.PlanCollaboratorCreateInput{
+		ModelPlanID: plan.ID,
+		UserName:    "CLAB",
+		TeamRoles:   []models.TeamRole{models.TeamRoleLeadership},
```
StevenWadeOddball, line 28: The AC for the user story says it should go to `model_leads`, not to `leadership`, are you trying to assert this user shouldn't get an email?

StevenWadeOddball: the model lead here would be whoever created the model. (maybe some comments would be helpful)

pkg/worker/model_status_update_batch_job.go
```diff
@@ -0,0 +1,86 @@
...
+// ModelStatusUpdateBatchJob is the job the cron job initiates to check if models need a status update.
+// It will batch all child jobs, and when complete it will fire a callback
+func (w *Worker) ModelStatusUpdateBatchJob(ctx context.Context, args ...interface{}) error {
+	helper := faktory_worker.HelperFor(ctx)
+	// decorate the logger, but exclude the bid, the bid will be decorated when we create the batch
+	logger := loggerWithFaktoryFieldsWithoutBatchID(w.Logger, helper)
+	logger.Info("Getting collection of model plans that require status checking")
+
+	// TODO: Implement the logic to return the models to check? Or do we check every model plan?
+	modelsToUpdate, err := w.Store.ModelPlanCollection(logger, false)
```
StevenWadeOddball, line 26: @OddTomBrooks, do you know if there are any models that shouldn't be checked? This code chunk currently only looks at active models, which makes sense to me. Are there any statuses we don't want to check for (so to save some faktory bandwidth)?

StevenWadeOddball (changes requested):
Thanks @OddTomBrooks ,

I checked that the emails send as expected. I requested some changes that can hopefully give a more confidence to some parts of the code, as well as some areas where reimplementing the data loader can increase some efficiency.

pkg/graph/resolvers/model_plan_phase.go
OddTomBrooks: A nil return type simply means that, for whatever reason, there is no currently suggested phase according to the strategies provided in the strategy pattern.

OddTomBrooks: I have added a comment at the function level to document this

pkg/graph/resolvers/model_plan_phase_test.go
OddTomBrooks: I thought about this for a while. I think I want to leave the assertion here not  because of the flow assertion but because if things break we understand why. You are correct that the core test for this feature set should be in plan collaborator.

pkg/worker/model_status_update_batch_job.go
OddTomBrooks: There are certainly ways to optimize this query down, for instance we could make a SQL query that selects statuses where not in the excluded states. For instance, this logic could be excluded entirely by a specialized SQL query:

`
	// If the model plan is paused or canceled, we shouldn't suggest a new phase
	if modelStatus == models.ModelStatusPaused || modelStatus == models.ModelStatusCanceled {
		return nil, nil
	}
`

That being said, this is a pre-optimization. It's dubious if we would even have an improvement depending on caching behavior. More importantly, the dev cost to build this versus spending time elsewhere is probably not worth the investment. I suggest that if this becomes problematic in terms of performance as part of a profiling pass that we revisit it. Profiling is something I hope to get to later in this sprint if I have the capacity.

StevenWadeOddball: That works for me, maybe it's worthwhile to note in the code that this could be revisited. I agree it's likely not a huge impact.

PR #1307, opened 2024-08-21
[NOREF] Add script to only allow adding migrations

ClayBenson94:
# NOREF

## Description

Improves the CI/CD check for migrations to also disallow modifications to migrations (as we shouldn't be allowed to modify, rename, or remove migrations from this folder)

## How to test this change

1. TBD

## PR Author Checklist

## PR Reviewer Guidelines


StevenWadeOddball (approved):
Thanks Clay,

I tested this by modifying the script to look at local changes,  (remove origin),
then modified a script in a commit. I then ran
BASE_REF=main HEAD_REF=NOREF/migration_cicd_checker ./scripts/check_migration_diff.sh

And received this output
```
Migration diff:
M       migrations/V170__New_View_Customization_Options.sql
Migration diff (non-adds):
M       migrations/V170__New_View_Customization_Options.sql
Migration files can only be added in migrations folder
```

I also added a new migration and got this output
```
bash-5.2$ BASE_REF=main HEAD_REF=NOREF/migration_cicd_checker ./scripts/check_migration_diff.sh
Migration diff:
A       migrations/V174__Update_CPI_Vetting_Operational_SolutionNEWNEWNEW.sql
Migration diff (non-adds):
```

so this looks good to go! Thanks @ClayBenson94

PR #1310, opened 2024-08-21
[MINT-1905] [MINT-3068] Address backend findings with automatic status update email

ClayBenson94:
# MINT-1905 / MINT-3068

## Description

Addresses the following changes:
- [x] Model Plan Suggested Status emails do not send to model leads
- [ ] Model Plan Suggested Status emails send repeatedly for the same suggested status

## How to test this change

1.  

## PR Author Checklist

## PR Reviewer Guidelines

### DIFF
--- a/pkg/graph/resolvers/model_plan_phase.go
+++ b/pkg/graph/resolvers/model_plan_phase.go
@@ -3,7 +3,9 @@ package resolvers
 import (
 	"context"
 	"fmt"
+	"time"
 
+	"github.com/cmsgov/mint-app/pkg/authentication"
 	"github.com/cmsgov/mint-app/pkg/email"
 	"github.com/cmsgov/mint-app/pkg/graph/model"
 	"github.com/cmsgov/mint-app/pkg/models"
@@ -80,6 +82,8 @@ func ShouldSendEmailForPhaseSuggestion(
 // TrySendEmailForPhaseSuggestion sends an email to the model plan leads if the suggested phase has changed
 func TrySendEmailForPhaseSuggestion(
 	logger *zap.Logger,
+	store *storage.Store,
+	principal authentication.Principal,
 	emailRecipients []string,
 	emailService oddmail.EmailService,
 	emailTemplateService email.TemplateService,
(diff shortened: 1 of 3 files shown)


pkg/worker/model_status_update_job.go
```diff
@@ -33,9 +36,16 @@ func (w *Worker) ModelStatusUpdateJob(ctx context.Context, args ...interface{})
...
+	// Ensure there's a context with the user account service
+	// TODO THis is a bit hacky -- we should probably be doing this on every job, not just this one, but we needed to get this working in MINT-3068
+	dataLoaders := loaders.NewDataLoaders(w.Store)
+	ctx = loaders.CTXWithLoaders(ctx, dataLoaders)
+	ctx = appcontext.WithUserAccountService(ctx, userhelpers.UserAccountGetByIDLOADER)
+
 	return resolvers.TrySendEmailForPhaseSuggestionByModelPlanID(
 		ctx,
 		w.Store,
+		appcontext.Principal(ctx),
```
ClayBenson94, line 48: Since this is running in a Faktory job, I'm not actually sure what `appcontext.Principal(ctx)` will return here.

I think it might make sense to see if we can avoid touching the `modifiedDts`/`modifiedBy` fields here if possible, since, at least from a users/auditing perspective, we probably don't want to show this as an update being performed by someone.

Does the audit trigger complain if we don't set `modifiedDts`/`modifiedBy`? Perhaps we could add this column to the `ignored_fields` for this table?

pkg/graph/resolvers/model_plan_phase.go
```diff
@@ -115,7 +119,24 @@ func TrySendEmailForPhaseSuggestion(
 		logger.Error(err.Error(), zap.Error(err))
 	}
 
-	return err
+	if currentPhaseSuggestion == nil {
```
ClayBenson94, line 122: Do you know what case we'd call this with a `nil` `currentPhaseSuggestion`?

Wondering if it might be safer to just exit early in this case rather than clearing the suggestion.

pkg/worker/model_status_update_job.go
ClayBenson94: Yeah, running this locally seems to get a nil pointer dereference in the logs and the job panics (and retries) -- I think we might need to update this to work around needing to set those fields

pkg/graph/resolvers/model_plan_phase.go
OddTomBrooks: This is actually not achievable due to an earlier escape condition. I added this mostly to prevent a linter warning.

ClayBenson94: Makes sense -- maybe we can just replace 
```go
		modelPlan.PreviousSuggestedPhase = nil
```
with
```go
		return nil
```

if that works for you?

OddTomBrooks: Changes are in

OddTomBrooks (approved):

PR #1313, opened 2024-08-23
[NOREF] Fix the loading spinner in Model Plan Card in Collaboration Area

garyjzhao:
# NOREF

## Description

* Add some margin to the spinner so that it does not spill over

| Before       | After       |
| -------------- | -------------- |
| Screenshot 2024-08-23 at 1 56 10 PM| image |

## How to test this change

1.  Verify that the spinner's spacing in not off the grid
2. 

## PR Author Checklist

## PR Reviewer Guidelines


patrickseguraoddball (changes requested):
I was thinking we should center it with the card vertically and horizontally, something like this: 

```
  if (loading && !modelPlan)
    return (
      <Grid
        desktop={{ col: 6 }}
        className="padding-1 display-flex flex-column flex-align-center flex-justify-center height-mobile"
      >
        <Spinner />
      </Grid>
    );
```

Also for the loading conditional we might want to add 
`  if (loading && !modelPlan)` to allow the cached data to render if it exists

patrickseguraoddball (approved):
Thanks!

PR #1335, opened 2024-08-29
[MINT-3078] Read View Scrolling Bug

garyjzhao:
# MINT-3078

## Description

When a user is viewing the read-view of a model plan and move from Model basics to General characteristic section the page jumps to the top and not maintaining the scroll position

### Expected behavior (attached an image of the expected behavior)

The page wouldn't jump to the top when changing tabs. It should only scroll to the top of that particular section.

### Steps to reproduce the issue

1. Login into MINT
2. Navigate to the read-view of a model plan
3. Click on Model basics then go to General characteristics --> this should produce the bug

## How to test this change

1. Navigate to the read-view of a model plan
3. Click on Model basics then go to General characteristics
4. Click through all sections and verify that scrolling to the top of the Read Only body content if user is scrolled pass the filter bar

## PR Author Checklist

## PR Reviewer Guidelines

### DIFF
--- a/src/views/ModelPlan/ReadOnly/_components/Sidenav/index.tsx
+++ b/src/views/ModelPlan/ReadOnly/_components/Sidenav/index.tsx
@@ -55,13 +55,15 @@ const SideNav = ({
 
       // Calculate all the things
       const distanceFromTopOfPage =
-        top +
+        Math.round(top) +
         window.scrollY -
         filterBannerHeight -
         navBarHeight -
         marginOfScrollElement;
 
-      window.scroll(0, distanceFromTopOfPage);
+      if (window.scrollY >= distanceFromTopOfPage) {
+        window.scroll(0, distanceFromTopOfPage);
+      }
     }, 0);
   };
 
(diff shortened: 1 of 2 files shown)


src/views/ModelPlan/ReadOnly/_components/Sidenav/index.tsx
```diff
@@ -55,13 +55,15 @@ const SideNav = ({
...
       const distanceFromTopOfPage =
-        top +
+        Math.round(top) +
         window.scrollY -
         filterBannerHeight -
         navBarHeight -
         marginOfScrollElement;
 
-      window.scroll(0, distanceFromTopOfPage);
+      if (window.scrollY >= distanceFromTopOfPage) {
```
ClayBenson94, line 64: Might just be useful to slap a comment on this check explaining what it does (I assume this is basically checking if the view has been scrolled _past_ the `distanceFromTopOfPage`, i.e. if you've scrolled past the header, navbar, etc.?

patrickseguraoddball (approved):
Thanks for adding the comment, this worked as expected for me

PR #1344, opened 2024-09-05
[Mint 3067]/be analytics queries

StevenWadeOddball:
# MINT-3067

## Description

This PR's goal is to create some basic queries to expose some  queries to provide basic analytics about  MINT data. Some data will need to be further modified as we proceed to fine tune it to the final tool we are using. This however will provide a starting point to answer the questions being asked.

You will note there is not a date rang on these queries currently, but look at data for all time. Also, many of these queries might be more useful when constrained to a single model at a time, but it is possible to run across the entire data set.

## How to test this change

1.  Run each query against the database.
2. Ensure that the queries function without error.

If desired, you can run this against a fuller data set. At the very least, I suggest that you seed the database before testing the queries

## PR Author Checklist

## PR Reviewer Guidelines


pkg/sqlqueries/SQL/analytics/number_of_followers_per_model.sql
```diff
@@ -0,0 +1,9 @@
+/* Number of Followers By Model */
+SELECT
+    model_plan.model_name,
```
ClayBenson94, line 3: Would you mind adding `model_plan.id` here as well like we do in the other queries?

ClayBenson94 (changes requested):
Looks good! Just one nit-pick on adding an extra field, but otherwise I'm ready to approve!

pkg/sqlqueries/SQL/analytics/number_of_followers_per_model.sql
StevenWadeOddball: Thanks @ClayBenson94 ! Added

ClayBenson94 (approved):
Changes look good -- thanks @StevenWadeOddball!

PR #1355, opened 2024-09-11
[MINT-3084] Updated 4i and aco-os point of contact info

OddTomBrooks:
# MINT-3084

## Description

- Added migration to update content according to Jira and Figma specification

## How to test this change

1. Bring up containers
2. Open 4i and ACO-OS operational solution help cards
3. View points of contact
4. Compare against Figma

## PR Author Checklist

## PR Reviewer Guidelines


StevenWadeOddball (approved):
Thanks @OddTomBrooks , approving based on functionality.

Small nitpick, you can adjust, or just proceed with the merge. I think it's best practice not to reference the the possible_solution by id, just for clarity of the migration. Since they happen to be the same in all environments, this works, but there is a bit of mystery reading the migration. Joining to the possible solution, and using the key would increase clarity.

OddTomBrooks:
> Thanks @OddTomBrooks , approving based on functionality.
> 
> Small nitpick, you can adjust, or just proceed with the merge. I think it's best practice not to reference the the possible_solution by id, just for clarity of the migration. Since they happen to be the same in all environments, this works, but there is a bit of mystery reading the migration. Joining to the possible solution, and using the key would increase clarity.

Thanks for bringing this up, I did consider it and agree but like you said due to the homogeneity of the environment I felt it was better to keep the query simple and have a faster turnaround on the ticket. As an intermediary solution, I have added a table comment which addresses the clarity issue without adding logical complexity.

PR #1361, opened 2024-09-13
[MINT-3099][MINT-3100]/cr tdl echimp data

StevenWadeOddball:
# MINT-3099, [MINT-3100](https://jiraent.cms.gov/browse/MINT-3100)

## Description

1. Introduce new echimp CR and TDL data type
2. Introduce new seed bucket for local CR / TDL data (more to come later)
   a. Note that there is not a model plan associated with the seeded data currently, that is a more involved effort
4. Rudimentary cache with expiration
5. parsing of flags into bools 
6. Postman collection updates

## How to test this change

1.  Seed the database
2. Use the postman collection to view available data.
3. Ensure \N is removed from text. 
4. Ensure that html is removed from all text.
5. Ensure that the postman collection provides the data in a format needed for the frontend
    a. note that there is not yet seeded model plans corresponding to the test data, as it involves reworking the seeding process.

## PR Author Checklist

## PR Reviewer Guidelines


pkg/echimpcache/echimp_cache.go
```diff
@@ -0,0 +1,192 @@
...
+
+	"github.com/cms-enterprise/mint-app/pkg/models"
+	"github.com/cms-enterprise/mint-app/pkg/parquet"
+	"github.com/cms-enterprise/mint-app/pkg/s3"
+)
+
+// echimpCacheTimeHours is the length of time before a echimpCache needs to be refreshed
+const echimpCacheTimeHours = 3
+const crKey = "echimp_cr"
+const tdlKey = "echimp_tdl"
```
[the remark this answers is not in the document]
ClayBenson94: I think that might actually be the case, since it looks like `package echimptestdata` codes in the same thing.

We don't have to address it as part of this PR if you're able to track this as a TODO with a comment -- but I think making this a configurable path is probably in our best interest!

pkg/graph/resolvers/resolver.go
```diff
@@ -25,7 +25,8 @@ import (
 type Resolver struct {
 	store                *storage.Store
 	service              ResolverService
-	s3Client             *upload.S3Client
+	s3Client             *s3.S3Client
```
ClayBenson94, line 28: Might want to flag this for a potential rename?
```suggestion
	s3Client             *s3.S3Client // TODO Maybe consider renaming to something like "fileUploadS3Client"
```

pkg/graph/schema/types/echimp.graphql
```diff
@@ -0,0 +1,45 @@
...
+
+
+
+
+extend type Query {
+  echimpCR(id: String!): EChimpCR!
+  echimpCRs: [EChimpCR!]!
+  echimpTDLs: [EChimpTDL!]!
+  echimpTDL(id: String!): EChimpTDL!
+  echimpCRAndTDLS: [EChimpCRAndTDLS!]!
```
ClayBenson94, line 44: I might suggest against exposing CRs and TDLs outside the scope of a model plan until we have a use case for it -- I think the new properties on `type ModelPlan` should suffice

ClayBenson94: We could leave the "fetch by ID" ones in here if that'd be helpful on the FE, though! CC @patrickseguraoddball @garyjzhao

ClayBenson94 (changes requested):
Throwing in a request for a few changes! I haven't finished reviewing the entire PR, but wanted to get feedback in sooner rather than later!

@StevenWadeOddball if you think something could/should be saved for later, feel free to just let me know!

pkg/graph/schema/types/echimp.graphql
StevenWadeOddball: Thanks @ClayBenson94 , I'm pretty sure from the FIGMA that there is a need to show all CRs and TDLS, not just ones associated to a model plan (that why I have these routes exposed. Do you see that in the figma as well?

StevenWadeOddball: I added fetch by ID just to support the possibility of the CR / TDL model appearing as a route (without needing to fetch all the records). It was vaguely talked about as a possibility, so I thought it might be better just to include it

pkg/appconfig/config.go
```diff
@@ -173,6 +173,12 @@ const ClientAddressKey = "CLIENT_ADDRESS"
 // AWSS3FileUploadBucket is the key for the bucket we upload files to
 const AWSS3FileUploadBucket = "AWS_S3_FILE_UPLOAD_BUCKET"
 
+// AWSS3ECHIMPBucket is the key for the bucket we EChimp data files are located
+const AWSS3ECHIMPBucket = "AWS_S3_ECHIMP_BUCKET"
+
+// LDEChimpEnabledKey is the key for retrieving the bool if echimp is enabled to grab from S3
+const LDEChimpEnabledKey = "echimp-enabled"
```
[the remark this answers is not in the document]
StevenWadeOddball: Thanks Clay, that makes sense to move it to this package! I don't see an option to edit this flags naming in LD, but if needed I could delete it and remake it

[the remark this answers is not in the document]
StevenWadeOddball: I updated it in [6eaf32b](https://github.com/CMS-Enterprise/mint-app/pull/1361/commits/6eaf32b752727724d09abdfc4cf569f08071ad53)

pkg/flags/client.go
```diff
@@ -76,3 +83,10 @@ func ContextKeyForID(id string) string {
 	_, _ = h.Write([]byte(id))
 	return fmt.Sprintf("%x", h.Sum(nil))
 }
+
+// GetBool fetches a bool from Launch Darkly
+func GetBool(principal authentication.Principal, ldClient *ld.LDClient, key string, defaultValue bool) (bool, error) {
```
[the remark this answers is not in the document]
StevenWadeOddball: [c814117](https://github.com/CMS-Enterprise/mint-app/pull/1361/commits/c814117cc280e6aae431da7fa1ef2ca3088c601e)

pkg/graph/resolvers/resolver.go
StevenWadeOddball: I like it, I went ahead and renamed it. There might be some test configs that we find later we want to rename as well later

pkg/echimpcache/echimp_cache.go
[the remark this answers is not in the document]
StevenWadeOddball: Thanks Clay, if you are ok with it, I think I might just leave it as a todo for now. The wiring for this would be a little convoluted with our current reliance on viper to read the env vars. 

Perhaps we can target this with some of the other changes we talked about for fetching configs? We might also consider making the eChimpCacheTime an envar too, in case we want to have varying times in diff environments

pkg/graph/schema/types/echimp.graphql
ClayBenson94: > Thanks @ClayBenson94 , I'm pretty sure from the FIGMA that there is a need to show all CRs and TDLS, not just ones associated to a model plan (that why I have these routes exposed. Do you see that in the figma as well?

I wasn't able to find that in the Figma myself, but I might be looking in the wrong spot. This won't really be harmful to introduce to just unblock the FE, but we might want to just reconsider it in the near future if it's not going to be needed!

StevenWadeOddball: Sounds good! This is the [area](https://www.figma.com/design/qXhO12xSSkDF4LaToGKJoV/MINT-Blue-Sky-Ideas?node-id=2697-35599&node-type=section&t=7fyKzjzrU9gs6tVn-0) I'm thinking of, but it might not need the separate CRs, and separate TDLs endpoint (perhaps just the union type).

pkg/graph/schema/types/echimp.graphql
```diff
@@ -0,0 +1,45 @@
+type EChimpCR {
+  crNumber: String!
+  versionNum: String!
+  initiator: String
+  firstName: String
+  lastName: String
+  title: String
+  sensitiveFlag: Boolean
+  implementationDate: String
+  crSummary: String
```
ClayBenson94, line 10: Follow up after this PR: Maybe make this HTML/TaggedContent?

pkg/parquet/reader.go
```diff
@@ -0,0 +1,54 @@
...
+	readerAt, size, err := client.GetS3ObjectReaderAt(key)
+	if err != nil {
+		return nil, err
+	}
+	data, err := parquet.Read[T](readerAt, size)
+	if err != nil {
+		return nil, err
+	}
+	return data, nil
+	// result := []*T{}
```
ClayBenson94, line 20: Follow up after this PR: Clear commented code

pkg/s3/download.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+func (c S3Client) DownloadStream() {
+
+}
+
+// TODO: (Parquet) Should we remove this func
+
+// DownloadToMemory downloads an object from the S3 bucket to a buffer in memory
+func (c S3Client) DownloadToMemory(key string) ([]byte, error) {
```
ClayBenson94, line 20: Follow up after this PR: Maybe delete this?

pkg/graph/schema/types/echimp.graphql
```diff
@@ -0,0 +1,45 @@
...
+# """
+# EChimpCRAndTDLS is a type that represents CRS and TDLS so they can be returned together in graphql
+# """
+union EChimpCRAndTDLS = EChimpCR  | EChimpTDL
+
+
+
+
+extend type Query {
+  echimpCR(id: String!): EChimpCR!
```
ClayBenson94, line 40: Follow up: add an ECHIMP URL resolver (and env var?)

pkg/graph/schema/types/echimp.graphql
```diff
@@ -0,0 +1,45 @@
...
+  implementationDate: String
+  crSummary: String
+  crStatus: String
+  emergencyCrFlag: Boolean
+  relatedCrNumbers: String
+  relatedCrTdlNumbers: String
+  associatedModelUids: UUID
+}
+
+type EChimpTDL {
```
ClayBenson94, line 18: Follow up: Add an ECHIMP URL resolver

ClayBenson94 (dismissed):

patrickseguraoddball (dismissed):
Thanks!

patrickseguraoddball (approved):

PR #1374, opened 2024-09-20
[MINT-3119] FE work for the Two Pager Help Article update

garyjzhao:
# MINT-3051, MINT-3119

## Description

- Update the artcile to match figma

## How to test this change

1.  Verify figma
2. Verify links

## PR Author Checklist

## PR Reviewer Guidelines


src/features/HelpAndKnowledge/Articles/TwoPagerMeeting/index.tsx
```diff
@@ -111,49 +138,166 @@ const TwoPagerMeeting = () => {
...
+                  }}
+                />
+              </p>
               <ul className="margin-top-0 margin-bottom-5 padding-left-6">
                 <li className="line-height-sans-4 margin-bottom-05">
-                  <strong>
-                    {twoPageMeetingT('conceptPaper.stepThree.alignment.bold')}
-                  </strong>
-                  {twoPageMeetingT('conceptPaper.stepThree.alignment.text')}
+                  <Trans
```
patrickseguraoddball, line 294: I think all of these list items where there are more than 2 can be iterated dynamically through the translations.  These are basically all the same components, just different text

src/features/HelpAndKnowledge/Articles/TwoPagerMeeting/index.tsx
```diff
@@ -176,108 +326,214 @@ const TwoPagerMeeting = () => {
...
+                    {twoPageMeetingT('about.summarybox.exampleSummaryBox.text')}
+                  </p>
+                  <ul className="margin-top-0">
+                    <li className="margin-bottom-1">
+                      <ExternalLink href="https://share.cms.gov/center/cmmi/SR/ModelDev/Submitted%20Two%20Pagers/AHEAD%20Concept%20Paper.docx?d=w9bd3973322384706a0207c756f773739">
+                        {twoPageMeetingT(
+                          'about.summarybox.exampleSummaryBox.list.one'
+                        )}
+                      </ExternalLink>
+                    </li>
```
patrickseguraoddball, line 419: More opportunity to loop through to render all of these

src/features/HelpAndKnowledge/Articles/TwoPagerMeeting/table.tsx
```diff
@@ -0,0 +1,71 @@
...
+const ModelSectionCriteriaTable = () => {
+  const headers = tArray('twoPageMeeting:reviewMeeting.criteria.table.header');
+
+  const content = tObject(
+    'twoPageMeeting:reviewMeeting.criteria.table.content'
+  ) as ContentTypes[];
+
+  return (
+    <TrussTable bordered={false} fullWidth fixed>
+      <thead>
```
patrickseguraoddball, line 24: Thanks for breaking this table out.  Let's sync when you return about a strategy for our PRs to be able to use all of these.  It might make sense to quickly break out our shared components and get a quick review on these so we can both incorporate them

patrickseguraoddball (changes requested):
I think the category page and work is missing, but don't think it was specifically covered in this story. I'll cover it in the six pager PR I'm working on if this gets merged first:
https://oddball.slack.com/archives/C059QQA3HK5/p1726851163650029

This code still has the legacy title
https://github.com/CMS-Enterprise/mint-app/blob/7d2a923394e7c5ad96ade32d019e8057e10b303d/src/i18n/en-US/helpAndKnowledge/helpAndKnowledge.ts#L1986

This one as well:
https://github.com/CMS-Enterprise/mint-app/blob/7d2a923394e7c5ad96ade32d019e8057e10b303d/src/i18n/en-US/routes.ts#L138

I have created reusuable components for the anchor links as well as the cards.  I think we should possibly wait for the PR with the reusable components before merging this one.

patrickseguraoddball (changes requested):
The spacing here is larger than in Figma.  The links are also a bit close together, but that should be covered I think in my PR if you want.

Screenshot 2024-09-24 at 3 00 58 PM

Clicking on the anchor links bring you below the intended html element rather than above.  This is something I also need to look into, but we should try to address this in you PR if possible.
Screenshot 2024-09-24 at 3 02 55 PM

Screenshot 2024-09-24 at 3 01 55 PM

I think the line-height on all the content is off.  I think we should create our own top level utility class to better match figma.  Im not seeing any utility class in USWDS that matches, but we need a `line-height: 25px` or `line-height: 1.5625rem` style not `line-height-sans-4`.  I've done this on my branch, I can update yours with the new style if you want to avoid conflict.
Screenshot 2024-09-24 at 3 04 41 PM
Basically added something like this to `custom.scss`
```
.line-height-normal {
  line-height: 1.5625rem !important;
}

.line-height-large {
  line-height: 2.1875rem !important;
}

.mint-body-normal {
  font-weight: 400 !important;
  line-height: 1.5625rem !important;
  font-size: 1rem !important;
}

.mint-body-large {
  font-weight: 300 !important;
  line-height: 2.1875rem !important;
  font-size: 1.375rem !important;
}
```
These related directly to the naming in Figma
Screenshot 2024-09-24 at 3 22 35 PM

Padding on these bullets in the table are a bit off
Screenshot 2024-09-24 at 3 24 17 PM

src/features/HelpAndKnowledge/Articles/index.tsx
```diff
@@ -92,14 +93,49 @@ export const covertToLowercaseAndDashes = (string: string) =>
...
+
+      // Find top margin of the heading element
+      const marginOfHeading = parseFloat(
+        window.getComputedStyle(headingElement).marginTop
+      );
+
+      window.scroll(0, top - marginOfHeading - navBarHeight);
+    }, 0);
+  };
+
```
garyjzhao, line 128: Taking much inspiration from the read view scroll function

patrickseguraoddball: Considering that all of our target anchors are `<h2 />` elements, we can add a class with a css property `scroll-margin-top`.  Those elements are `2rem` plus the `2rem` margin.  We can then give it maybe a little more padding a create a class like so and add it to the anchor elements:

```
.scroll-target {
    scroll-margin-top: 5rem; /* Adjust this value based on the anchor/h2/header height */
}
```

garyjzhao: Another thought I just had, should I move the covertToLowercaseandDashes to the `h2` element, instead of the div?

patrickseguraoddball (approved):

PR #1375, opened 2024-09-20
[MINT-3082] Okta session auto login

patrickseguraoddball:
# MINT-3082

## Description

- Added hook in `useOktaSession` to check if okta session exists via Okta api
- Implemented session check on `<Landing />`.  
  - If session exists, replace router link with button that calls `signInWithRedirect`
- Implemented session check on `<OktaSignInWidget />`.  
  - While checking session disable widget with an overlay and loading spinner
  - If session exists immediately `signInWithRedirect` to authenticate within app
- Unauthenticated redirects should still work as expected within the app
  - For example, hitting `/help-and-knowledge` while unauthenticated will redirect to `/signin` then check for existing okta session.  If exists, it will redirect you back to `/help-and-knowledge`
- Addressed existing flicker within initial app load of `<UserInfoWrapper />` by checking of `authState` has loaded.  
  - If there is no local `authState`, and an Okta session does not exist, render the `<Landing />` page.
- If user is logged out of their session in another environment - `https://test.mint.cms.gov/`, the user will not be logged out of MINT locally, as the `authState` still persists on the user's machine and takes priority over the Okta session
  - The same holds true if a user logs out of MINT locally, the test environment will remain logged in until logged out manually or app times out
- Sessions are shared, however each device has their own `authState`, which takes priority in the local app.
- After logging out, session will not be restored upon refresh as the session does not exist.  The user will need to use the widget login mechanism to start a new session
- Added a `display: none` class to `<LoginCallback />` component as to not flicker when redirecting.  We still want to process the component, just not render

## How to test this change

1.  Log out of MINT locally and in test env - `https://test.mint.cms.gov/`
2. Log into test env
3. Load the app/refresh MINT
4. Confirm that the app loads with the landing page.
5. Inspect the html to verify both sign in links on the landing page and header have been replace with buttons
6. Verify that clicking either will automatically authenticate the user
7. Log out of test
8. Verify user can still access MINT locally until manually signed out
9. Repeat the process but this time log out of MINT locally instead of test.
10. Verify that refreshes to no automatically sign the user in (as the session has been deleted)
11. Verify that test remains authenticated after refresh.
12. Repeat from step 1, except when attempting to hit the landing/home page, attempt to the the Help and Knoledge center directly
13. Verify redirect works after session check and authentication

## PR Author Checklist

## PR Reviewer Guidelines


src/components/OktaSignInWidget/index.scss
```diff
@@ -7,9 +7,42 @@
...
+  position: relative;
+}
+
+.nested-overlay {
+  position: absolute;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
+  background-color: rgba(0, 0, 0, 0.3);
```
ClayBenson94, line 43: Definitely a `nit` -- but would you be open to considering a lighter color for the overlay?

Here's what this color looks like (using a _more_ opaque white instead of a _less_ opaque black)
 
```suggestion
  background-color: rgba(255, 255, 255, 0.6);
```

Kapture 2024-09-24 at 09 05 41

src/components/Header/index.tsx
```diff
@@ -147,18 +155,38 @@ const Header = () => {
                   {t('landing:getAccess')}
                 </UswdsReactLink>
               )}
-              <Link
-                className={classnames(
-                  'mint-header__nav-link  radius-md padding-y-105 padding-x-2',
-                  {
-                    'text-white border': isLanding,
-                    'text-base-darker': !isLanding
+              {hasSession ? (
```
ClayBenson94, line 158: Could you add a quick comment/comments clarifying the purpose of this code?

I understand it to be: If a user has a session, the `Sign In` button should effectively use `oktaAuth.signInWithRedirect()` to skip straight to an authenticated page (`/pre-decisional-notice`), rather than just linking to the `/sign-in` route.

src/components/OktaSignInWidget/index.tsx
```diff
@@ -4,22 +4,29 @@ import React, { useEffect, useRef } from 'react';
...
+import Spinner from 'components/Spinner';
+import useOktaSession from 'hooks/useOktaSession';
+
 import './index.scss';
 
 type OktaSignInWidgetProps = {
   onSuccess: (auth: any) => any;
   onError: () => void;
   setError: (value: boolean) => void;
+  className?: string;
```
ClayBenson94, line 16: I can't find anywhere that this new prop is used -- can we nix it if it's not needed?

src/features/Landing/index.tsx
```diff
@@ -58,13 +68,27 @@ export const LandingHeader = () => {
...
 
         <span className="hide-print">
-          <UswdsReactLink
-            className="usa-button bg-mint-cool-vivid text-white width-auto"
-            variant="unstyled"
-            to="/signin"
-          >
-            {t('signIn')}
-          </UswdsReactLink>
+          {hasSession ? (
```
ClayBenson94, line 71: Similar to the code in the header -- I think a quick comment on the ternary for `hasSession` could be useful!

src/hooks/useOktaSession.ts
```diff
@@ -0,0 +1,23 @@
+import { useEffect, useState } from 'react';
+import { useOktaAuth } from '@okta/okta-react';
+
+const useOktaSession = () => {
```
ClayBenson94, line 4: Apologies for the comments about comments, but I think a top-level comment on custom hooks like this would be useful! Something like
```suggestion
// This is a custom hook that uses @okta/okta-react to asynchronously call to the Okta API to check to see if the user has a valid Okta session (which the user may have even when they don't have a valid application session)
const useOktaSession = () => {
```

ClayBenson94 (dismissed):

patrickseguraoddball:
@ClayBenson94 I'm sorry I didnt put this back in draft, it wasn't ready for review

garyjzhao (approved):
whoa this is cool! LGTM

PR #1379, opened 2024-09-24
[NOREF]cr tdl config follow up

StevenWadeOddball:
## Description

1. Configurable file path for CR and TDL
2. CR Summary is now Tagged Content ( it is not actually tagged, but we utilize this existing type for simplicity.) 

## How to test this change

1.  Query CR and TDLs
2. Ensure CR Summary is returned as a Tagged Content String

## PR Author Checklist

## PR Reviewer Guidelines

### DIFF
--- a/pkg/echimpcache/echimp_cache.go
+++ b/pkg/echimpcache/echimp_cache.go
@@ -13,9 +13,6 @@ import (
 	"github.com/cms-enterprise/mint-app/pkg/s3"
 )
 
-// TODO (echimp) make the file name configurable by envar file
-//TODO: consider making these constants configurable from environment variable
-
 // echimpCacheTimeHours is the length of time before a echimpCache needs to be refreshed
 const echimpCacheTimeHours = 3
 
--- a/pkg/echimpcache/echimp_cache.go
+++ b/pkg/echimpcache/echimp_cache.go
@@ -7,6 +7,7 @@ import (
 	"github.com/samber/lo"
 	"github.com/spf13/viper"
 
+	"github.com/cms-enterprise/mint-app/pkg/appconfig"
 	"github.com/cms-enterprise/mint-app/pkg/models"
 	"github.com/cms-enterprise/mint-app/pkg/parquet"
 	"github.com/cms-enterprise/mint-app/pkg/s3"
@@ -18,12 +19,6 @@ import (
 // echimpCacheTimeHours is the length of time before a echimpCache needs to be refreshed
 const echimpCacheTimeHours = 3
 
-// CRKey is the name of the file that is stored in the ECHIMP bucket representing all the echimp CR data
-const CRKey = "FFS_CR_DATA.parquet"
-
-// TDLKey is the name of the file that is stored in the ECHIMP bucket representing all the echimp TDL data
-const TDLKey = "TDL_DATA.parquet"
-
 var CRAndTDLCache *crAndTDLCache
 
 // GetECHIMPCrAndTDLCache returns a cached of data for CR and TDLs from an echimp s3 bucket.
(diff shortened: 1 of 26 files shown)


pkg/graph/resolvers/resolver_test_utilities.go
```diff
@@ -49,18 +52,17 @@ func GetDefaultTestConfigs() *TestConfigs {
...
+		Region:  viperConfig.GetString(appconfig.AWSRegion),
 		IsLocal: true,
 	}
 	//OS ENV won't get environment variables set by VSCODE for debugging
-	_ = os.Setenv(appconfig.LocalMinioAddressKey, config.GetString(appconfig.LocalMinioAddressKey))
-	_ = os.Setenv(appconfig.LocalMinioS3AccessKey, config.GetString(appconfig.LocalMinioS3AccessKey))
-	_ = os.Setenv(appconfig.LocalMinioS3SecretKey, config.GetString(appconfig.LocalMinioS3SecretKey))
+	_ = os.Setenv(appconfig.LocalMinioAddressKey, viperConfig.GetString(appconfig.LocalMinioAddressKey))
+	_ = os.Setenv(appconfig.LocalMinioS3AccessKey, viperConfig.GetString(appconfig.LocalMinioS3AccessKey))
+	_ = os.Setenv(appconfig.LocalMinioS3SecretKey, viperConfig.GetString(appconfig.LocalMinioS3SecretKey))
```
ClayBenson94, line 65: Can you clarify these lines? I know they're not entirely new in this PR, but I'm curious why they're actually needed. The only time we should be reading env vars is at the _very_ start of the program when we call `Viper.AutomaticEnv()` -- after that actual env vars shouldn't be needed (since everything's embedded in the viper config)

pkg/testconfig/s3testconfigs/s3_test_configs.go
```diff
@@ -4,40 +4,40 @@ package s3testconfigs
...
+		Region:  viperConfig.GetString(appconfig.AWSRegion),
 		IsLocal: true,
 	}
 	//OS ENV won't get environment variables set by VSCODE for debugging
-	_ = os.Setenv(appconfig.LocalMinioAddressKey, config.GetString(appconfig.LocalMinioAddressKey))
-	_ = os.Setenv(appconfig.LocalMinioS3AccessKey, config.GetString(appconfig.LocalMinioS3AccessKey))
-	_ = os.Setenv(appconfig.LocalMinioS3SecretKey, config.GetString(appconfig.LocalMinioS3SecretKey))
+	_ = os.Setenv(appconfig.LocalMinioAddressKey, viperConfig.GetString(appconfig.LocalMinioAddressKey))
+	_ = os.Setenv(appconfig.LocalMinioS3AccessKey, viperConfig.GetString(appconfig.LocalMinioS3AccessKey))
+	_ = os.Setenv(appconfig.LocalMinioS3SecretKey, viperConfig.GetString(appconfig.LocalMinioS3SecretKey))
```
ClayBenson94, line 24: Same question as before with regards to the `os.Setenv` lines -- can't quite make out why these are needed from reading through this

pkg/echimpcache/echimp_cache.go
ClayBenson94, line 1: You can probably nix the TODO comment on L16 (and maybe L17) now that we're passing the Viper config around!

ClayBenson94 (changes requested):

pkg/echimpcache/echimp_cache.go
StevenWadeOddball: Thanks for catching that!

pkg/graph/resolvers/resolver_test_utilities.go
StevenWadeOddball: Thanks Clay, it's been a while, but from what I'm remembering and the comment, this enabled debugging through the vscode testing tab. newS3 client calls os.Getenv instead of getting environment variables from viperConfig. This was a work around to get the variables so you can still debug and run tests with VSCode.

This is also part of the entrypoint for the test files (it's not called except for testing).

StevenWadeOddball: I think the other possibility is to rework NewS3Client to also take a viper.Viper param, and use that to get the envVar. I suspect it was originally done this way just for simplicity

ClayBenson94 (dismissed):
Looks good! Let's try out the new code without the `HEAD` call in `test` and see if it works without a permissions error!

ClayBenson94 (approved):

PR #1380, opened 2024-09-24
[MINT-2720] Data Exchange Approach

OddTomBrooks:
# MINT-0000

## How to test this change

1.  

## PR Author Checklist

## PR Reviewer Guidelines


pkg/models/plan_data_exchange_approach.go
```diff
@@ -0,0 +1,109 @@
+package models
+
+import (
+	"github.com/lib/pq"
+)
+
+// PlanDataExchangeApproach represents the data exchange approach of a model plan
+type PlanDataExchangeApproach struct {
+	baseTaskListSection
```
StevenWadeOddball, line 9: Unfortunately, this doesn't follow base task list section. I think you will have to manually implement some of these, eg embed some of fields and fucntionality in baseTaskListSection (
    it doesn't have a ready for review section.

    it doesn't have ready for clearance section (https://www.figma.com/design/qXhO12xSSkDF4LaToGKJoV/MINT-Blue-Sky-Ideas?node-id=1897-15624&node-type=section&t=LW5WWJE9wgY2GRBR-0)

    the status isn't a standard tasklist status (https://www.figma.com/design/qXhO12xSSkDF4LaToGKJoV/MINT-Blue-Sky-Ideas?node-id=1032-45951&node-type=section&t=LW5WWJE9wgY2GRBR-0)
)

pkg/graph/schema/types/plan_data_exchange_approach.graphql
```diff
@@ -0,0 +1,89 @@
...
+
+type PlanDataExchangeApproach {
+  id: UUID!
+  modelPlanID: UUID!
+  modelPlan: ModelPlan!
+
+  dataToCollectFromParticipants: [DataToCollectFromParticipants!]!
+  dataToCollectFromParticipantsReportsDetails: String
+  dataToCollectFromParticipantsOther: String
+  dataWillNotBeCollectedFromParticipants: Boolean!
```
StevenWadeOddball, line 55: we normally have these bools as nullable fields, (null meaning the question has not been answered)

pkg/graph/schema/types/plan_data_exchange_approach.graphql
```diff
@@ -0,0 +1,89 @@
...
+  dataToCollectFromParticipantsNote: String
+
+  dataToSendToParticipants: DataToSendToParticipants
```
StevenWadeOddball, line 58: this looks like a multi select in FIGMA, should probably be an array
https://www.figma.com/design/qXhO12xSSkDF4LaToGKJoV/MINT-Blue-Sky-Ideas?node-id=1897-15624&node-type=section&t=LW5WWJE9wgY2GRBR-0

pkg/graph/schema/types/plan_data_exchange_approach.graphql
```diff
@@ -0,0 +1,89 @@
...
+  dataToSendToParticipantsNote: String
+
+  doesNeedToMakeMultiPayerDataAvailable: YesNoType
```
StevenWadeOddball, line 61: what would you think about removing `does` from all of these? I think the name still makes sense without it, might be slightly better.

pkg/graph/schema/types/plan_data_exchange_approach.graphql
```diff
@@ -0,0 +1,89 @@
...
+  anticipatedMultiPayerDataAvailabilityUseCase: AnticipatedMultiPayerDataAvailabilityUseCase
+  doesNeedToMakeMultiPayerDataAvailableOther: String
```
StevenWadeOddball, line 63: I don't see other here in FIGMA

https://www.figma.com/design/qXhO12xSSkDF4LaToGKJoV/MINT-Blue-Sky-Ideas?node-id=2567-13949&node-type=frame&t=LW5WWJE9wgY2GRBR-0

pkg/graph/schema/types/plan_data_exchange_approach.graphql
```diff
@@ -0,0 +1,89 @@
...
+  doesNeedToMakeMultiPayerDataAvailableNote: String
+
+  doesNeedToCollectAndAggregateMultiSourceData: YesNoType
+  multiSourceDataToCollect: MultiSourceDataToCollect
```
StevenWadeOddball, line 67: the enums are correct, but this should be a multi-select
https://www.figma.com/design/qXhO12xSSkDF4LaToGKJoV/MINT-Blue-Sky-Ideas?node-id=2567-13949&node-type=frame&t=LW5WWJE9wgY2GRBR-0

pkg/graph/schema/types/plan_data_exchange_approach.graphql
```diff
@@ -0,0 +1,89 @@
...
+  multiSourceDataToCollectOther: String
+  doesNeedToCollectAndAggregateMultiSourceDataNote: String
+
+  willImplementNewDataExchangeMethods: YesNoType
+  newDataExchangeMethodsDescription: String
+  newDataExchangeMethodsNote: String
```
StevenWadeOddball, line 73: There is a field missing, there is a please describe field for willImplementNewDataExchangeMethods, and there is another for `Please describe any additional data exchange considerations.`

https://www.figma.com/design/qXhO12xSSkDF4LaToGKJoV/MINT-Blue-Sky-Ideas?node-id=2567-14638&node-type=frame&t=LW5WWJE9wgY2GRBR-0

pkg/graph/schema/types/plan_data_exchange_approach.graphql
```diff
@@ -0,0 +1,89 @@
...
+
+  isDataExchangeApproachComplete: Boolean!
```
StevenWadeOddball, line 75: I think this should probably get the same treatment that ready for review and ready for clearance had (eg markedCompleteBy and markedCompleteDts)

pkg/graph/schema/types/plan_data_exchange_approach.graphql
```diff
@@ -0,0 +1,89 @@
...
+  createdByUserAccount: UserAccount!
+  createdDts: Time!
+  modifiedBy: UUID
+  modifiedByUserAccount: UserAccount
+  modifiedDts: Time
+  readyForReviewBy: UUID
+  readyForReviewDts: Time
+  readyForClearanceBy: UUID
+  readyForClearanceDts: Time
+  status: TaskStatus!
```
StevenWadeOddball, line 88: This doesn't actually use the same task list status as the other task list sections unfortuantely. I think you will need to implement a new enum

StevenWadeOddball (changes requested):
Thanks Tom, this is just a high level review of the data type in graphql. There is some deviation from FIGMA that I wanted to call out now instead of waiting till the PR is promoted.

pkg/graph/schema/types/plan_data_exchange_approach.graphql
OddTomBrooks: I like this naming convention because it infers a discrete type, which follows general clean code principals. Typically this would be used for a boolean, but the YesNoType enum is essentially wrapping a boolean response.

pkg/models/plan_data_exchange_approach.go
OddTomBrooks: Please review this section again. I have set up a more discrete "Core" struct segment. It is currently intentionally redundant in interface definition to respect the Open-Close Principle. This allows for easier implementation of similar task list sections in the future. As this solution is more vetted, we can trim down the base task list section implementation and reuse more of the Core logic.

pkg/graph/schema/types/plan_data_exchange_approach.graphql
OddTomBrooks: The corresponding fields are captured in the sections highlighted in this image
image

ClayBenson94: image

Yeah -- looks like this should all be captured on this page as far as I can see @StevenWadeOddball -- is there something else from this page of the Figma that is missing, or do these fields cover it?

ClayBenson94: Forgot to mark `isDataExchangeApproachComplete` in there too for that final checkbox, but that's there!

OddTomBrooks: Resolving this as it appears to be a false positive, we can re-open the issue if we find that there is something missing

StevenWadeOddball: Thanks @OddTomBrooks and @ClayBenson94 , my apologies I missed the notification for this, but that Is my mistake. Thanks for clarifying!

pkg/graph/schema/types/plan_data_exchange_approach.graphql
```diff
@@ -0,0 +1,119 @@
...
+  MEDICAID_CLAIMS
+  MEDICARE_CLAIMS
+  PATIENT_REGISTRY
+  OTHER
+}
+
+type PlanDataExchangeApproach {
+  id: UUID!
+  modelPlanID: UUID!
+  modelPlan: ModelPlan!
```
ClayBenson94, line 50: Can probably delete this since we'll already have the `ModelPlan` if we're querying the `PlanDataExchangeApproach`

pkg/graph/resolvers/plandata_exchange_approach_helper.go
StevenWadeOddball, line 1: can you update the file name to have an `_`?

pkg/storage/plan_data_exchange_approachStore.go
```diff
@@ -0,0 +1,113 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return &approach, nil
+}
+
+// PlanDataExchangeApproachGetByModelPlanID returns the plan data exchange approach for a given model plan id
+func (s *Store) PlanDataExchangeApproachGetByModelPlanID(_ *zap.Logger, modelPlanID uuid.UUID) (*models.PlanDataExchangeApproach, error) {
+
+	println("PlanDataExchangeApproachGetByModelPlanID")
```
ClayBenson94, line 79: remove `println` or replace with `logger.Debug`

pkg/graph/resolvers/plan_data_exchange_approach.resolvers.go
```diff
@@ -0,0 +1,61 @@
...
+
+// !!! WARNING !!!
+// The code below was going to be deleted when updating resolvers. It has been copied here so you have
+// one last chance to move it out of harms way if you want. There are two reasons this happens:
+//   - When renaming or deleting a resolver the old code will be put in here. You can safely delete
+//     it when you're done.
+//   - You have helper methods in this file. Move them out to keep these resolver files clean.
+func (r *planDataExchangeApproachResolver) MarkedCompleteByUserAccount(ctx context.Context, obj *models.PlanDataExchangeApproach) (*authentication.UserAccount, error) {
+	return obj.MarkedCompleteByUserAccount(ctx)
+}
```
ClayBenson94, line 61: Can probs just nix this!

ClayBenson94 (dismissed):

pkg/graph/resolvers/plandata_exchange_approach_helper.go
OddTomBrooks: Oh! How embarrassing. Fixed. Nice catch.

pkg/graph/resolvers/plan_data_exchange_approach.go
```diff
@@ -62,7 +62,28 @@ func PlanDataExchangeApproachUpdate(
...
 	retDataExchangeApproach, err := store.PlanDataExchangeApproachUpdate(logger, existing)
 	return retDataExchangeApproach, err
 }
+
+// EvaluateStatus derives the status of the data exchange approach based on the current state of the model
+func EvaluateStatus(dea *models.PlanDataExchangeApproach) (models.DataExchangeApproachStatus, error) {
+
+	if dea.ModifiedDts == nil {
+		return models.DataExchangeApproachStatusNotStarted, nil
+	}
```
ClayBenson94, line 82: I don't think `ModifiedDts` is ever set as part of the Update function prior to this function being called.

I'd probably recommend that instead of calculating/evaluating the status based on the `models.PlanDataExchangeApproach`, we could probably just add a clause in the resolver itself that uses `isSettingToComplete` to determine which of the 2 statuses to set (`COMPLETE` if `isSettingToComplete`, otherwise `IN_PROGRESS`)

OddTomBrooks: 🤔 I see your point. Ok, 2 minutes.

patrickseguraoddball (changes requested):
I'm getting this error when creating a model plan and trying to query `status`

```
{
  "errors": [
    {
      "message": "the requested element is null which the schema does not allow",
      "path": [
        "modelPlan",
        "dataExchangeApproach",
        "status"
      ]
    }
  ],
  "data": null
}
```

ClayBenson94:
> I'm getting this error when creating a model plan and trying to query `status`
> 
> ```
> {
>   "errors": [
>     {
>       "message": "the requested element is null which the schema does not allow",
>       "path": [
>         "modelPlan",
>         "dataExchangeApproach",
>         "status"
>       ]
>     }
>   ],
>   "data": null
> }
> ```

@OddTomBrooks can you add a `NOT NULL` and `DEFAULT` value in the migration to avoid this?

pkg/models/plan_data_exchange_approach.go
```diff
@@ -0,0 +1,128 @@
...
+// PlanDataExchangeApproach represents the data exchange approach of a model plan
+type PlanDataExchangeApproach struct {
+	coreTaskListSection
+	markedCompleteByRelation
+
+	// Page 2
+	DataToCollectFromParticipants               pq.StringArray `json:"dataToCollectFromParticipants" db:"data_to_collect_from_participants" statusWeight:"1"`
+	DataToCollectFromParticipantsReportsDetails *string        `json:"dataToCollectFromParticipantsReportsDetails" db:"data_to_collect_from_participants_reports_details"`
+	DataToCollectFromParticipantsOther          *string        `json:"dataToCollectFromParticipantsOther" db:"data_to_collect_from_participants_other"`
+	DataWillNotBeCollectedFromParticipants      bool           `json:"dataWillNotBeCollectedFromParticipants" db:"data_will_not_be_collected_from_participants" statusWeight:"1"`
```
StevenWadeOddball, line 16: I think this was only updated on the front end , but this bool should also be a pointer ( and can then be removed from the instantiation)

pkg/sqlqueries/SQL/plan_data_exchange_approach/delete.sql
StevenWadeOddball, line 1: We generally don't delete form sections. Is there any reason to have this?

pkg/sqlqueries/plan_data_exchange_approach.go
```diff
@@ -0,0 +1,35 @@
...
+//go:embed SQL/plan_data_exchange_approach/get_by_id.sql
+var planDataExchangeApproachGetByIDSQL string
+
+//go:embed SQL/plan_data_exchange_approach/get_by_model_plan_id.sql
+var planDataExchangeApproachGetByModelPlanIDSQL string
+
+type planDataExchangeApproachScripts struct {
+	Create           string
+	Update           string
+	Delete           string
```
StevenWadeOddball, line 23: same as above, might be good to remove the delete method.

pkg/graph/resolvers/plan_data_exchange_approach.resolvers.go
```diff
@@ -0,0 +1,50 @@
...
+	return models.ConvertEnums[models.DataToSendToParticipants](obj.DataToSendToParticipants), nil
+}
+
+// MultiSourceDataToCollect is the resolver for the multiSourceDataToCollect field.
+func (r *planDataExchangeApproachResolver) MultiSourceDataToCollect(ctx context.Context, obj *models.PlanDataExchangeApproach) ([]models.MultiSourceDataToCollect, error) {
+	return models.ConvertEnums[models.MultiSourceDataToCollect](obj.MultiSourceDataToCollect), nil
+}
+
+// IsDataExchangeApproachComplete is the resolver for the isDataExchangeApproachComplete field.
+func (r *planDataExchangeApproachResolver) IsDataExchangeApproachComplete(ctx context.Context, obj *models.PlanDataExchangeApproach) (bool, error) {
```
StevenWadeOddball, line 41: this shouldn't hold up the PR, but this could be a method on the struct, then GQL auto resolves it and it would be testable

StevenWadeOddball (approved):
Thanks Tom, I think this is in a good state to merge to a feature branch. There are some thing that I think can be updated, but I think it would be better to merge this, and iterate on another PR.

I'd like to revisit the core struct before this targets main. I left a comment for a few other things I noticed, but as I don't think they will affect the front end, I think they can be addressed later.

PR #1381, opened 2024-09-25
[NOREF] Seed data for echimp

StevenWadeOddball:
## Description
1. Introduce seed data with existing echimp model data
2. Update Postman collection

## How to test this change

1.  Query  the seeded model plan, look at the related echimp CR or TDL
2. ( you can use the updated postman collection to do the query)

## PR Author Checklist

## PR Reviewer Guidelines

### DIFF
--- a/MINT.postman_collection.json
+++ b/MINT.postman_collection.json
@@ -2842,7 +2842,7 @@
 		},
 		{
 			"key": "modelPlanID",
-			"value": ""
+			"value": "082b49d3-b548-48cb-9119-77a281bc2a8c"
 		},
 		{
 			"key": "basicsID",
--- a/cmd/dbseed/main.go
+++ b/cmd/dbseed/main.go
@@ -163,7 +163,7 @@ func (s *Seeder) SeedData() {
 	now := time.Now()
 
 	// Seed an empty plan
-	emptyPlan := s.createModelPlan("Empty Plan", "MINT")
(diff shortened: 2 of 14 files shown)


patrickseguraoddball (changes requested):
Thanks Steven.  I'm still getting this error when trying to create a model plan manually.  I'm open to sync if you are

```
"failed to get S3 object: NoSuchKey: The specified key does not exist. status code: 404, request id: 17F8D2CA6B44E612, host id: d0af7ef13e9a79de08686e85c361d9af88c5ebdbbacf07a734ab2c11fd85e57f"
```

patrickseguraoddball:
I realize this may have been outside scope, but this was the original issue I brought to clay.  I'm unable to query the model plan when querying `echimpCRsAndTDLs`.  This is the error it throws.

StevenWadeOddball:
Thanks @patrickseguraoddball! As we discussed, the error occurs locally when data is not seeded.  Though not a result of this PR, we decided to implement a change to address this in this PR!

Now, if you can't get the cache, we should expect that nil is returned, and an error is logged in the backend. This will make sure we are alerted if this happens in a deployed environment, but allows flexibility when developing. You can confirm that the error shouldn't exist in the GQL response now, but you can look at the docker logs to see an error is logged if you query echimp crs or tdls without seeded data.

patrickseguraoddball (approved):
LG Thanks!

PR #1393, opened 2024-10-02
[Mint-3129]/Dataloaders new version refactor

StevenWadeOddball:
# MINT-3129

## Description

Since data loaders were first implemented in MINT, there have been some notable improvements in libraries. We initially looked to implement dataloadgen as a replacement to graph gophers, but instead decided to implement a more recent version of graph gophers which implements generics. The generic structure gives us some safety that was lacking, and also allows us to handle caching discretely (which is not allows in dataloadgen).

Also, this PR looks to abstract some of the repetitive patterns we have had to implement in the dataloaders, and instead utilize some generic structs to link methods together. 

Notable Changes
1.  Loaders grouped and exported by type
    a. See example for ModelPlan, Plan Basics, and Operational Solution
    b. Especially note how Operational Solution has two loader configs.
2. Note, loaders are no longer receiver methods, so they can be associated with a configuration
3. Loaders from context now returns a bool if the loader exists on the context or not
4. Dataloader new methods aren't being exported out of the package, the loader config from the model is now exported to support a semantic method of referencing a data loader,
    a. EX
       i. `loaders.OperationalSolutions.ByID.Load(ctx, id)`
       ii. `	return loaders.OperationalSolutions.AndPossibleByOperationalNeedID.Load(ctx,
		storage.SolutionAndPossibleKey{
			OperationalNeedID: operationalNeedID,
			IncludeNotNeeded:  includeNotNeeded,
		})`
5. Store methods for loaders are type safe thanks to the new generic types. If data is being serialized to JSON, that happens internally in the store method. See [OperationalSolutionExample](https://github.com/CMSgov/mint-app/blob/8f7a51547f1079f23d5278c42ba6efe323925ccf/pkg/storage/operational_solutionStore.go#L61)
6. Simple use cases can pass arrays as pq.Array. See [ModelPlan Example](https://github.com/CMSgov/mint-app/blob/5e55402c05cbb30ae2c6d8b487070319c1751fc7/pkg/storage/model_planStore.go#L27)
		

## How to test this change

1. Verify tests pass.
3. Hit dataloaders work through postman if you like
4. Mostly, read the code and make sure it makes sense to you

## PR Author Checklist

## PR Reviewer Guidelines


pkg/storage/loaders/loader_meta.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+// LoadFunc is a function called to load dataloader and return a result
+type LoadFunc[K comparable, V any] func(context.Context, K) (V, error)
+
+type LoaderConfig[K comparable, V any] struct {
+	batchFunction dataloader.BatchFunc[K, V]
+	// Load is an abstraction used to call a dataloader
+	Load LoadFunc[K, V]
+}
```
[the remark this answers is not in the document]
ClayBenson94: I think the most useful point to clarify is the _responsibility_ of the batch function. Put succinctly, I think the responsibility of the batch function is basically to:
1. Take a slice of input keys and use them to make a call to the database (typically a `package storage` call)
2. Assemble (often by ordering/mapping) a `[]*dataloader.Result` which the dataloader package uses to return the right data to the right spot in the GQL call

pkg/storage/loaders/data_loader_middleware.go
```diff
@@ -26,9 +26,15 @@ func NewDataLoaderMiddleware(loaders *DataLoaders) func(http.Handler) http.Handl
...
 
-// Loaders returns the dataLoaders for a given context
-func Loaders(ctx context.Context) *DataLoaders {
-	return ctx.Value(loadersKey).(*DataLoaders)
+// Loaders returns the dataLoaders for a given context, along with a bool to say if it returned. It will return true if the loader is present, and false if not
+func Loaders(ctx context.Context) (*DataLoaders, bool) {
+	loaders := ctx.Value(loadersKey)
+	if loaders == nil {
+		return nil, false
+	}
```
ClayBenson94, line 34: I think every call I found to this function ends up returning `ErrNoLoaderOnContext` -- maybe we should just return `(*DataLoaders, err)` from this function instead of a bool, and then we can just return that error in all those places instead of them each referencing `ErrNoLoaderOnContext`?

pkg/storage/SQL/operational_solution/and_possible_get_by_operational_need_id_LOADER.sql
```diff
@@ -13,9 +13,9 @@ possibleSolution AS (
     /*Get the mapped possible solutions for given operational need */
     SELECT
         OpSol.id,
-        qID.operational_need_id AS operational_need_id,
+        qID.operational_need_id,
         pOpSol.id AS solution_type,
-        OpSol.needed AS needed,
```
ClayBenson94, line 18: Were these just "useless" instances of name casting? (at least as far as `sqlx` is concerned?)

pkg/storage/SQL/operational_solution/and_possible_get_by_operational_need_id_LOADER.sql
```diff
@@ -32,10 +32,10 @@ possibleSolution AS (
...
         TRUE AS is_common_solution -- Linked Possible Solutions are always a common solution
     FROM QUERIED_IDS AS qID
-    INNER JOIN operational_need AS OpNd ON OpNd.id = qID.operational_need_id
-    INNER JOIN possible_need_solution_link AS PNSL ON PNSL.need_type = OpNd.need_type
-    INNER JOIN possible_operational_solution AS pOpSol ON pOpSol.id = PNSL.solution_type
-    LEFT JOIN operational_solution AS OpSol ON OpSol.solution_type = pOpSol.id AND OpSol.operational_need_id = OpNd.id
+    INNER JOIN operational_need AS OpNd ON qID.operational_need_id = OpNd.id
+    INNER JOIN possible_need_solution_link AS PNSL ON OpNd.need_type = PNSL.need_type
+    INNER JOIN possible_operational_solution AS pOpSol ON PNSL.solution_type = pOpSol.id
+    LEFT JOIN operational_solution AS OpSol ON pOpSol.id = OpSol.solution_type AND OpNd.id = OpSol.operational_need_id
```
ClayBenson94, line 38: Were these changes functional, or just re-ordering?

pkg/graph/resolvers/operational_solution.go
```diff
@@ -70,22 +70,11 @@ func OperationalSolutionUpdate(logger *zap.Logger, id uuid.UUID, changes map[str
...
-	if err != nil {
-		return nil, err
-	}
-
-	return result.([]*models.OperationalSolution), nil
+	return loaders.OperationalSolutions.AndPossibleByOperationalNeedID.Load(ctx,
+		storage.SolutionAndPossibleKey{
+			OperationalNeedID: operationalNeedID,
+			IncludeNotNeeded:  includeNotNeeded,
+		})
```
ClayBenson94, line 77: _Love_ this new structure of types for keys! 🔥

ClayBenson94 (changes requested):
Looks awesome, Steven -- love these changes!

I put in a few notes in my review:
- Some requests for additional documentation / comments
- Some _tiny_ refactors
- Some questions about some changes to make sure I understand their nature

pkg/storage/SQL/operational_solution/and_possible_get_by_operational_need_id_LOADER.sql
StevenWadeOddball: exactly! I have a sql fluff linter installed locally for VSCode. It suggested you shouldn't self alias, so I just implemented it.

pkg/storage/SQL/operational_solution/and_possible_get_by_operational_need_id_LOADER.sql
StevenWadeOddball: that's correct, this was the SQL fluff linter from VSCode, functionally they are equivalent

pkg/models/utilitites.go
```diff
@@ -142,7 +142,7 @@ func MapArrayToJSONArray(mapSlice []map[string]interface{}) (string, error) {
 }
 
 // StructArrayToJSONArray converts an array of structs to a JSON array
```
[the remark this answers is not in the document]
StevenWadeOddball: good call! I'll add it.

StevenWadeOddball:
@ClayBenson94, thanks for the great suggestions! It should be ready for re-review now.

ClayBenson94 (approved):
Synced with Steven and Sam on this -- things look good after some more final changes!

PR #1397, opened 2024-10-04
[MINT-3105] Front end work for FFS CRS and TDLs page

garyjzhao:
# MINT-0000

## How to test this change

1.  

## PR Author Checklist

## PR Reviewer Guidelines


src/components/TablePageSize/index.tsx
```diff
@@ -31,14 +33,14 @@ const TablePageSize = ({
...
         value={pageSize}
       >
-        <Option value={5} />
-        <Option value={10} />
-        <Option value={25} />
-        <Option value={50} />
-        <Option value={100} />
+        {valueArray.map(value => (
+          <Option value={value} />
+        ))}
```
patrickseguraoddball, line 43: This needs to have a key prop

src/components/EChimpCards/EChimpCard.tsx
```diff
@@ -0,0 +1,115 @@
...
+          )}
+        </div>
+      </CardBody>
+
+      <CardFooter>
+        <Button type="button" className="usa-button">
+          {crtdlsT('echimpCard.viewMore')}
+        </Button>
+        <ExternalLink
+          href="https://echimp.cmsnet/"
```
ClayBenson94, line 104: We're probably going to need to update this link at some point. Not sure if we want to do it in this PR or just a follow-up ticket

@garyjzhao @patrickseguraoddball any thoughts on this? We got some example URLs from the echimp team, but in order to make this work based on different environments, we'll probably need to add a new env var to the build pipeline, which maybe is worth it's own ticket?

src/components/EChimpCards/EChimpCardsTable.tsx
```diff
@@ -0,0 +1,209 @@
...
+    }
+    if (sort === 'title') {
+      // Handle undefined titles by putting them at the end
+      if (!a.title) return 1; // a has no title, so it should come after b
+      if (!b.title) return -1; // b has no title, so it should come after a
+      return a.title.localeCompare(b.title);
+    }
+    return 0;
+  });
+};
```
patrickseguraoddball, line 48: We might want to add a unit test for each of these functions/business logic

patrickseguraoddball (changes requested):
Nice work Gary. 

Do we have confirmation from Product/UX on what fields we should be searching for when using the global filter?  I'd imagine we also want to filter on status or date.  I didnt see anything specific in the AC. 

The padding between the heading and the model plan is too tight

Screenshot 2024-10-07 at 12 04 21 PM

When filtering by title, it works correct, then when adding a query, the sort doesn't revert back, it gets locked into the sort that was effected by the query.  You should probably run that title/id sort whenever the query changes as well

Screenshot 2024-10-07 at 12 09 11 PM

ClayBenson94 (changes requested):
I'm just going to request changes to make sure we also handle the new echimp URL added in 62d0fc9b

src/components/EChimpCards/EChimpCard.tsx
ClayBenson94: Resolving this in favor of my new comment!

patrickseguraoddball (changes requested):
None of the links from the env var to ECHIMP work or open up in a new tab.  Upon dev tool inspection, there is no embedded link/looks like the env vars are not getting baked in.  I added the var to my local env as well

There's padding here that's really large.  I actually think this may have been introduced as a part of the latest USWDS upgrade.  I would consider overwriding that updated style with the marge we used to have on `.usa-alert--info .usa-alert__body > *`

Screenshot 2024-10-08 at 1 11 37 PM

garyjzhao:
> There's padding here that's really large. I actually think this may have been introduced as a part of the latest USWDS upgrade. I would consider overwriding that updated style with the marge we used to have on `.usa-alert--info .usa-alert__body > *`
> 
> Screenshot 2024-10-08 at 1 11 37 PM

The padding is exactly what is in figma:

Figma:
image

Local code of the alert element (hard to screenshot and show the margins) (margin-bottom-6 is 48px):
image

patrickseguraoddball (approved):
LG

PR #1403, opened 2024-10-07
[MINT-2550] Data Exchange Approach Marked Complete Notification

OddTomBrooks:
# MINT-2550

## Description

- Added notification for when Data Exchange Approach is marked complete

## How to test this change

1. Open Postman
2. Create a model plan
3. Set the Post Man user account notification preference flags to include both EMAIL and IN_APP (I do this directly in the database)
4. Update the Data Exchange Approach section
5. Upon setting the Marked Complete bool to true, you should see an email in Mailcatcher and a should be able to see an in-app notification by getting the user's notifications in postman

## PR Author Checklist

## PR Reviewer Guidelines


ClayBenson94 (changes requested):
I think the A.C. for the user being able to select which models they get a notification for isn't quite accounted for from what I can see

image


This should probably just follow the same pattern that we did for the "Dates Changed" email ([migration here](https://github.com/CMS-Enterprise/mint-app/blob/922e75e6d55b0316caa2174cebe38ba2157f12bb/migrations/V160__Add_Dates_Changed_Notification.sql#L1-L16))

You could probably even just re-use/rename that type to something like `MODEL_NOTIFICATION_SELECTION_TYPE` or something like that since both use the same type of dropdown.

pkg/graph/resolvers/plan_data_exchange_approach.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	deaIsChangedMarkedCompleted := (existing.Status == nil || targetStatus != *existing.Status) && targetStatus == models.DataExchangeApproachStatusCompleted
+	existing.Status = &targetStatus
+
+	// Update the plan data exchange approach
+	retDataExchangeApproach, err := store.PlanDataExchangeApproachUpdate(logger, existing)
+
+	if deaIsChangedMarkedCompleted {
+		go func() {
```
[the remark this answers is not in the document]
StevenWadeOddball: Is `SendDataExchangeApproachMarkedCompleteNotification` suppoosed to be connected here?

cmd/dbseed/main.go
```diff
@@ -441,10 +441,48 @@ func (s *Seeder) SeedData() {
 			"cmmiGroups":      []string{"PATIENT_CARE_MODELS_GROUP", "SEAMLESS_CARE_MODELS_GROUP"},
 			"completeICIP":    "2020-05-13T20:47:50.12Z",
 			"phasedIn":        true,
-			"clearanceStarts": now.AddDate(0, 3, 0),
```
StevenWadeOddball, line 444: was this change needed?

StevenWadeOddball (changes requested):
Thanks Tom,  I tested your existing functionality and confirmed I received emails and notifications. I agree with Clay's note about needing to specify which models to receive notifications. 

Most of the feedback I've provided is in regards to implementing some standard paradigms for store methods, and a few about data instantiation. Please let me know if you have any questions.

pkg/storage/user_accountStore.go
```diff
@@ -184,3 +184,29 @@ func (s *Store) UserAccountsGetNotificationRecipientsForDatesChanged(
 
 	return recipients, nil
 }
+
+// UserAccountGetNotificationPreferencesForDataExchangeApproachMarkedComplete returns a collection of
+// user accounts that should be notified of when a data exchange approach is marked complete
+func (s *Store) UserAccountGetNotificationPreferencesForDataExchangeApproachMarkedComplete(
```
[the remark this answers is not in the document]
StevenWadeOddball: Thanks, note this still should not be a receiver method on store.

Also, the method should be taking the named preparer, not the store. (sorry I put that in wrong in the suggestion).
`	return sqlutils.SelectProcedure[models.UserAccountAndNotificationPreferences](np, sqlqueries.UserAccount.GetNotificationPreferencesDataExchangeApproachMarkedComplete, arg)`

cmd/dbseed/main.go
OddTomBrooks: This is part of an ugly merge. This is a good catch because it actually should be the `now` variable reused for assertions to avoid false negatives.

pkg/sqlqueries/SQL/user_account/get_notification_preferences_data_exchange_approach_marked_complete.sql
```diff
@@ -0,0 +1,54 @@
...
+    -- Combine collaborators and followers, removing duplicates
+    SELECT user_id FROM all_models
+    UNION
+    SELECT user_id FROM collaborators
+    UNION
+    SELECT user_id FROM followers
+)
+
+-- Join with user_account and user_notification_preferences to fetch user details and preferences
+SELECT DISTINCT
```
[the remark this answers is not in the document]
StevenWadeOddball: Note, this works fine as is though, you can change if you want or leave it.

pkg/graph/resolvers/plan_data_exchange_approach_helper.go
```diff
@@ -62,8 +61,7 @@ func SendDataExchangeApproachCompletedEmailNotification(
 	return nil
 }
 
-func SendDataExchangeApproachCompletedEmailNotifications(
-	ctx context.Context,
+func SendDataExchangeApproachMarkedCompleteEmailNotifications(
```
StevenWadeOddball, line 64: Question, is the email in any way modified when we call this function? It not, I think it might make sense to use the BCC function to send one email to multiple users instead of build and send the email multiple times.

pkg/graph/resolvers/plan_data_exchange_approach.go
```diff
@@ -27,53 +37,167 @@ func PlanDataExchangeApproachGetByModelPlanID(logger *zap.Logger, store *storage
...
-
-			existing.MarkedCompleteBy = &principal.Account().ID
-			existing.MarkedCompleteDts = models.TimePointer(time.Now().UTC())
-			existing.Status = models.DataExchangeApproachStatusCompleted
-		} else if !isSettingToComplete {
-			existing.MarkedCompleteBy = nil
-			existing.MarkedCompleteDts = nil
-			existing.Status = models.DataExchangeApproachStatusInProgress
-		}
+func TrySendDataExchangeApproachNotifications(
```
StevenWadeOddball, line 129: You have two really similar functions that do this, is there a reason they both exist? Looks like the other one is only called in dbseed `resolvers.SendDataExchangeApproachMarkedCompleteNotification`. Could / Should this be unified?

pkg/graph/resolvers/plan_data_exchange_approach_test.go
StevenWadeOddball, line 1: I don't see anything in this suite or the activity suite which validates who gets emails. Could you please add a test which verifies that users get a notification for the three preference types? (my models, following, etc)? (maybe just in app for simplicity)

migrations/V177__Add_Data_Exchange_Approach_Marked_Complete_Notification.sql
```diff
@@ -0,0 +1,15 @@
...
+  'MY_MODELS'
+  );
+
+
+ALTER TYPE ACTIVITY_TYPE
+ADD VALUE 'DATA_EXCHANGE_APPROACH_MARKED_COMPLETE' AFTER 'NEW_MODEL_PLAN';
+
+ALTER TABLE user_notification_preferences
+  ADD COLUMN data_exchange_approach_marked_complete USER_NOTIFICATION_PREFERENCE_FLAG[] DEFAULT '{}'::USER_NOTIFICATION_PREFERENCE_FLAG[],
+  ADD COLUMN data_exchange_approach_marked_complete_notification_type DATES_CHANGED_NOTIFICATION_TYPE DEFAULT NULL;
```
StevenWadeOddball, line 13: Maybe there should be a constraint on these columns? Something to ensure that if a user has non empty preferences for `data_exchange_approach_marked_complete` that `data_exchange_approach_marked_complete_notification_type` is not null? Otherwise, it probably needs to be handled in the app code.

This could be fine as is, I think effectively they just wont get any notifications, because the SQL will not return their preferencs if the type column is not filled out. It might be beneficial to enforce though so we don't have unexpected behavior.

pkg/graph/resolvers/plan_data_exchange_approach.go
```diff
@@ -24,56 +34,157 @@ func PlanDataExchangeApproachGetByModelPlanID(logger *zap.Logger, store *storage
...
+				go func() {
+					TrySendDataExchangeApproachNotifications(
+						ctx,
+						existing,
+						logger,
+						emailService,
+						emailTemplateService,
+						emailAddressBook,
+						principal,
+						store,
```
StevenWadeOddball: disregard, outdated

StevenWadeOddball (dismissed):
Thanks Tom, this looks good. There are just a couple things I noticed that I'd appreciate you taking a look at and responding to / or updating. Thanks!

pkg/graph/resolvers/plan_data_exchange_approach_helper.go
ClayBenson94: Yeah, if it's the same email being sent to multiple users I then using `BCC` would be preferred to making many calls to the SMTP server

pkg/graph/resolvers/plan_data_exchange_approach.go
OddTomBrooks: I went over it, specific functions are for SRP. These functions are verbose, following clean code principles. While they can be unified, I don't think this improves the code.

migrations/V177__Add_Data_Exchange_Approach_Marked_Complete_Notification.sql
OddTomBrooks: Discussed with @ClayBenson94, we are going to keep it homogenous with the dates change feature for now as a matter of consistency, then do a tech debt pass to update both solutions to default notification types to 'ALL_MODELS' and NOT NULL constraint.

PR #1407, opened 2024-10-08
[MINT-3157] Plan Data Exchange Data Loader

StevenWadeOddball:
# MINT-3157

## Description

This PR introduces a dataloader for the new type, plan data exchange. It uses the new version of dataloaders with generics.
--> Note, there is now generic sort and error functionality, but it still remains in [PR](https://github.com/CMS-Enterprise/mint-app/pull/1402). Once that is merged, this loader should also implement that sorting functionality, and error handling instead of the manual approach that happens in this PR.
--> Note, the failing front end tests are not caused by this PR, the FE development is still in progress.

--> This PR also introduces a generic function to validate the new loader structure. Effectively, a test suite just has to ensure that the data exists, and then specify a set of data to compare per key, and a function to specify how to compare it. This will remove a lot of the boiler plate code needed to test a new data loader.

## How to test this change

1.  Seed the data base.
2.  Query model plans and retrieve the data exchange type.
3.  Ensure that the returned data matches the expected model plan
4. Ensure all unit tests pass.

## PR Author Checklist

## PR Reviewer Guidelines


pkg/graph/resolvers/plan_data_exchange_approach.go
```diff
@@ -77,3 +79,8 @@ func PlanDataExchangeApproachUpdate(
 	retDataExchangeApproach, err := storage.PlanDataExchangeApproachUpdate(store, logger, existing)
 	return retDataExchangeApproach, err
 }
+
+// PlanDataExchangeApproachGetByModelPlanIDLoader calls a data loader to batch fetching a a plan data exchange object that corresponds to a model plan
+func PlanDataExchangeApproachGetByModelPlanIDLoader(ctx context.Context, modelPlanID uuid.UUID) (*models.PlanDataExchangeApproach, error) {
```
ClayBenson94, line 84: With the introduction of this function, do we still need `PlanDataExchangeApproachGetByID`?
- [resolvers.PlanDataExchangeApproachGetByID](https://github.com/CMS-Enterprise/mint-app/blob/dc19ab115c2d626a3cf73b9cacc06b8d123ca138/pkg/graph/resolvers/plan_data_exchange_approach.go#L19)
- We should keep the `storage` method by the same name, since it's used in `resolvers.PlanDataExchangeApproachUpdate`

ClayBenson94 (approved):

PR #1420, opened 2024-10-11
[NOREF] DataLoader One To Many Simplification (plan discussion loader refactor)

StevenWadeOddball:
## Description

1. This PR refactors plan discussion to use the new format dataloader
2. Using the simple case of plan discussion, we introduce a simplification wrapper to the one to many function
3. We assume that the result map is using the same key as the dataloader, and assume we can get the result simply by using the dataloaderkey. 
4. This allows us to retain flexibility if we need a more complicated get result method, or to keep it simple depending on needs

## How to test this change

1.  Run Unit tests. 
2. Get Plan Discussions on  model plan  to hit the new endpoint

## PR Author Checklist

## PR Reviewer Guidelines


pkg/helpers/mapping.go
```diff
@@ -108,3 +108,24 @@ func OneToManyFunc[K comparable, V any, mapKey comparable, Output any](keys []K,
...
+
+// getResultSimplified is a standard function to get the result of a one to many map assuming a simple map operation where the map key is the same as the dataloader key
+
+func getResultSimplified[K comparable, V any](key K, resultMap map[K][]V) ([]V, bool) {
+	res, ok := resultMap[key]
+	return res, ok
+}
+
+// OneToManyFuncSimplified uses the standard OneToManyFunc, but assumes that the mapKey is the same as the dataloaderkey K.
+func OneToManyFuncSimplified[K comparable, V any, Output any](keys []K, vals []V, getKey func(V) K, transformOutput func([]V, bool) Output) []Output {
```
ClayBenson94, line 127: Would it make sense to have this be the "default" version of the function, and instead clarify the "non-simple" version with extra behavior (so we don't have a bunch of calls to `xyzSimple()`, and only a few calls to `xyz()`?

StevenWadeOddball: That's a good call! I'll refactor it

StevenWadeOddball: Thanks @ClayBenson94 , I updated the names. I also removed `func` from the other functions in the spirit of simplifying the names of these functions.

PR #1422, opened 2024-10-14
[MINT-2720] Add basic data exchange seed data

ClayBenson94:
# MINT-2720

## Description

- Adds data exchange seed data
  - Currently doesn't add many variations of data (different statuses, different answers), but this is just aiming to have _some_ seed data in place.
- Updates postman collection for a method to fetch data exchange, rather than just update.

## How to test this change

1.  Run `scripts/dev up`, followed by `scripts/dev db:seed`
2. Open the UI and find the model with data exchange info, copying it's `id`
3. Run the new postman query to fetch data exchange with that ID, ensuring all data is correctly populated.

## PR Author Checklist

## PR Reviewer Guidelines


cmd/dbseed/resolver_wrappers.go
```diff
@@ -93,6 +93,37 @@ func (s *Seeder) updatePlanBasics(
...
 }
 
+// updateDataExchangeApproach is a wrapper for resolvers.UpdatePlanDataExchangeApproach
+// It will panic if an error occurs, rather than bubbling the error up
+// It will always update the Data Exchange object with the principal value of the Model Plan's "createdBy"
+func (s *Seeder) updatePlanDataExchangeApproach(
+	// ctx context.Context,
+	// emailService oddmail.EmailService, //TODO Add these once notification is set up
+	// emailTemplateService email.TemplateService,
+	// addressBook email.AddressBook,
```
OddTomBrooks, line 103: Might be worth waiting until notification is merged since the call spec is differentiated

OddTomBrooks (approved):
LGTM, I would suggest holding off for the notification merge since it changes the call spec

cmd/dbseed/resolver_wrappers.go
ClayBenson94: Yeah, I think I'll wait for the notification PR, then update it here rather than the other way around!

ClayBenson94: @OddTomBrooks 5e6e1959 just was pushed to resolve this. 

Currently set to hard-code email args to `nil` or `models.AddressBook{}` since we never really send emails with seed data anyways

PR #1431, opened 2024-10-17
[NOREF] Update TruncatedText component to take line clamp prop

garyjzhao:
# [NOREF]

## Description

The `TruncatedText` component currently only takes a character limit. But recently, there have been designs asking to truncate text to _x_
 amount of lines (line clamping) instead of character count. This PR updates the `TruncatedText` component to take in line clamp value

This PR proposes moving away from using `MentionTextArea` to parse and render the text with all of its text entity and uses [html-react-parser](https://github.com/remarkablemark/html-react-parser) to parse and render said text

## How to test this change

1. `yarn install` && restart servers
2.  Navigate to a CR and TDL side panel and see that CR Summary is line clamped to 5, instead of a character count.
3. Look elsewhere in the app to ensure any other instances of `TruncatedText` still renders correct

## PR Author Checklist

## PR Reviewer Guidelines


src/components/TruncatedText/index.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+      query: GetModelPlanBaseDocument,
+      variables: { id: 'f11eb129-2c80-4080-9440-439cbe1a286f' }
+    },
+    result: {
+      data: {
+        modelPlan: {}
+      }
+    }
+  }
+];
```
garyjzhao, line 20: Test throw error if Provider (and therefore mocks) were not included in test. Is there another way to solve this? Currently, this is just a random mock and not used in `TruncatedText` component

patrickseguraoddball: That's the incorrect query to mock.  TruncatedText utilized the `MentionTextArea`.  We might actually might want to not use this.  The `MetionTextArea` has a query embedded within it - `useSearchOktaUsersLazyQuery`.  I forgot why I decided to utilize `MentionTextArea`.  I don't think that was the right choice, considering how much overhead that component has built in.  We should probably try to build `TruncatedText` without it.  Sorry this was my fault

patrickseguraoddball (changes requested):
I dont think you addressed that unit test in your PR.  It's mocking the incorrect query.  Also we can replace MentionTextArea if you want, would be a good time.  This would make us not need to mock the query

patrickseguraoddball (approved):
LG, nice!

PR #1438, opened 2024-10-21
[MINT-3176] MTO initial schema

ClayBenson94:
# MINT-3176

## Description

- Adds initial MTO GraphQL schema to represent Milestones, Categories, Solutions, and Common Milestones/Solutions

## How to test this change

This doesn't introduce any functional changes -- only new schema / models. Please just review the structure of the schema, comments, and ask questions if anything doesn't seem clear or align with the Figmas!

## PR Author Checklist

## PR Reviewer Guidelines


pkg/graph/schema/types/model_to_operations_matrix.graphql
```diff
@@ -0,0 +1,95 @@
+# ModelsToOperationMatrix Notes:
+# - There's purposefully no `Milestones` query right now, since there's currently no need to query milestones oustide the scope of a Category
+type ModelsToOperationMatrix {
+  categories: [Category!]! # Categories is at the top level because you _can_ have empty categories
```
StevenWadeOddball, line 4: We need a way to retrieve un-categorized milestones as part of this. 

We may need to represent an empty category in the response, just so we can return milestones that have a null category.

pkg/graph/schema/types/model_to_operations_matrix.graphql
```diff
@@ -0,0 +1,95 @@
...
+  name: String!
+  description: String!
+  facilitatedBy: String # TODO: Make enum for `SolutionFacilitator` (TBD if that's the same as `MilestoneFacilitator`)
+  status: String! # TODO: Make enum for `SolutionStatus`
+  riskIndicator: String! # TODO: Make enum for `MTORiskIndicator`
+
+  # Custom Resolvers
+  relatedMilestones: [Milestone!]! # Can be empty, since you can have a solution with no milestones
+  addedFromMilestoneLibrary: Boolean! # TRUE if this is linked to a common milestone
+  solutions: [Solution!]!
```
StevenWadeOddball, line 42: Does a solution relate to other solutions?

pkg/graph/schema/types/model_to_operations_matrix.graphql
```diff
@@ -0,0 +1,95 @@
...
+  id: UUID!
+  name: String!
+  description: String!
+  facilitatedBy: String # TODO: Make enum for `SolutionFacilitator` (TBD if that's the same as `MilestoneFacilitator`)
+  status: String! # TODO: Make enum for `SolutionStatus`
+  riskIndicator: String! # TODO: Make enum for `MTORiskIndicator`
+
+  # Custom Resolvers
+  relatedMilestones: [Milestone!]! # Can be empty, since you can have a solution with no milestones
+  addedFromMilestoneLibrary: Boolean! # TRUE if this is linked to a common milestone
```
StevenWadeOddball, line 41: Is the bool above necessary? We probably would want to leave the link to the specific common solution, (though we could still implement this resolver if useful for the front end)

pkg/graph/schema/types/model_to_operations_matrix.graphql
```diff
@@ -0,0 +1,95 @@
...
+# It should not be used in contexts where a milestone has been added to a Model Plan (at that point, it'll be a "Milestone", not a "CommonMilestone")
+type CommonMilestone {
+  # DB Fields
+  id: UUID! # TODO Is this really needed? It might be kind of confusing to reference `id` like this
+  key: CommonMilestoneKey! # Similar to OperationalSolutionKey, enum of possible common milestones
+  name: String!
+  description: String!
+
+  # Custom Resolvers
+  isAdded: Boolean! # To represent if this Common Milestone is _already_ part of the Model Plan's MTO
```
StevenWadeOddball, line 55: I'm interested in this field.  At it's simplest, we just need to have knowledge of the model plan to know if they milestone was already added or not correct?

From lessons learned from operational solutions, I wonder if we should think about Unions to return common milestones?

 Would it be useful to have a resolver to get the milestone here by model_plan_id and common_milestone key?

StevenWadeOddball (commented):
Thanks Clay, great initial work on this. I think that this is looking good. I'd love to spend some time thinking about how we are planning on approaching custom milestones and solutions as well.

pkg/graph/schema/types/model_to_operations_matrix.graphql
ClayBenson94: Yeah -- currently this deesign (of just having a `categories` and no `milestones` property would demand that we have some sort of `"Uncategorized"` category returned)

If we don't go that route, we should proabably just add a `milestones` property here, which is also a pretty simple resolver!

pkg/graph/schema/types/model_to_operations_matrix.graphql
ClayBenson94: It doesn't -- I thought I actually removed this earlier, but looks like I must have slipped up and kept it around!

pkg/graph/schema/types/model_to_operations_matrix.graphql
ClayBenson94: Probably not - we will probably maintain that link as you said, but I don't think there's currently anything in the designs that would point towards needing to identify if a solution originated from the library or not. I'll nix it for now!

ClayBenson94: 56ab74e8

pkg/graph/schema/types/model_to_operations_matrix.graphql
ClayBenson94: 56ab74e8

patrickseguraoddball:
This looks good to me, thanks Clay!

pkg/graph/schema/types/model_to_operations_matrix.graphql
ClayBenson94: I think this is probably covered from our call, but for posterity:

Since we only query (currently) common milestones _in the context_ of a Model Plan, we can resolve this `isAdded` based on whether or not we see a Milestone with the same UUID/Fkey/whatever in that model plan's MTO!

ClayBenson94: But feel free to un-resolve if you've got any more questions/thoughts on this 😁

pkg/graph/schema/types/model_to_operations_matrix.graphql
```diff
@@ -0,0 +1,149 @@
...
+
+# TODO Update these with real values
+# Right now, this is just a placeholder type to represent the possible common solutions
+enum CommonSolutionKey {
+  SOLUTION_ONE
+  SOLUTION_TWO
+}
+
+# TODO: Add a way to determine order of Categories and subcategories
+type Category {
```
StevenWadeOddball, line 131: I just realized we may want to call this MTOCategory as the type (to avoid any name conflicts in the future. I can update when building out the category table though!

StevenWadeOddball (approved):
Thanks @ClayBenson94 ! I think this likes a great starting point. I think it's good to merge to feature at this point, and make changes as we discover a need when implementing

PR #1445, opened 2024-10-24
[MINT-3176] SQL for Shared MTO enums

ClayBenson94:
# MINT-3176

## Description

- Adds SQL migration (V177 at time of writing) for adding `MTO_FACILITATOR` and `MTO_RISK_INDICATOR` enums, based on this [GraphQL](https://github.com/CMS-Enterprise/mint-app/blob/3527a655e778898879a59da89cf105d67852b3cb/pkg/graph/schema/types/model_to_operations_matrix.graphql#L60-L83)

> [!NOTE]
> This PR doesn't (currently) introduce these enums into the `models` package (as they're generated in the `model` package). We can add them in this PR if preferred, or leave them as-is!

## How to test this change

1. Ensure migrations run as expected
2. Ensure SQL enums match GraphQLe

## PR Author Checklist

## PR Reviewer Guidelines


StevenWadeOddball:
Thanks Clay! I think we will want these in the models package instead of just generated, because they will be on the milestone and solution type when we make them.

ClayBenson94:
> Thanks Clay! I think we will want these in the models package instead of just generated, because they will be on the milestone and solution type when we make them.

That's actually what I'm about to push up 😁 as I wrote that note I realized I should just add them

StevenWadeOddball (approved):
Thanks @ClayBenson94 , this looks good comparing it to [FIGMA](https://www.figma.com/design/BcI7pNwfkmwaDD2EGGTsYC/IT-Lead?node-id=3836-13418&node-type=canvas&t=t3vIjAP86jISMfVv-0). Approving based on matching values and passing CI/CD

OddTomBrooks (approved):

PR #1446, opened 2024-10-24
[MINT-3182] Common milestones and common solutions

OddTomBrooks:
# MINT-3182

## How to test this change

1.  

## PR Author Checklist

## PR Reviewer Guidelines


migrations/V178__Add_MTO_Common_Milestone_And_Solution.sql
```diff
@@ -0,0 +1,18 @@
+CREATE TABLE mto_common_milestone (
+                            id UUID PRIMARY KEY,
+                            category_name ZERO_STRING NOT NULL,
```
ClayBenson94, line 3: I think this should probably just be
```suggestion
                            name ZERO_STRING NOT NULL,
```

migrations/V178__Add_MTO_Common_Milestone_And_Solution.sql
```diff
@@ -0,0 +1,18 @@
+CREATE TABLE mto_common_milestone (
```
ClayBenson94, line 1: I think both `mto_common_milestone` and `mto_common_solution` should probably have 2 new fields on them:
- `name` (to represent what the common milestone/solution is called)
- `key` (to represent a predictable enum/identifier for the Milestone/Solution -- I've got enums for them in GQL here! https://github.com/CMS-Enterprise/mint-app/blob/aa5fe3e9570117c3db1fb42f9768da2a481a811d/pkg/graph/schema/types/model_to_operations_matrix.graphql#L116-L128)

StevenWadeOddball: That makes sense to me! I'd be ok merging this to feature as is though, and implement the change in another PR for the sake of making other table relations. 

(I think just the migration will need to be re-numberd to 179 since category was merged.

ClayBenson94: I'm good with that, then! @OddTomBrooks can you just fix the failing lint on this and push that back up and we should be good!

ClayBenson94: I just pulled in the new branch -- you might need to pull it down and check migration #'s too!

StevenWadeOddball (approved):
Thanks @OddTomBrooks,

Per our slack conversations, I think we can go ahead and merge this to unblock other work, but come back to it to finish.

Some things to note, this doesn't match the graphql schema yet (like @ClayBenson94  mentioned, it should at least have a key and name, gql also shows description. Also, the type is created twice, once in models_top_operations_matrix, and again in the new graphql.

OddTomBrooks:
Merged per request @StevenWadeOddball

PR #1454, opened 2024-10-30
[MINT-3176] mto milestone resolvers

StevenWadeOddball:
# MINT-3176
This PR introduces basic functionality for the MTO milestone type. Additional work will continue in future PRs, but this allows us to get, create, and update a milestone.
## How to test this change

1.  Bring up the app
2. Seed Data
3. Use the postman collection to query the seeded data (there is a query just for that)
4. Create a model plan
5. Use the postman collection to create and update milestones.
6. Use the category resolver to create categories, and categorize your milestones as well

## PR Author Checklist

## PR Reviewer Guidelines


pkg/graph/schema/types/mto_milestone.graphql
```diff
@@ -12,17 +12,43 @@ type MTOMilestone {
   # DB Fields
   id: UUID!
   name: String! # If sourced from a Common Milestone, this will resolve from the common milestone table (you can't edit this for common milestones!)
+  mtoCommonMilestoneID: UUID
+  mtoCategoryID: UUID
   facilitatedBy: MTOFacilitator
   needBy: Time
   status: MTOMilestoneStatus!
   riskIndicator: MTORiskIndicator
```
ClayBenson94, line 20: This is marked as `NOT NULL` in the migration, but is nullable here -- which should it be?

pkg/models/mto_milestone.go
```diff
@@ -0,0 +1,44 @@
...
+	MTMInProgress MTOMilestoneStatus = "IN_PROGRESS"
+	MTMCompleted  MTOMilestoneStatus = "COMPLETED"
+)
+
+type MTOMilestone struct {
+	baseStruct
+	modelPlanRelation
+
+	Name                       *string            `json:"name" db:"name"`
+	mtoCommonMilestoneRelation                    //TODO (mto) maybe we shouldn't have an mto relation here? maybe just add the field?
```
ClayBenson94, line 22: I was going to provide similar feedback here (as is in the `TODO` comment)

I think just the ID here would be preferred since I don't think anything else in the MTO space will have a common milestone relation (unless common solutions will? but maybe just simpler to keep it just the ID here and extract it out later if we want to?)

ClayBenson94 (changes requested):
I know this is still a WIP, but just wanted to get some high-level thoughts out on this!

pkg/graph/schema/types/mto_milestone.graphql
StevenWadeOddball: You are right! I think this should be riskIndicator: MTORiskIndicator!

pkg/models/mto_milestone.go
StevenWadeOddball: yeah I think that makes sense, there isn't a ton of shared logic to it. I think the only other place like you said is common solution. It probably is clearer just adding the field itself.

StevenWadeOddball: removed in [36ca5a1](https://github.com/CMS-Enterprise/mint-app/pull/1454/commits/36ca5a1daa1d8d4c48cb4ea623534e2b33b1f373)

pkg/graph/resolvers/mto_category.resolvers.go
```diff
@@ -22,8 +20,17 @@ func (r *mTOCategoryResolver) SubCategories(ctx context.Context, obj *models.MTO
 }
 
 // Milestones is the resolver for the milestones field.
-func (r *mTOSubcategoryResolver) Milestones(ctx context.Context, obj *models.MTOSubcategory) ([]*model.MTOMilestone, error) {
-	panic(fmt.Errorf("not implemented: Milestones - milestones"))
+func (r *mTOSubcategoryResolver) Milestones(ctx context.Context, obj *models.MTOSubcategory) ([]*models.MTOMilestone, error) {
+	var categoryID uuid.UUID
+	if obj.IsUncategorized() {
+		// We show everything as a subcategory, if it is uncategorized, but has a parent id, it belongs to the parent cat
```
ClayBenson94, line 26: Took me a bit of noodling to figure out what exactly this code was doing -- took a stab at a new comment. Does this make sense?

```suggestion
	// All milestones (in the DB) have a mto_category_id, so we need to identify which to filter by in the query
	// We do this by passing the subcategory's parent ID (i.e. the Category ID) if the milestone doesn't actually have a subcategory
	// Otherwise, we can just filter by the actual subcategory ID
	var categoryID uuid.UUID
	if obj.IsUncategorized() {
```

StevenWadeOddball: Thanks Clay! That sounds good. This was part of the area I was scratching my head about yesterday, I think your verbiage is clearer.

cmd/dbseed/main.go
```diff
@@ -445,6 +445,8 @@ func (s *Seeder) SeedData() {
 			"highLevelNote":   "Some high level note",
 		},
 	)
+
+	s.seedModelPlanWithMTOData("Model Plan for MTO testing", "MINT", &mtoModelPlanUUID)
```
ClayBenson94, line 449: Appreciate the dev data 🙏

pkg/graph/schema/types/mto_milestone.graphql
```diff
@@ -12,17 +12,53 @@ type MTOMilestone {
   # DB Fields
   id: UUID!
   name: String! # If sourced from a Common Milestone, this will resolve from the common milestone table (you can't edit this for common milestones!)
+  key: CommonMilestoneKey
+  mtoCommonMilestoneID: UUID
+  mtoCategoryID: UUID
```
ClayBenson94, line 17: Will this technically never be null, or do we not return anything here in the case of `uncategorized`?

If it's the former, could you update the type to be
```suggestion
  mtoCategoryID: UUID!
```

If it's the latter, I think a comment here would be helpful (since that's not how it behaves on the actual Category/SubCategory types, which always have IDs)

migrations/V181__Add_MTO_Solution.sql
```diff
@@ -22,7 +22,7 @@ CREATE TABLE mto_solution(
     type MTO_SOLUTION_TYPE,
     facilitated_by MTO_FACILITATOR,
     status MTO_SOLUTION_STATUS NOT NULL,
-    risk_indicator MTO_RISK_INDICATOR,
+    risk_indicator MTO_RISK_INDICATOR NOT NULL DEFAULT 'ON_TRACK',
     is_draft BOOLEAN NOT NULL DEFAULT TRUE,  --TODO (mto) is this field needed?
```
ClayBenson94, line 26: Can probably just nix this field and the associated `TODO`

Looking through the Figma mockups, there's a pretty clear indicator that milestones can be marked as Draft ([src](https://www.figma.com/design/BcI7pNwfkmwaDD2EGGTsYC/IT-Lead?node-id=3619-31283&node-type=frame&t=3WFlUET2ohiwUTT8-0)), but there's not a field for editing a solution ([src](https://www.figma.com/design/BcI7pNwfkmwaDD2EGGTsYC/IT-Lead?node-id=3267-98504&node-type=instance&t=3WFlUET2ohiwUTT8-0))

pkg/graph/resolvers/mto_milestone.go
```diff
@@ -0,0 +1,70 @@
...
+	principalAccount := principal.Account()
+	if principalAccount == nil {
+		return nil, fmt.Errorf("principal doesn't have an account, username %s", principal.String())
+	}
+	existing, err := storage.MTOMilestoneGetByID(store, logger, id)
+	if err != nil {
+		return nil, fmt.Errorf("unable to update MTO Milestone. Err %w", err)
+	}
+
+	// Just check access, don't apply changes here
```
ClayBenson94, line 50: This comment doesn't seem accurate given that `applyChanges` and `checkAccess` are _both_ `true`

Should this just be
```suggestion
	// Check access and apply changes
```

ClayBenson94 (changes requested):
Just a few requested changes, but other than that it looks ready to merge!

pkg/graph/schema/types/mto_milestone.graphql
ClayBenson94: Realizing that this is resolved from the models field, so I presume this _can_ be `NULL` -- can you just add a comment clarifying that this means something along the lines of:
```md
`MTOMilestone.mtoCategoryID` does _not_ always equal `MTOMilestone.Category.ID`
```

migrations/V181__Add_MTO_Solution.sql
ClayBenson94: Or, honestly, feel free to leave it if you don't want to touch this a bunch here, since I know we opted to push `solution` work out and focus this PR on `milestones`!

I'll leave the call to you, but I think we can pretty safely/confidently remove this field

pkg/graph/schema/types/mto_milestone.graphql
StevenWadeOddball: Sure thing! Sounds good. But yeah you are correct. The Category will have UUID.Nil eg (`0000-0000-000-etc`), but the milestone itself can have a `nil` category id from the database. I'll add some clarifying comments! I think it probably is appropriate to leave this nullable as it says something about the underlying data type

migrations/V181__Add_MTO_Solution.sql
StevenWadeOddball: Thanks Clay! I'll just remove it here since we have all the context right now :D Thanks!

pkg/graph/resolvers/mto_milestone.go
StevenWadeOddball: Thanks for catching Clay! I updated it

migrations/V181__Add_MTO_Solution.sql
StevenWadeOddball: Resolved in [781550b](https://github.com/CMS-Enterprise/mint-app/pull/1454/commits/781550b7c44442a244b34acb43d26c3f9a6d31d8)

pkg/graph/schema/types/mto_milestone.graphql
StevenWadeOddball: Resolved in [781550b](https://github.com/CMS-Enterprise/mint-app/pull/1454/commits/781550b7c44442a244b34acb43d26c3f9a6d31d8)

ClayBenson94 (approved):

PR #1500, opened 2024-10-31
[MINT-3182] MTO Common Milestones and Common Solutions

OddTomBrooks:
# MINT-3182

## Description

Steven's notes
1. Created Common Milestone, Common Solution, and Common Milestone Solution Link table
2. Created a few test entries in the database
3. Data loaders for Common Solutions and Milestones
   a. Currently, it will provide contextual data about if a model is using a common milestone or common solution  ONLY at the top level.
   b. EX if you look for a solution associated with a common milestone, it will not provide any meta data. We can expand later if this is needed
 4. Common Solution says that the solution is added ONLY for the Model Plan. It doesn't return data about if it was added for the Common Milestone. 

## How to test this change

Steven's instructions
1.  Use the new postman queries
2. Query common Milestones
   a. At the Model To Operation Matrix top level
       i. Add the milestone and query again, note that the solution is now marked added
   b. At the top level query by key
       i. note is added is always false
3. Query Common Solutions
       a. At the Model To Operation Matrix top level
           i. Is added is based on if the Solution is added to a Model Plan)
       b. At the top level query by key
             i. note is added is always false
         c. At the Common Milestone Level
             i. note is added is always false (We could expand this if desired, but would need to interpret if we should check if it's added to a model plan, or that specifc need type.). This could create issues if the field means different things at different times.
 

## PR Author Checklist

## PR Reviewer Guidelines


StevenWadeOddball:
Thanks @OddTomBrooks , I think you need to pull in the feature, I made changes to the migration earlier this week. There is something weird with this git history here.

pkg/models/mto_common_milestone.go
```diff
@@ -0,0 +1,7 @@
+package models
+
+type ModelsToOperationCommonMilestone struct {
```
StevenWadeOddball, line 3:
```suggestion
type MTOCommonMilestone struct {

```

pkg/models/mto_common_solution.go
```diff
@@ -0,0 +1,5 @@
+package models
+
+type ModelsToOperationCommonSolution struct {
```
StevenWadeOddball, line 3:
```suggestion
type MTOCommonSolution struct {
```

pkg/graph/schema/types/model_to_operations_common_milestone.graphql
```diff
@@ -0,0 +1,16 @@
...
+
+  createdBy: UUID!
+  createdByUserAccount: UserAccount!
+  createdDts: Time!
+  modifiedBy: UUID
+  modifiedByUserAccount: UserAccount
+  modifiedDts: Time
+}
+
+extend type Query {
```
StevenWadeOddball, line 13: I don't think this needs to be exposed at the query level, I think we have it on the models to operation matrix, so we can get the model plan id from there

ClayBenson94: Correct -- we actually (currently) would want to avoid exposing this at anything other than at the level of a specific Model Plan, because properties like `isSuggested` / `isAdded` won't be resolvable

ClayBenson94 (changes requested):

StevenWadeOddball:
Thanks @ClayBenson94 , I implemented the changes. As we discussed, I'll proceed with merging to the feature branch now.

StevenWadeOddball (approved):

PR #1501, opened 2024-10-31
[NOREF] Only run 'Check Migrations' job when PR targets main

ClayBenson94:
# NOREF

## Description

- Updates `.github/workflows/check_migrations.yml` to only run when a PR targets main

This helps cut down on the number of failing jobs that we see when operating in feature branches (which often update/edit migrations _before_ they hit `main`)

> [!NOTE]
> Here's the [GitHub actions documentation](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#running-your-pull_request-workflow-based-on-the-head-or-base-branch-of-a-pull-request) for this modification!

## How to test this change

1.  Ensure that the modification matches the documentation -- otherwise, we just have to wait for this to hit `main` before we can really test it practically!

## PR Author Checklist

## PR Reviewer Guidelines

### DIFF
--- a/.github/workflows/check_migrations.yml
+++ b/.github/workflows/check_migrations.yml
@@ -10,6 +10,8 @@ name: Check Migrations
 
 on:
   pull_request:
+    branches:
+      - main # Only run this when targeting main, otherwise we should be free to edit / have bad migrations.
 
 jobs:
   check_migrations:


StevenWadeOddball (approved):
🚀

PR #1511, opened 2024-11-06
[MINT-3177]/category reordering be

StevenWadeOddball:
# MINT-3177

## Description

1. This enables ordering and reordering of a category in an MTO
2. There is a new column called position.
3. There is a trigger that runs on insert, update and delete. It will adjust related records. If a category is moved up, all categories between the positions get increased. If going the other way, the other positions are decreased

Note, the uncategorized categories currently only have a default position. If needed, an additional value can be populated for these.

Also, there is no DB constraint for a unique position, as this is handled by

## How to test this change

1.  Use the new reorder mutation to reorder categories
4. Reorder multiple categories and note the behavior when one is moved up and one is moved down

## PR Author Checklist

## PR Reviewer Guidelines


pkg/graph/resolvers/mto_category.go
```diff
@@ -56,6 +56,30 @@ func MTOCategoryUpdate(ctx context.Context, logger *zap.Logger, principal authen
...
+) (*models.MTOCategory, error) {
+	principalAccount := principal.Account()
+	if principalAccount == nil {
+		return nil, fmt.Errorf("principal doesn't have an account, username %s", principal.String())
+	}
+	existing, err := storage.MTOCategoryGetByID(store, logger, id)
+	if err != nil {
+		return nil, fmt.Errorf("unable to update MTO category. Err %w", err)
+	}
+	// update the name
```
ClayBenson94, line 72:
```suggestion
	// update the position to the new value
	// the re-ordering of other rows is handled in the trigger added in migrations/V188__Add_MTO_Category_Reorder_Trigger.sql
```

pkg/models/mto_category.go
```diff
@@ -29,6 +31,7 @@ func NewMTOCategory(createdBy uuid.UUID, name string, modelPlanID uuid.UUID, par
 
 // MTOUncategorized returns a placeholder category to hold all milestones that aren't categorized into a subcategory
 func MTOUncategorized(modelPlanID uuid.UUID, parentID *uuid.UUID) *MTOCategory {
+	//TODO (mto) should we add a position? Or ok leaving this as the base implementation?
```
ClayBenson94, line 34: What's the effect here? Is `order` returned as `0` for Uncategorized?

pkg/sqlqueries/SQL/mto/category/create.sql
```diff
@@ -3,20 +3,34 @@ INSERT INTO mto_category (
...
     created_by
 )
 VALUES (
     :id,
     :name,
     :parent_id,
     :model_plan_id,
+    COALESCE(
+        ( -- Place category at the bottom in order
+        -- TODO also join on model plan ID!!!! Otherwise the order is for everything
```
ClayBenson94, line 16: Guessing we can nix this second comment, as I see the `WHERE` statement captures the `model_plan_id` part now

.vscode/settings.json
```diff
@@ -29,7 +29,7 @@
     "AWS_S3_ECHIMP_BUCKET": "mint-app-echimp-uploads",
     "AWS_ECHIMP_CR_FILE_NAME": "FFS_CR_DATA.parquet",
     "AWS_ECHIMP_TDL_FILE_NAME": "TDL_DATA.parquet",
-    "AWS_ECHIMP_CACHE_TIME_MINS": 180,
+    "AWS_ECHIMP_CACHE_TIME_MINS": "180",
```
ClayBenson94, line 32: Not sure if this can just get reverted or not, but seems unrelated here!

ClayBenson94 (changes requested):
Just a few things noted when reviewing!

Overall, I think this is in a good spot to merge, but just a few comments/clarifications might be in place

.vscode/settings.json
StevenWadeOddball: Thanks Clay, this is needed. I just didn't catch it to previously. Otherwise, VsCode fails to start the debugger.  Surprisingly, it can't parse an int. I was debugging on this feature, so it needs to be here.

pkg/sqlqueries/SQL/mto/category/create.sql
StevenWadeOddball: Good call! Thanks Clay :D

pkg/models/mto_category.go
StevenWadeOddball: Yes, it's a 0. It's returned in the correct sequence, but the position doesn't match. Thinking about this more, I think we probably should handle this. I'll make a quick fix

StevenWadeOddball: Added this in [61f810b](https://github.com/CMS-Enterprise/mint-app/pull/1511/commits/61f810b7ab6f104543f3a514be31449bc052e36b)

ClayBenson94 (approved):

PR #1518, opened 2024-11-14
[MINT-3204] MTO Solution

OddTomBrooks:
# MINT-3204

## Description

1. Added MTO Solution GQL, SQL, Migration, Resolvers, and Store methods

## How to test this change

1.  Use Postman queries to evaluate creation of MTO Solutions, validate update behaviors, and validate resulting data

## PR Author Checklist

## PR Reviewer Guidelines


StevenWadeOddball (changes requested):
Thanks @OddTomBrooks, here is some high level feedback. You might already be tracking some of this, but just to have some early feedback!

pkg/graph/resolvers/mto_solution.go
```diff
@@ -0,0 +1,152 @@
...
+	principal authentication.Principal,
+	store *storage.Store,
+	id uuid.UUID,
+	changes map[string]interface{},
+) (*models.MTOSolution, error) {
+	principalAccount := principal.Account()
+	if principalAccount == nil {
+		return nil, fmt.Errorf("principal doesn't have an account, username %s", principal.String())
+	}
+	existing, err := storage.MTOSolutionGetByIDLoader(store, logger, []uuid.UUID{id})
```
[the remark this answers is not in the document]
OddTomBrooks: In order to keep things moving forward I'm going to open the PR while I figure out how to address this piece of feedback.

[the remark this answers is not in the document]
ClayBenson94: I'm alright to flag this with a `TODO` to revisit naming for now, but to just share my immediate thoughts:

This storage function isn't doing anything that's technically loader specific. It just says "Give me a list of IDs, and I'll return MTO Solutions for each of those IDs"

If that's used by a loader, that makes sense, but there's not any reason I'd say a function like this needs to be specifically designed or named around the loaders if it can be used in other contexts (like this one, technically!)

Let's sync on this sometime next week and see if we can approach naming / designing these in a way that feels more clear, but I'd rather just flag this with a comment for now and move past it for the sake of this PR 😁

pkg/graph/resolvers/mto_solution.resolvers.go
```diff
@@ -0,0 +1,72 @@
...
+		pocName,
+		pocEmail,
+	)
+}
+func (r *queryResolver) UpdateMTOSolution(ctx context.Context, id uuid.UUID, changes map[string]interface{}) (*models.MTOSolution, error) {
+	logger := appcontext.ZLogger(ctx)
+	principal := appcontext.Principal(ctx)
+
+	return MTOSolutionUpdate(logger, principal, r.store, id, changes)
+}
```
ClayBenson94, line 72: I think since these were moved from Queries to mutations you just need to copy this code back up into the newly generated resolvers above

pkg/graph/resolvers/mto_solution.resolvers.go
```diff
@@ -0,0 +1,72 @@
...
+// RelatedMilestones is the resolver for the relatedMilestones field.
+func (r *mTOSolutionResolver) RelatedMilestones(ctx context.Context, obj *models.MTOSolution) ([]*models.MTOMilestone, error) {
+	panic(fmt.Errorf("not implemented: RelatedMilestones - relatedMilestones"))
+}
+
+// CommonSolution is the resolver for the commonSolution field.
+func (r *mTOSolutionResolver) CommonSolution(ctx context.Context, obj *models.MTOSolution) (*models.MTOCommonSolution, error) {
+	logger := appcontext.ZLogger(ctx)
+
+	return MTOSolutionGetCommonSolutionByKeyLoader(r.store, logger, obj.Key)
```
ClayBenson94, line 26: This seems to be taking the wrong params -- I think we just need to pass `ctx` here too and it should be good!

```suggestion
	return MTOSolutionGetCommonSolutionByKeyLoader(ctx, r.store, logger, obj.Key)
```

pkg/graph/resolvers/mto_solution.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+func MTOSolutionGetCommonSolutionByKeyLoader(
+	ctx context.Context,
+	namedPreparer sqlutils.NamedPreparer,
+	logger *zap.Logger,
+	key *models.MTOCommonSolutionKey,
+) (*models.MTOCommonSolution, error) {
+	if key == nil {
+		return nil, fmt.Errorf("common solution key is nil")
```
ClayBenson94, line 35: We might need to look at this for both this code and for Milestones, but my guess is that we don't want to throw an error if we try and fetch a Common Solution for a given MTO Solution, but would rather just return `nil, nil`, since it's not really an error if you don't have a key, it just means you don't also have a common solution (which the return type is a pointer, so we can probably just modify this code as such)


```suggestion
		return nil, nil
```

ClayBenson94: https://github.com/CMS-Enterprise/mint-app/blob/c7dbfa452a1f97c6d9b127c28cfaccfb8f7868bf/pkg/graph/resolvers/mto_milestone.resolvers.go#L18-L24

Actually, there's this code on the Milestones that handles this -- might just be worth adding that to the resolver for this one (not in this file, but in `mto_solution.resolvers.go`, similar to how the change I linked is in `mto_milestone.resolvers.go`)

ClayBenson94 (changes requested):
Appreciate you addressing the changes @OddTomBrooks -- looks like there's a few things that might still need to be addressed outside the individual comments in my review:

- Looks like codegen & linting needs to be re-run, as there's a number of changes in other resolver functions that are unrelated to this change (that look like `gofmt`/`goimports` changes should catch)
- There's a merge conflict with the base branch, but it looks like it's also just in generated files, so you should be able to `git merge` the feature branch and just re-run codegen to address those

pkg/graph/resolvers/mto_solution.go
```diff
@@ -0,0 +1,138 @@
...
+func MTOSolutionGetByModelPlanIDLOADER(
+	ctx context.Context,
+	modelPlanID uuid.UUID,
+) ([]*models.MTOCommonSolution, error) {
+
+	// Translate a nil key to UUID nil, as we need a primitive type for translating results later
+	return loaders.MTOCommonSolution.ByModelPlanID.Load(ctx, modelPlanID)
+}
+
+func MTOSolutionGetCommonSolutionByKeyLoader(
```
ClayBenson94, line 27: Whatever method's calling this can probably just be replaced directly by a call to `MTOCommonSolutionGetByKeyLOADER`

pkg/graph/resolvers/mto.resolvers.go
```diff
@@ -34,7 +33,7 @@ func (r *modelsToOperationMatrixResolver) Milestones(ctx context.Context, obj *m
 }
 
 // Solutions is the resolver for the solutions field.
-func (r *modelsToOperationMatrixResolver) Solutions(ctx context.Context, obj *models.ModelsToOperationMatrix) ([]*model.MTOSolution, error) {
+func (r *modelsToOperationMatrixResolver) Solutions(ctx context.Context, obj *models.ModelsToOperationMatrix) ([]*models.MTOSolution, error) {
 	panic(fmt.Errorf("not implemented: Solutions - solutions"))
```
StevenWadeOddball, line 37: @OddTomBrooks , I think this is where the dataloader for get by model plan id would be called. I think the get by milestone id can / should hold off until we link solutions and milestones.

pkg/graph/resolvers/mto_milestone.resolvers.go
```diff
@@ -26,7 +25,7 @@ func (r *mTOMilestoneResolver) CommonMilestone(ctx context.Context, obj *models.
 }
 
 // Solutions is the resolver for the solutions field.
-func (r *mTOMilestoneResolver) Solutions(ctx context.Context, obj *models.MTOMilestone) ([]*model.MTOSolution, error) {
+func (r *mTOMilestoneResolver) Solutions(ctx context.Context, obj *models.MTOMilestone) ([]*models.MTOSolution, error) {
 	panic(fmt.Errorf("not implemented: Solutions - solutions"))
```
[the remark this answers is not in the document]
ClayBenson94: Apologies - for clarity, I think I meant to leave my comment -- Thanks @StevenWadeOddball !

https://github.com/CMS-Enterprise/mint-app/pull/1518#discussion_r1849103324

MINT.postman_collection.json
StevenWadeOddball, line 1: Please add the update mutation to the postman collection. It is needed in order to test all the functionality introduced in this PR

pkg/graph/schema/types/mto_solution.graphql
```diff
@@ -33,3 +35,23 @@ type MTOSolution {
   addedFromSolutionLibrary: Boolean!
   commonSolution: MTOCommonSolution
 }
+
+input MTOSolutionChanges @goModel(model: "map[string]interface{}") {
+  name: String # If sourced from a Common Solution, this will resolve from the common solution table (you can't edit this for common solution!)
+  facilitatedBy: MTOFacilitator
+  status: MTOSolutionStatus
+  riskIndicator: MTORiskIndicator
+  key: MTOCommonSolutionKey
```
StevenWadeOddball, line 44: Can you please take another look at what fields should be exposed in this Changes Struct? Notably, I think that `key` and `type` should not be included here ([figma](https://www.figma.com/design/BcI7pNwfkmwaDD2EGGTsYC/IT-Lead?node-id=4339-89395&node-type=frame&t=uqJnchVYFtXgZTG3-0) )

pkg/graph/resolvers/mto_solution.go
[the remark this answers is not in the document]
StevenWadeOddball: @OddTomBrooks, distinct from the naming convention of the store method, can you please update this to call a data loader method? 

We landed on using the data loader in the update resolver is acceptable (and preferred so the data is batched if called elsewhere).

pkg/storage/loaders/mto_solution.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"github.com/cms-enterprise/mint-app/pkg/appcontext"
+	"github.com/cms-enterprise/mint-app/pkg/models"
+	"github.com/cms-enterprise/mint-app/pkg/storage"
+
+	"github.com/graph-gophers/dataloader/v7"
+)
+
+// mtoSolutionLoaders is a struct that holds LoaderWrappers related to MTO Solutions
+type mtoSolutionLoaders struct {
```
StevenWadeOddball, line 16: As noted elsewhere, can you please implement the get by id  loader?

StevenWadeOddball (changes requested):
HI @OddTomBrooks , there is some functionality that is not working.

Can you please
1. Add the get by ID data loader (and utilize in the update statement)
2. Add all relevant mutations to the postman collection
3. Verify locally that the mutations and queries function as expected
4. Update the changes input type for the solution to only expose valid fields
5. Update the PR with testing instructions

pkg/graph/schema/types/mto_solution.graphql
OddTomBrooks: On reviewing this I noticed we were missing the needed_by field as well so I added it to the relevant areas.

MINT.postman_collection.json
```diff
@@ -3065,7 +3065,7 @@
 								"body": {
 									"mode": "graphql",
 									"graphql": {
-										"query": "mutation ReorderMTOCategory {\nreorderMTOCategory(id: \"{{mtoCategoryID}}\"\n,newOrder: 0\n# ,parentID: \"{{mtoCategoryID}}\"\n) {\n    position\n    id\n    name\n    isUncategorized\n    }\n}",
+										"query": "mutation ReorderMTOCategory {\nreorderMTOCategory(id: \"{{mtoCategoryID}}\",\nnewOrder: 0\n) {\n    position\n    id\n    name\n    isUncategorized\n    }\n}",
```
StevenWadeOddball, line 3068: @OddTomBrooks , it looks like you might have squashed some changes from the feature branch for postman. Can you please undo anything that overwrote existing queries from the feature? Sometimes I find it most helpful to just import the old one and drag the new queries into the collection and export it again. That can help resolve conflicts if they pop up

pkg/graph/schema/types/mto_solution.graphql
```diff
@@ -39,12 +39,9 @@ type MTOSolution {
 input MTOSolutionChanges @goModel(model: "map[string]interface{}") {
   name: String # If sourced from a Common Solution, this will resolve from the common solution table (you can't edit this for common solution!)
   facilitatedBy: MTOFacilitator
```
StevenWadeOddball, line 41: note that this is an array, but I think we discussed I will handle in my common solution update PR. If it blocks Gary before that PR is merged, I can work to get him unblocked.

pkg/graph/schema/types/mto_solution.graphql
```diff
@@ -39,12 +39,9 @@ type MTOSolution {
...
+  neededBy: Time
   status: MTOSolutionStatus
   riskIndicator: MTORiskIndicator
-  key: MTOCommonSolutionKey
-  type: MTOSolutionType # If sourced from a Common Solution, this will resolve from the common solution table (you can't edit this for common solution!)
-  pocName: String
```
StevenWadeOddball: good call for poc here. I'm a bit surprised they aren't editable, I have a follow up question about this in figma [here](https://www.figma.com/design/BcI7pNwfkmwaDD2EGGTsYC/IT-Lead?node-id=4339-101269&t=7VhEwAnqVOpk0jMH-4). We can reintroduce later if needed.

pkg/graph/resolvers/mto_solution.go
```diff
@@ -0,0 +1,123 @@
...
+	principal authentication.Principal,
+	store *storage.Store,
+	id uuid.UUID,
+	changes map[string]interface{},
+) (*models.MTOSolution, error) {
+	principalAccount := principal.Account()
+	if principalAccount == nil {
+		return nil, fmt.Errorf("principal doesn't have an account, username %s", principal.String())
+	}
+	existing, err := loaders.MTOSolution.ByID.Load(ctx, id)
```
StevenWadeOddball, line 31: We have been following a paradigm of loading methods living in their own function in resolvers unless there is a need to call it explicitly (eg fetching categories).
This  doesn't need to block the PR

StevenWadeOddball (changes requested):
Thanks @OddTomBrooks , I tested this and it works functionally. Only requesting changes because of overwriting changes from feature for postman. Could you please update?

pkg/graph/schema/types/mto_solution.graphql
StevenWadeOddball: @OddTomBrooks , I just checked with Zoë, poc and poc email should be update-able. Can you add that back into the changes object please?

StevenWadeOddball (approved):
Thanks @OddTomBrooks for correcting the postman collection.

PR #1519, opened 2024-11-14
[MINT-3212]/category reorder to parent

StevenWadeOddball:
# MINT-3212

## Description

This is a follow up to https://github.com/CMS-Enterprise/mint-app/pull/1511 after discovering a new requirement to be able to change a parent id of a child category.
A parent can't become a child.

This was accomplished by modifying the trigger to handle when a parent was changed to update the children of the old parent like a delete, and the parent of the new children like an insert. If an order is not specified for a new parent, the app code will provide the max of the current positions, effectively putting it at the end of the categories.

## How to test this change

NOTE!!! If you haven't reviewed and approved https://github.com/CMS-Enterprise/mint-app/pull/1511, please review that PR first. This builds on top of that PR, and will be simpler to review in that context.
1.  Run all unit tests
2.  Use Postment
3. Make categories
4. Change the parent and order
5. Validate the old parent children positions update as expected
6. Validate the new parent children positions update as expected
7. Validate that you can't make a parent category a child
8. Validate that you have to provide either an order or a parent, or you will get an error.

## PR Author Checklist

## PR Reviewer Guidelines

### DIFF
--- a/migrations/V188__Add_MTO_Category_Reorder_Trigger.sql
+++ b/migrations/V188__Add_MTO_Category_Reorder_Trigger.sql
@@ -29,17 +29,50 @@ BEGIN
     
     -- Handle Update: Reorder positions when a row's position is changed 
     ELSIF TG_OP = 'UPDATE' THEN
-        -- Determine if record moved up or down
-        IF NEW.position > OLD.position THEN
-            position_adjustment := -1;
-        ELSE
-            position_adjustment := 1;
-        END IF;
+    IF OLD.model_plan_id IS DISTINCT FROM NEW.model_plan_id THEN
+        RAISE EXCEPTION 'updating model_plan_id is not allowed. Caught in trigger function update_position_based_on_parent_and_model_plan';
+    END IF;
 
  /* Prevent updates to model_plan_id and parent_id as these are fixed for each category */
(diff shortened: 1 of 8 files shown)


migrations/V188__Add_MTO_Category_Reorder_Trigger.sql
```diff
@@ -29,17 +29,50 @@ BEGIN
...
-        IF NEW.position > OLD.position THEN
-            position_adjustment := -1;
-        ELSE
-            position_adjustment := 1;
-        END IF;
+    IF OLD.model_plan_id IS DISTINCT FROM NEW.model_plan_id THEN
+        RAISE EXCEPTION 'updating model_plan_id is not allowed. Caught in trigger function update_position_based_on_parent_and_model_plan';
+    END IF;
 
  /* Prevent updates to model_plan_id and parent_id as these are fixed for each category */
```
ClayBenson94, line 36: Should probably just move this comment up and remove the bit about parent_id
```suggestion
    /* Prevent updates to model_plan_id as categories cannot move from one model plan to another */
    IF OLD.model_plan_id IS DISTINCT FROM NEW.model_plan_id THEN
        RAISE EXCEPTION 'updating model_plan_id is not allowed. Caught in trigger function update_position_based_on_parent_and_model_plan';
    END IF;

```

PR #1520, opened 2024-11-14
[MINT-3201]/mto ready for review by

StevenWadeOddball:
# MINT-3201

## Description

1. Add ready for review columns to the model plan
2. Add mutation for toggling ready for review

Also updated the linking table to be correctly reference the user account table

## How to test this change

1.  Mark section ready for review by user
2. Ensure that it marked ready for review and has a time stamp
4. Change user (through postman variables)
5. Try to mark ready for review again
6. Ensure that it didn't change who marked ready for review or the time stamp
7. Set it as not ready for review
8. Ensure it no longer is ready for review
9. Mark ready for review again with the other user
10. Ensure the time and time stamp was marked

Verify that the unit tests work as expected.

## PR Author Checklist

## PR Reviewer Guidelines


ClayBenson94 (changes requested):
I chatted with @patrickseguraoddball about this today and we _might_ want to consider some other approaches for this that might be a bit more friendly with change history

Functionally, I think this all works just fine and the resolvers/mutations seem to be in the right spots, but it just might be worth considering if we want to break out MTO Data like this into another table that's _not_ on the model plan.

Let's sync up next week, but just leaving this comment so we can discuss further!

StevenWadeOddball:
Thanks @ClayBenson94, I've made this branch `MINT-3201/mto_ready_for_review_by_new_table`, and am currently testing functionality with a new table instead of using the model plan table.

PR #1531, opened 2024-11-20
[MINT-3218]/placeholder common milestones

StevenWadeOddball:
# MINT-3218

## Description

This is a first pass to make the common milestone library on the backend. They currently use operational need keys for simplicity, but we can update if it makes more sense based on name changed.

Note this is a work in progress, we are merging at this point to unblock front end work. There will be a follow up PR for solutions and solution milestone links

## How to test this change

1.  Look at the common milestones. Do they look as you expect
2. Can you hit the endpoints? (Postman)
3. Can you make a milestone from common milestone (Postman)?

## PR Author Checklist

## PR Reviewer Guidelines


patrickseguraoddball (approved):
Thanks!

PR #1532, opened 2024-11-21
SQLFluff upgrade to v3

ClayBenson94:
# NOREF

## Description

- Updates SQLFluff to the latest version
- Simplify config to specify [core rules](https://docs.sqlfluff.com/en/stable/reference/rules.html#core-rules) as our default, with some rules-specific configs on top of that
- Re-run SQLFluff with new version & config (which is why this touches so many files!)

## How to test this change

The main goal with testing this PR is to ensure that the linter runs consistently and behaves in a way that we find useful. For that reason, feel free to poke around with editing SQL files and re-running the linter as you feel necessary!

1.  Run `pre-commit run --all sqlfluff-fix` and ensure no files are touched and the command completes successfully (`echo $?` should return `0`!)
2. Edit a SQL File in a way that you think should break the linter config (mixing case of keywords, bad indentation, etc.)
3. Re run `pre-commit run --all sqlfluff-fix` and ensure the linter either fails or updates the file as expected (if an auto-fixable rule was broken)

## PR Author Checklist

## PR Reviewer Guidelines


.sqlfluff
```diff
@@ -3,21 +3,16 @@
...
 dialect = postgres
 ignore = parsing
+rules = core
 
 # Ignoring capitalisation.identifiers because it seems to want to uppercase column names if there are capitalized enums defined before the table (like in V11__Add_Plan_General_Characteristics)
 # Ignoring layout.long_lines because we break line-length requirements in a few places with comments
-# Ignoring aliasing.forbid so we can use table aliases
-# Ignoring structure.column_order becase we like to order SQL statements similarly across files
-exclude_rules = capitalisation.identifiers, layout.long_lines, aliasing.forbid, structure.column_order, layout.spacing
+exclude_rules = capitalisation.identifiers, layout.long_lines, layout.spacing, layout.indent
```
StevenWadeOddball, line 10: thanks Clay, do you want to re-enable the layout.indent check?

StevenWadeOddball (commented):
Thanks Clay, the checks for self aliasing look good

StevenWadeOddball (approved):
Thanks @ClayBenson94 ! This looks good. 

In my testing I found some unexpected behavior for sql fluff ( I don't think it was introduced by this change). If the sql is not proper sql, the pre-commit checks can still pass. 

I changed the keyword SELECT --> SELE and ran the commands, and the commands both passed 😢 . I think that might be a limitation in the linter. I think this is ok, because app code tests verify that the SQL works as expected. The linter worked with the well formed SQL as expected.

My indentation differences I was seeing from my vscode extension was just due to a version mismatch. I updated and now it indents the same

PR #1537, opened 2024-11-25
[NOREF]mto solution and milestone facilitated by multiselect

StevenWadeOddball:
# NOREF

## Description

Milestone and Solution facilitated by fields should be multi-select. This PR updates them to be.
A common Milestone will auto populate the facilitated by field when you create it.
ALSO, Solution poc fields should be nullable. This PR updates them.

## How to test this change

1.  Create a milestone from a common milestone
   a. make sure the result populates the facilitated by multi select
2. Make sure you can update the facilitated by field for a milestone
3. Make a solution
4. Make sure you can update the facilitated by field
5. Note when you make a solution from common solution, POC name and email are now both nullable.

## PR Author Checklist

## PR Reviewer Guidelines


pkg/graph/resolvers/mto_solution.go
```diff
@@ -81,27 +78,24 @@ func MTOSolutionCreateCustom(
...
-
 	// TODO: Populate data from mto common solution
 	mtoSolution := models.NewMTOSolution(
 		modelPlanID,
-		commonSolutionKey,
-		nil,
+		&commonSolutionKey,
 		nil,
-		models.MTOFacilitatorOther,
+		nil, // type is not stored for a solution from the common library
```
ClayBenson94, line 97: Similar comments on the other `nil` fields would be helpful, sometimes reading `nil, nil, nil` can be a bit confusing

pkg/graph/schema/types/mto_solution.graphql
```diff
@@ -13,14 +13,14 @@ enum MTOSolutionStatus {
...
+  facilitatedBy:  [MTOFacilitator!]
   neededBy: Time
   status: MTOSolutionStatus!
   riskIndicator: MTORiskIndicator
   key: MTOCommonSolutionKey
   type: MTOSolutionType # If sourced from a Common Solution, this will resolve from the common solution table (you can't edit this for common solution!)
-  pocName: String!
-  pocEmail: String!
+  pocName: String
+  pocEmail: String
```
ClayBenson94, line 23: Is it still safe to leave these as `String!`?

I can't remember if Common Solutions will come with POC info or not. If they do, then I think `String!` is still valid here, since creating a custom solution requires it, and so would a common one?

Not a big deal though, if we want it to match DB constraints that's fine, but might mean the Frontend has to build around a case that might not ever happen.

ClayBenson94 (changes requested):

migrations/V186__Add_MTO_Solution.sql
```diff
@@ -18,12 +18,14 @@ CREATE TABLE mto_solution(
...
+    facilitated_by MTO_FACILITATOR[],
     needed_by TIMESTAMP WITH TIME ZONE,
     status MTO_SOLUTION_STATUS NOT NULL,
     risk_indicator MTO_RISK_INDICATOR NOT NULL DEFAULT 'ON_TRACK',
-    poc_name ZERO_STRING NOT NULL,
-    poc_email EMAIL NOT NULL,
+    -- we allow null because this is is only required for custom solutions
+    poc_name ZERO_STRING,
+    -- we allow null because this is is only required for custom solutions
+    poc_email EMAIL,
```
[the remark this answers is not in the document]
StevenWadeOddball: You know, I'm not sure. I considered it, but also thought we might want to verify when Zoë returns if this is a field that users will be able to set or not. I left it for now since it isn't clear.

pkg/graph/schema/types/mto_solution.graphql
StevenWadeOddball: Thanks Clay! From my convos with Zoë last week, I don't believe common solutions have that field provided. We probably should discuss some more to be on the safe side.

I suspect that the common_solution_POC table will fill a similar role, but that is just speculation

StevenWadeOddball: This also goes along with your earlier comment if we should have a db constraint that disallows it for a common_solution 🤔 . We probably need to verify with Zoë

pkg/graph/resolvers/mto_solution.resolvers.go
```diff
@@ -36,9 +44,10 @@ func (r *mTOSolutionResolver) CommonSolution(ctx context.Context, obj *models.MT
 }
 
 // CreateMTOSolutionCustom is the resolver for the createMTOSolutionCustom field.
-func (r *mutationResolver) CreateMTOSolutionCustom(ctx context.Context, modelPlanID uuid.UUID, solutionType models.MTOSolutionType, facilitatedBy models.MTOFacilitator, name string, pocName string, pocEmail string) (*models.MTOSolution, error) {
+func (r *mutationResolver) CreateMTOSolutionCustom(ctx context.Context, modelPlanID uuid.UUID, solutionType models.MTOSolutionType, name string, pocName string, pocEmail string) (*models.MTOSolution, error) {
 	logger := appcontext.ZLogger(ctx)
 	principal := appcontext.Principal(ctx)
+	// models.EnumArray[models.MTOFacilitator]
```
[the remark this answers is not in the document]
StevenWadeOddball: Thanks! Good catch

pkg/graph/resolvers/mto_solution.go
StevenWadeOddball: That's a good call, thanks Clay. I added it. I removed the TODO as well, we don't actually need anything else from the common solution table at this point.

StevenWadeOddball:
Thanks @ClayBenson94 , I think the only questions left rely on knowing if you can ever set POC data for a solution from a common solution. We will probably need an answer from Zoë for that. Would you be good with proceeding with the assumptions currently in the PR and making changes if Zoë's answers differ?

migrations/V186__Add_MTO_Solution.sql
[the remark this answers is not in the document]
ClayBenson94: My guess is that anything that comes from the Common Milestone / Solution _probably_ isn't going to be editable, with the exception of Category?

I'm fine to leave this for now, though! Could you maybe flag a `TODO` to come back to this and add the constraint if we determine that to be the case?

ClayBenson94:
> Thanks @ClayBenson94 , I think the only questions left rely on knowing if you can ever set POC data for a solution from a common solution. We will probably need an answer from Zoë for that. Would you be good with proceeding with the assumptions currently in the PR and making changes if Zoë's answers differ?

Totally fine with that -- let's just leave some `TODO`s on those areas so we don't lose track in the Feature branch that we need to come back to that, and we can sync with her next week!

I'll continue my review ignoring those cases (I didn't get through all the files just yet!)

ClayBenson94 (dismissed):
Just finished reviewing all the files -- I think everything else is good to go!

(Just if you wouldn't mind adding those TODO comments where we were talking about the inherited fields from Common Solutions and if we should make a DB constraint!)

StevenWadeOddball:
> Just finished reviewing all the files -- I think everything else is good to go!
> 
> (Just if you wouldn't mind adding those TODO comments where we were talking about the inherited fields from Common Solutions and if we should make a DB constraint!)

Thanks @ClayBenson94! I added the todos

OddTomBrooks (approved):
Functionality works as expected

PR #1538, opened 2024-11-26
[NOREF] sqlfluff targets migrations after v180

StevenWadeOddball:
## Description

This PR introduces linting to the migration folder! IT uses regex to exclude any migration previous to migration 180.

## How to test this change

1.  Run `pre-commit run --all sqlfluff-fix`
2. Ensure that no migrations are changed (for example V80 has indentation that could change)
3. Add and stage a new migration with a number greater than 180, ensure that  `pre-commit run --all sqlfluff-fix` runs against it

Hint, you can look at the files it matches by scrolling up after running the command

## PR Author Checklist

## PR Reviewer Guidelines

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -5,7 +5,7 @@
 #########################
 
 # Helpers for .envrc
-APP_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+# APP_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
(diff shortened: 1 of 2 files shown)


StevenWadeOddball:
@ClayBenson94, do you think your [fix in EASI removing the tailscale references](https://github.com/CMS-Enterprise/easi-app/pull/2905) is also the fix for this lint failure?

OddTomBrooks (approved):
Tested and functions as expected

OddTomBrooks (approved):
Still works as expected

PR #1541, opened 2024-11-27
[MINT-3222] Milestone Solution Link

OddTomBrooks:
# MINT-3222

## Description

- Added support for automatic upsertion of solutions when creating a new common milestone
- Added support for milestone-solution links

## How to test this change

1. Validate `scripts/dev test:go` runs successfully
2. Through Postman create a new common solution including common solution keys and validate that milestone-solution links are created in the mto_mileston_solution_links table corresponding to your inputs

## PR Author Checklist

## PR Reviewer Guidelines


pkg/sqlqueries/SQL/mto/solution/get_by_model_plan_id_and_common_solution_key_LOADER.sql
```diff
@@ -0,0 +1,36 @@
...
+    mto_solution.modified_by,
+    mto_solution.modified_dts
+FROM mto_solution
+INNER JOIN QUERIED_IDS AS qIDs
+    ON
+        mto_solution.model_plan_id = qIDs.model_plan_id
+        AND (
+            mto_solution.mto_common_solution_key = qIDs.common_solution_key
+        -- TODO: Is this irrelevant for ENUM values?
+        -- OR (mto_solution.mto_common_solution_key IS NULL AND qIDs.mto_category_id = '00000000-0000-0000-0000-000000000000')
```
OddTomBrooks, line 34: @StevenWadeOddball This was sampled from a similar milestone query. I'm not sure what the thinking was on this segment. Refinements need to be made to this logic to handle cases where there is not a corresponding model plan id + common solution key. This is something you're specialized in so I'm curious about your thoughts here.

pkg/sqlqueries/SQL/mto/milestone/create_milestone_solution_link.sql
```diff
@@ -0,0 +1,32 @@
...
+)
+SELECT
+    qIDs.id,
+    qIDs.milestone_id,
+    qIDs.solution_id,
+    qIDs.created_by,
+    qIDs.created_dts,
+    qIDs.modified_by,
+    qIDs.modified_dts
+FROM QUERIED_IDS AS qIDs;
```
OddTomBrooks, line 32: @StevenWadeOddball This is a distinct approach in an attempt to batch creation using loaders. It may be over-engineering. I can rework this to a simple query if you think it's appropriate.

pkg/graph/resolvers/mto_milestone.go
```diff
@@ -78,7 +85,50 @@ func MTOMilestoneCreateCommon(ctx context.Context, logger *zap.Logger, principal
...
+		})
+
+		solutions, err := storage.MTOSolutionGetByModelPlanIDAndCommonSolutionKeyLoader(tx, logger, keys)
+
+		for i, solution := range solutions {
+			if solution == nil {
+				// TODO: Can this be moved upstream as part of the loader?
+				// TODO: Is there another more proper way to do this?
+				// TODO: Are non-existent solutions returned as nil? Can they be?
+				solution, err = MTOSolutionCreateCustom(
```
OddTomBrooks, line 110: @StevenWadeOddball This relates to another comment in the query. Thinking about the proper query-logic interface design here. I think it's about 90% there, but obviously some nuance remains.

pkg/sqlqueries/SQL/mto/solution/get_by_model_plan_id_and_common_solution_key_LOADER.sql
StevenWadeOddball: Thanks @OddTomBrooks ! I think this is on the right track good, I think you can remove the parenthesis here. 

That bit of code in the milestone resolver was because a milestone can have a NIL category, but NIL = NIL doesn't join correctly. I don't think we have this situation here, since you are just looking to join conditions where the solution is for a non nil value. 

So this
```sql
mto_solution.model_plan_id = qIDs.model_plan_id AND mto_solution.mto_common_solution_key = qIDs.common_solution_key
```
should return the correct response.

Does that make sense?

pkg/sqlqueries/SQL/mto/milestone/create_milestone_solution_link.sql
StevenWadeOddball: Thanks Tom! 

This is an Interesting idea, I'm not sure if there are any pitfalls with it but I can't think of it now! I think it would be interesting to proceed with it and see what happens! This wouldn't be a dataloader correct? Just a direct store call? (I think that's good, a dataloader might have other repercussions)

Were you thinking that you would do an ON CONFLICT / MERGE statement? If you did that, you wouldn't have to worry if the link already exists, you could just not error. 

That does bring up considerations that need to be handled around the generating the ids. You might consider getting the ID with gen_random_uuid or similar in the SQL instead of app code (since you wouldn't know if the link exists yet or not).

pkg/graph/resolvers/mto_milestone.go
StevenWadeOddball: Thanks Tom,

The other approach we briefly discussed was just trying to do an insert. So you could follow the insert common solution logic, but you just provide a list of common solution keys. If their is a record, you don't insert. If there isn't one, you insert.

Then you select and return all the solutions. With that approach, you don' have to iterate through every possible solution, or figure out if it exists or not, you just try to make all of them, and get a list of all of them. I think you can do the same thing with the links afterwards as well, just try to create them, and just ignore the insert attempt if the link already exists.  (another case for ON CONFLICT, or potentially a merge statement)

How do you think you want to proceed?

pkg/sqlqueries/SQL/mto/milestone/create_milestone_solution_link.sql
OddTomBrooks: I wound up simplifying this logic due to timeboxing constraints

pkg/graph/resolvers/mto_milestone.go
OddTomBrooks: Thanks for the input Steven, between your comment and a chat with Clay and after carefully weighing the options I agree that an upsert approach is better here and updated the logic

ClayBenson94 (changes requested):
Overall this looks good! I think the new `UPSERT`-style resolver reads cleanly!

There's a few pieces of code I think we can nix, but the main thing to look at is probably adding the JOIN to the create-solution (allowing conflicts) query

pkg/sqlqueries/SQL/mto/solution/create_allow_conflicts.sql
```diff
@@ -0,0 +1,66 @@
...
+    poc_email,
+    created_by
+FROM insert_attempt
+UNION ALL
+SELECT
+    id,
+    model_plan_id,
+    mto_common_solution_key,
+    name,
+    type,
```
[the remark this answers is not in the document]
OddTomBrooks: I'll try it and see

pkg/storage/loaders/mto_solution.go
```diff
@@ -18,6 +18,9 @@ type mtoSolutionLoaders struct {
 	ByID LoaderWrapper[uuid.UUID, *models.MTOSolution]
 	// ByModelPlanID Gets a list of mto Solution records associated with a model plan by the supplied model plan id.
 	ByModelPlanID LoaderWrapper[uuid.UUID, []*models.MTOSolution]
+
+	// ByModelPlanIDAndCommonKeys Gets a list of mto Solution records associated with a model plan by the supplied model plan id and common keys.
+	ByModelPlanIDAndCommonKeys LoaderWrapper[uuid.UUID, []*models.MTOSolution]
```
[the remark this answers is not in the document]
OddTomBrooks: Good catch. Yes this is ghost code.

pkg/storage/dataloader_keys.go
```diff
@@ -17,3 +21,10 @@ type MTOMilestoneByModelPlanAndCategoryKey struct {
 	// MTOCategoryID is the category the milestone is associated with. Note, this can be null
 	MTOCategoryID uuid.UUID `json:"mto_category_id"`
 }
+
+type MTOSolutionGetByModelPlanIDAndCommonSolutionKey struct {
+	// ModelPlanID is the model plan the solution is in reference to
+	ModelPlanID uuid.UUID `json:"model_plan_id"`
+	// CommonSolutionKey is the key of the common solution
+	models.MTOCommonSolutionKey `json:"common_solution_key"`
+}
```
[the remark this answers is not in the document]
OddTomBrooks: Good catch. Yes this is ghost code.

pkg/sqlqueries/mto_milestone.go
```diff
@@ -25,12 +31,18 @@ type mtoMilestoneScripts struct {
 	GetByModelPlanIDLoader string
 	// returns all Milestones by a model plan and category
 	GetByModelPlanIDAndCategoryIDLoader string
+	// creates a link between a milestone and a solution and returns the link ID
+	CreateMilestoneSolutionLink string
+	// returns all milestone-solution links for a given milestone by ID
+	GetMilestoneSolutionLinksByMilestoneID string
```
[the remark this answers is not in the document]
OddTomBrooks: I went back and forth on this. I'll add a TODO for now, it's a coinflip in my head and I think it depends more on how the code evolves.

pkg/sqlqueries/SQL/mto/milestone/create_milestone_solution_link.sql
```diff
@@ -0,0 +1,25 @@
+INSERT INTO mto_milestone_solution_link (
+    id,
+    milestone_id,
+    solution_id,
+    created_by,
+    created_dts,
+    modified_by,
+    modified_dts
```
StevenWadeOddball, line 8: I'd suggest only inserting created by here, and and allow the default value for created_dts, and don't set any modified by entries. This is consistent with other create statements. 

```suggestion
    created_by
```

pkg/storage/mto_milestone.go
```diff
@@ -88,3 +88,39 @@ func MTOMilestoneGetByID(np sqlutils.NamedPreparer, _ *zap.Logger, id uuid.UUID)
...
+// MTOMilestoneSolutionLinkGetByMilestoneID returns all MTO Milestone-Solution links for a given Milestone ID
+func MTOMilestoneSolutionLinkGetByMilestoneID(
+	np sqlutils.NamedPreparer,
+	_ *zap.Logger,
+	milestoneID uuid.UUID,
+) ([]*models.MTOMilestoneSolutionLink, error) {
+
+	arg := map[string]interface{}{"milestone_id": milestoneID}
+
+	returned, procErr := sqlutils.SelectProcedure[models.MTOMilestoneSolutionLink](np, sqlqueries.MTOMilestone.GetMilestoneSolutionLinksByMilestoneID, arg)
```
StevenWadeOddball, line 120: I'm not sure that you need to fetch links explicitly here, I think you can validate by fetching linked solutions instead. We don't really need to fetch links, just the solutions / milestones themselves.

pkg/storage/dataloader_keys.go
```diff
@@ -1,6 +1,8 @@
 package storage
 
-import "github.com/google/uuid"
+import (
+	"github.com/google/uuid"
```
StevenWadeOddball, line 4: Nitpick: you can probably revert this back to a single line import so it doesn't show a change here.

pkg/sqlqueries/SQL/mto/milestone/get_milestone_solution_links_by_milestone_id.sql
StevenWadeOddball, line 1: Related to earlier conversation, I think this might not be needed, instead we can just fetch the actual milestone / solution that is linked.

pkg/graph/resolvers/mto_milestone.go
```diff
@@ -78,13 +79,61 @@ func MTOMilestoneCreateCommon(ctx context.Context, logger *zap.Logger, principal
...
 		milestone.FacilitatedBy = &commonMilestone.FacilitatedByRole
 
 		if err := BaseStructPreCreate(logger, milestone, principal, store, true); err != nil {
 			return nil, err
 		}
-		return storage.MTOMilestoneCreate(tx, logger, milestone)
+		createdMilestone, err := storage.MTOMilestoneCreate(tx, logger, milestone)
+
+		var solutions []*models.MTOSolution
+		for _, commonSolutionKey := range commonSolutions {
```
StevenWadeOddball, line 97: I think these code blocks could be more efficient and batch the creation of solutions and links. I think you had an original approach that was looking pretty good for that. I think you can mostly do that in two SQL statements. 1. Create all solutions, 2. Create all links 

We have a similar approach for document solution links. https://github.com/CMS-Enterprise/mint-app/blob/9d729412c9295fd2a8b8fda01c03f9cf3033c961/pkg/storage/plan_document_solution_linksStore.go#L24


I don't think this has to happen now for the sake of unblocking front end work, but could you please add some todos to revisit this code block flow?

pkg/graph/resolvers/mto_milestone.go
```diff
@@ -78,13 +79,61 @@ func MTOMilestoneCreateCommon(ctx context.Context, logger *zap.Logger, principal
...
+		}
+
+		for _, solution := range solutions {
+			mtoMilestoneSolutionLink := models.NewMTOMilestoneSolutionLink(
+				principal.Account().ID,
+				createdMilestone.ID,
+				solution.ID,
+			)
+
+			_, milestoneSolutionLinkErr := storage.MTOMilestoneSolutionLinkCreate(
```
StevenWadeOddball, line 121: same as above, consider making one store method. A `TODO` is appropriate for this PR

pkg/graph/resolvers/mto_milestone_test.go
```diff
@@ -0,0 +1,43 @@
+package resolvers
+
+import (
+	"github.com/cms-enterprise/mint-app/pkg/models"
+	"github.com/cms-enterprise/mint-app/pkg/storage"
+)
+
+func (suite *ResolverSuite) TestCreateMilestoneSolutionLinks() {
```
StevenWadeOddball, line 8: This looks good @OddTomBrooks, Can you also create a test to confirm it works with no solution linked?

pkg/graph/resolvers/resolver_test.go
```diff
@@ -390,6 +390,24 @@ func (suite *ResolverSuite) createDefaultTestAnalyzedAudit(mp *models.ModelPlan,
 
 }
 
+func (suite *ResolverSuite) createMilestoneCommon(
```
StevenWadeOddball, line 393: @OddTomBrooks, this is fine here, wanted to see what you thought about moving to the mto_milestone test file instead though? I think it might feel a bit better organizationally, though we aren't doing this now.

pkg/storage/mto_solution.go
```diff
@@ -89,6 +89,28 @@ func MTOSolutionCreate(
 	return returned, nil
 }
 
+// MTOSolutionCreateAllowConflicts creates a new MTOSolution in the database, but
+// in the case of a conflict, instead just returns the conflicting row (and doesn't return an error)
+func MTOSolutionCreateAllowConflicts(
```
StevenWadeOddball, line 94: see other comment, I think we might want to consider bulk addition instead of one at a time. I'm good with leaving a todo

StevenWadeOddball (changes requested):
Thanks @OddTomBrooks , I think that this looks good overall! I left a few comments, mostly around todos for future enhancement and minor changes.

I think we can get this to be a bit more efficient, but I think a TODO could be appropriate for the sake of this PR.

pkg/storage/mto_milestone.go
OddTomBrooks: You're right. I thought about this quite a bit yesterday. In an ideal world our store would be mocked and we could do flow assertions, but the only way to validate the linking table is functioning as expected was to create a simple fetch. I feel like for unit testing / validation purposes having a fetch method is worthwhile.

pkg/graph/resolvers/resolver_test.go
OddTomBrooks: I think this is worth a higher level discussion on how we want to organize the logic in our test suite, let's have a chat after standup today

StevenWadeOddball (approved):
Thanks for making these changes @OddTomBrooks, it looks good. 

I think that git did something weird with some files (eg, it's showing you adding the new categories resolver in GQL). I don't think it matters functionally though

ClayBenson94 (approved):

OddTomBrooks:
> Thanks for making these changes @OddTomBrooks, it looks good.
> 
> I think that git did something weird with some files (eg, it's showing you adding the new categories resolver in GQL). I don't think it matters functionally though

This happened when I merged in from the base branch. I took the incoming logic, not sure why it would be an addition. Maybe the formatting is off or something.

PR #1542, opened 2024-11-27
[MINT-3229]milestone category resolvers and milestone by ID

StevenWadeOddball:
# MINT-3229

## Description

1. This implemented a resolver to return a new parent level MTOCategories type on an MTO milestone
    a. After syncing, we decided to rely on a parent level call and a dataloader to focus on simple maintainable code, instead of creating a more complicated dataloader that would return both records and deserialize.
2.  Syncing with the FE team, it was determined we should also be able to get a milestone by ID, so we don't have to call this category resolver when we are looking at all milestones for the model

## How to test this change

Use the postman queries to
1.  Create an MTO milestone
2. Get the MTO milestone by ID
3.  Verify the Categories information for the milestone.
    a. note a nil category will return uncategorized uncategorized

## PR Author Checklist

## PR Reviewer Guidelines

### DIFF
--- a/pkg/graph/resolvers/mto_category.go
+++ b/pkg/graph/resolvers/mto_category.go
@@ -151,8 +151,8 @@ func MTOCategoryGetByID(ctx context.Context, id uuid.UUID) (*models.MTOCategory,
 // We are not doing a larger SQL call that would return both objects, as that would result in less maintainable code
 func MTOCategoriesGetByID(ctx context.Context, id *uuid.UUID, modelPlanID uuid.UUID) (*models.MTOCategories, error) {
 	Categories := &models.MTOCategories{
-		Category:    *models.MTOUncategorized(modelPlanID, nil, 0),
-		SubCategory: *models.MTOUncategorizedSubcategory(modelPlanID, helpers.PointerTo(uuid.Nil), 0),
+		Category:    models.MTOUncategorized(modelPlanID, nil, 0),
+		SubCategory: models.MTOUncategorizedSubcategory(modelPlanID, helpers.PointerTo(uuid.Nil), 0),
 	}
 	if id == nil {
 		return Categories, nil
@@ -163,7 +163,7 @@ func MTOCategoriesGetByID(ctx context.Context, id *uuid.UUID, modelPlanID uuid.U
 	}
 	// Check if it is a parent category, if so early return without fetching parent
 	if immediateCategory.ParentID == nil {
-		Categories.Category = *immediateCategory
+		Categories.Category = immediateCategory
 		return Categories, nil
 	}
 
@@ -172,8 +172,8 @@ func MTOCategoriesGetByID(ctx context.Context, id *uuid.UUID, modelPlanID uuid.U
 	if err != nil {
 		return nil, err
 	}
-	Categories.SubCategory = *immediateCategory.ToSubcategory()
-	Categories.Category = *parentCategory
+	Categories.SubCategory = immediateCategory.ToSubcategory()
+	Categories.Category = parentCategory
 
 	return Categories, nil
 }
--- a/pkg/graph/resolvers/mto_category.go
+++ b/pkg/graph/resolvers/mto_category.go
@@ -8,6 +8,7 @@ import (
 	"go.uber.org/zap"
 
 	"github.com/cms-enterprise/mint-app/pkg/authentication"
+	"github.com/cms-enterprise/mint-app/pkg/helpers"
 	"github.com/cms-enterprise/mint-app/pkg/models"
 	"github.com/cms-enterprise/mint-app/pkg/storage"
 	"github.com/cms-enterprise/mint-app/pkg/storage/loaders"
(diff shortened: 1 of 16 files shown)


pkg/graph/resolvers/mto_category.go
```diff
@@ -138,3 +139,39 @@ func MTOSubcategoryGetByParentIDLoader(ctx context.Context, modelPlanID uuid.UUI
...
 	return append(dbSubcategories, models.MTOUncategorizedSubcategoryFromArray(modelPlanID, &parentID, dbSubcategories)), nil
 }
+
+func MTOCategoryGetByID(ctx context.Context, id uuid.UUID) (*models.MTOCategory, error) {
+	return loaders.MTOCategory.ByID.Load(ctx, id)
+}
+
+// MTOCategoriesGetByID returns the subcategory and parent category information given an mto category id
+// It will return uncategorized for nil values
+// We are not doing a larger SQL call that would return both objects, as that
```
ClayBenson94, line 149: Can you finish out this comment?

ClayBenson94: Might also be worth adding the overall "flow" of this function as a comment, i.e.

> This function first fetches the category by the provided ID. Then, if that category is a parent category (and we have nothing else to fetch), returns that category. However, if the supplied ID points to a subcategory (i.e. it has a parent ID), then it also fetches the parent category information so both category AND subcategory are 
 returned as part of this resolver.

pkg/graph/schema/types/mto_category.graphql
```diff
@@ -23,6 +22,15 @@ type MTOSubcategory {
   milestones: [MTOMilestone!]!
 }
 
+"""
+MTOCategories combines the concept of Category and Subcategory
+This allows it to be fetched mor conveniently in one resolver
```
ClayBenson94, line 27:
```suggestion
This allows it to be fetched more conveniently in one resolver
```

pkg/graph/schema/types/mto_milestone.graphql
```diff
@@ -36,16 +36,12 @@ type MTOMilestone {
...
-  #   subCategory: MTOSubcategory
-  # }
-  # TODO (mto) do we need to resolve category? Should we need to resolve it here? Also, 
-  # category: UNION MTOCategory | MTOSubcategory ?
-  category: MTOCategory!
-  subCategory: MTOSubcategory!
+  """
+  Category resolves relational category information.
+  if a milestone is uncategorized, this struct will still return
+  it will have a null value for both
```
ClayBenson94, line 42: Is this still accurate? My understanding of our code/approach is that it returns non-nullable types (i.e. the `Uncategorized` items) since the GQL is non-nullable for each of the fields in this type.

pkg/models/mto_category.go
```diff
@@ -22,6 +22,12 @@ type MTOCategory struct {
 	ParentID *uuid.UUID `json:"parent_id" db:"parent_id"`
 }
 
+type MTOCategories struct {
+	// TODO (mto) should these fields be pointers? Technically they should never be null
```
ClayBenson94, line 26: I could see an argument for both, but if we _do_ make them pointers, it _will_ return a GQL error if we try and return `nil` (since the Schema won't allow it).

That could be viewed as a pro (error if you try and do something you shouldn't) or a con (maybe it'd be better to just accidentally return a zero-valued category?)

Though, in just typing that out, maybe a pointer is better, that way if we _do_ somehow accidentally return a `nil` for one of these fields, we'll get a clear error rather than just returning a zero'd out `MTOCategory` or `MTOSubcategory`, which isn't really helpful.

pkg/models/mto_category.go
```diff
@@ -22,6 +22,12 @@ type MTOCategory struct {
 	ParentID *uuid.UUID `json:"parent_id" db:"parent_id"`
 }
```
ClayBenson94, line 24: Just a comment on this type might be helpful to explain what it's for, contextually!

```suggestion
// MTOCategories is a struct meant to represent the Category and Subcategory associated with a given MTO Milestone.
// This struct exists to allow us to resolve/fetch both Category and Subcategory at the same time and return them together in a single resolver.
```

pkg/storage/mto_milestone.go
```diff
@@ -12,6 +12,20 @@ import (
 	"github.com/cms-enterprise/mint-app/pkg/sqlutils"
 )
 
+// MTOMilestoneGetByIDsLoader returns all milestones by the provided ids
+func MTOMilestoneGetByIDsLoader(np sqlutils.NamedPreparer, _ *zap.Logger, ids []uuid.UUID) ([]*models.MTOMilestone, error) {
```
ClayBenson94, line 16: The other methods/loaders in this file reference the singular when referring to the keys, should we have this mimic that? (also lines up with Storage/SQL?)


```suggestion
func MTOMilestoneGetByIDLoader(np sqlutils.NamedPreparer, _ *zap.Logger, ids []uuid.UUID) ([]*models.MTOMilestone, error) {
```

pkg/storage/mto_category.go
```diff
@@ -12,6 +12,20 @@ import (
 	"github.com/cms-enterprise/mint-app/pkg/sqlutils"
 )
 
+// MTOCategoryGetByIDsLoader returns all categories by the provided
+func MTOCategoryGetByIDsLoader(np sqlutils.NamedPreparer, _ *zap.Logger, ids []uuid.UUID) ([]*models.MTOCategory, error) {
```
ClayBenson94, line 16: Same feedback here as in the milestone file -- should this be singular?

```suggestion
func MTOCategoryGetByIDLoader(np sqlutils.NamedPreparer, _ *zap.Logger, ids []uuid.UUID) ([]*models.MTOCategory, error) {
```

ClayBenson94 (changes requested):
I've left a few points of feedback and some suggestions -- mainly around `TODO` comments, other code comments, and function naming. Nothing too major, IMO!

pkg/graph/resolvers/mto_category.go
StevenWadeOddball: Thanks for catching this Clay!

pkg/graph/schema/types/mto_milestone.graphql
StevenWadeOddball: good call, you are right! I forgot to update this

pkg/models/mto_category.go
StevenWadeOddball: Good call, I'll switch to it. I also like that in favor of passing by reference instead of by value!

pkg/storage/mto_milestone.go
StevenWadeOddball: good call! Thanks

StevenWadeOddball:
Thanks for the feedback Clay! Changes should be implemented :D

ClayBenson94 (approved):
Changes look good to me! This should help unblock most of the FE for the edit milestone modal! 🚀

PR #1548, opened 2024-12-03
[NOREF] Initial MTO Milestone resolver tests

ClayBenson94:
# NOREF

## Description

- Adds in some tests for some existing logic in MTO Milestone resolvers and mutations

> [!IMPORTANT]
> This PR is purposefully _not_ exhaustive - I just wanted to get _some_ tests written for this resolver file so we had somewhere to write other tests, and to validate some of the more complex logic (like Common Milestones creating categories upon their instantiation)

## How to test this change

1.  Review the tests in `pkg/graph/resolvers/mto_milestone_test.go` and make sure they all make sense and are readable
2. Make sure CI/CD tests pass!

## PR Author Checklist

## PR Reviewer Guidelines

### DIFF
--- a/pkg/graph/resolvers/mto_milestone_test.go
+++ b/pkg/graph/resolvers/mto_milestone_test.go
@@ -89,6 +89,10 @@ func (suite *ResolverSuite) TestMTOMilestoneCreateCommonWithCategories() {
 	suite.Nil(milestone.ModifiedBy)
 	suite.Nil(milestone.ModifiedDts)
 
+	// We could potentially validate/check names of Categories on the Milestone object using MTOCategoriesGetByID,
+	// but we already want to fetch the entire list of categories later to make sure the right number show up,
+	// and we can just check for the right names to show up while we fetch them.
+
 	// Check and see if the categories came along as expected
 	// You can check which we expect to get created with migrations/V190__Add_Common_Milestone_Library.sql (or future similar migrations that edit the library)
 	// In this case MTOCommonMilestoneKeyManageCd should create 2 categories (parent, subcategory) of ('Operations','Participant and beneficiary tracking')
--- a/pkg/graph/resolvers/mto_milestone_test.go
+++ b/pkg/graph/resolvers/mto_milestone_test.go
@@ -157,3 +157,9 @@ func (suite *ResolverSuite) TestMTOMilestoneCreateCommonDuplicates() {
 	suite.Error(err)
 	suite.Nil(milestoneDupe)
 }
+
+// TODO (mto) Write tests for MTOMilestoneUpdate
+
+// TODO (mto) Maybe(?) write MTOMilestoneGetByModelPlanIDLOADER
+
+// TODO (mto) Maybe(?) write MTOMilestoneGetByModelPlanIDAndCategoryIDLOADER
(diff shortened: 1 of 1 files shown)


pkg/graph/resolvers/mto_milestone_test.go
```diff
@@ -0,0 +1,165 @@
...
+	suite.Nil(milestone.ModifiedDts)
+
+	// Check and see if the categories came along as expected
+	// You can check which we expect to get created with migrations/V190__Add_Common_Milestone_Library.sql (or future similar migrations that edit the library)
+	// In this case MTOCommonMilestoneKeyManageCd should create 2 categories (parent, subcategory) of ('Operations','Participant and beneficiary tracking')
+	categories, err = MTOCategoryGetByModelPlanIDLOADER(suite.testConfigs.Context, plan.ID)
+	suite.NoError(err)
+	suite.Len(categories, 2) // 1 new category (Operations) + Uncategorized
+
+	operationsCategory, found := lo.Find(categories, func(item *models.MTOCategory) bool {
```
StevenWadeOddball, line 95: Now that the resolver is merged in, I might suggest calling `MTOCategoriesGetByID` to verify the category and subcategory. I do like that this is validating the category creation though!

Also fine to keep this the way it is since you are validating all categories for the MTO!

pkg/graph/resolvers/mto_milestone_test.go
```diff
@@ -0,0 +1,165 @@
...
+	_, foundParticipant := lo.Find(operationsSubCategories, func(item *models.MTOSubcategory) bool {
+		return item != nil && item.Name == "Participant and beneficiary tracking"
+	})
+	suite.True(foundParticipant)
+
+	// Find 'Internal functions'
+	_, foundInternalFunctions := lo.Find(operationsSubCategories, func(item *models.MTOSubcategory) bool {
+		return item != nil && item.Name == "Internal functions"
+	})
+	suite.True(foundInternalFunctions)
```
StevenWadeOddball, line 141: Would it be useful to validate a case where the  subcategory already exists? Or that it will make a new category if you rename a category?

StevenWadeOddball (approved):
Thanks Clay! This looks great! I added some thoughts about other bits that could be tested, but they don't need to block this. Feel free to add or merge as is!

pkg/graph/resolvers/mto_milestone_test.go
ClayBenson94: I might add some `TODO` comments to consider those cases! I didn't do those right away since I wanted to just kickstart these tests with the happy paths, but I agree these would be good test cases!

pkg/graph/resolvers/mto_milestone_test.go
ClayBenson94: I think I'm just going to add a comment for a potential change here -- I had a bit of trouble implementing `MTOCategoriesGetByID ` because I still ended up wanting to fetch the total list of categories/subcategories to ensure the proper number of categories existed post-creation (which `MTOCategoriesGetByID` doesn't necessarily validate)

ClayBenson94: 077d6784d

pkg/graph/resolvers/mto_milestone_test.go
ClayBenson94: e5a276b9b

PR #1550, opened 2024-12-03
[NOREF]Constrain  mto solution poc fields

StevenWadeOddball:
# NOREF

## Description

1. POC email and POC name are now not editable if you make a custom solution
2. Milestone and solutions sourced from common options now update correctly, and don't attempt to set name and type

## How to test this change

1.  Create a solution from a common solution
2. attempt to edit the poc info.
3. Confirm the action fails.
4. Edit other information. Confirm it passes
5. Confirm you can update a milestone made from the common milestone library.

## PR Author Checklist

## PR Reviewer Guidelines

### DIFF
--- a/migrations/V186__Add_MTO_Solution.sql
+++ b/migrations/V186__Add_MTO_Solution.sql
@@ -53,10 +53,10 @@ COMMENT ON INDEX unique_name_per_model_plan_when_mto_common_solution_is_null IS
 
 
 ALTER TABLE mto_solution
-ADD CONSTRAINT check_name_type_and_common_solution CHECK (
-    (mto_common_solution_key IS NULL AND name IS NOT NULL AND type IS NOT NULL)
-    OR (mto_common_solution_key IS NOT NULL AND name IS NULL AND type IS NULL)
+ADD CONSTRAINT check_custom_solution_fields CHECK (
+    (mto_common_solution_key IS NULL AND ( name IS NOT NULL AND type IS NOT NULL AND poc_email IS NOT NULL AND poc_name IS NOT NULL))
+    OR (mto_common_solution_key IS NOT NULL AND ( name IS NULL AND type IS NULL AND poc_email IS NULL AND poc_name IS NULL))
 );
 
-COMMENT ON CONSTRAINT check_name_type_and_common_solution ON mto_solution IS
-'Ensures that if mto_common_solution_key is null, both name and type must be non-null; if mto_common_solution_key is provided, both name and type must be null.';
+COMMENT ON CONSTRAINT check_custom_solution_fields ON mto_solution IS
+'Ensures that if mto_common_solution_key is null (custom solution), name, type, poc_email, and poc_name must be non-null; if mto_common_solution_key is provided, these fields are required to be null (not entered).';
--- a/pkg/graph/resolvers/mto_milestone.go
+++ b/pkg/graph/resolvers/mto_milestone.go
@@ -101,6 +101,10 @@ func MTOMilestoneUpdate(ctx context.Context, logger *zap.Logger, principal authe
 	if err != nil {
 		return nil, fmt.Errorf("unable to update MTO Milestone. Err %w", err)
 	}
+	// Clear out the name returned from the database for convenience so we don't try to set the custom name field
+	if existing.AddedFromMilestoneLibrary() {
+		existing.Name = nil
+	}
 
 	// Check access and apply changes
 	err = BaseStructPreUpdate(logger, existing, changes, principal, store, true, true)
--- a/pkg/graph/resolvers/mto_milestone.go
+++ b/pkg/graph/resolvers/mto_milestone.go
@@ -101,7 +101,9 @@ func MTOMilestoneUpdate(ctx context.Context, logger *zap.Logger, principal authe
 	if err != nil {
 		return nil, fmt.Errorf("unable to update MTO Milestone. Err %w", err)
 	}
-	// Clear out the name returned from the database for convenience so we don't try to set the custom name field
+	// Since storage.MTOMilestoneGetByID will return a `Name` property when
+	// fetching milestones sourced from the common milestone library, we need to clear out that field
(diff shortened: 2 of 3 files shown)


migrations/V186__Add_MTO_Solution.sql
```diff
@@ -53,10 +53,10 @@ COMMENT ON INDEX unique_name_per_model_plan_when_mto_common_solution_is_null IS
 
 
 ALTER TABLE mto_solution
-ADD CONSTRAINT check_name_type_and_common_solution CHECK (
-    (mto_common_solution_key IS NULL AND name IS NOT NULL AND type IS NOT NULL)
-    OR (mto_common_solution_key IS NOT NULL AND name IS NULL AND type IS NULL)
+ADD CONSTRAINT check_custom_solution_fields CHECK (
+    (mto_common_solution_key IS NULL AND ( name IS NOT NULL AND type IS NOT NULL AND poc_email IS NOT NULL AND poc_name IS NOT NULL))
+    OR (mto_common_solution_key IS NOT NULL AND ( name IS NULL AND type IS NULL AND poc_email IS NULL AND poc_name IS NULL))
```
ClayBenson94, line 58: In doing some recent SQL work, I recently learned about the `IS DISTINCT FROM` or `IS NOT DISTINCT FROM` operator.

I wonder if this would be a different/cleaner way to write this?

```suggestion
ADD CONSTRAINT check_custom_solution_fields CHECK (
    (mto_common_solution_key IS DISTINCT FROM name)
    AND (mto_common_solution_key IS DISTINCT FROM poc_email)
    AND (mto_common_solution_key IS DISTINCT FROM poc_name)
```

By no means do we need to change it, but wanted to share my thoughts on this as I was re-reading this constraint!

pkg/graph/resolvers/mto_milestone.go
```diff
@@ -101,6 +101,10 @@ func MTOMilestoneUpdate(ctx context.Context, logger *zap.Logger, principal authe
 	if err != nil {
 		return nil, fmt.Errorf("unable to update MTO Milestone. Err %w", err)
 	}
+	// Clear out the name returned from the database for convenience so we don't try to set the custom name field
+	if existing.AddedFromMilestoneLibrary() {
+		existing.Name = nil
```
ClayBenson94, line 106: Maybe throw a `logger.Warn()` in here -- we shouldn't expect this to happen, but might be nice to find it if it does :)

pkg/graph/resolvers/mto_solution.go
```diff
@@ -32,6 +32,11 @@ func MTOSolutionUpdate(
 	if err != nil {
 		return nil, fmt.Errorf("unable to update MTO Solution. Err %w", err)
 	}
+	// Clear out the name and type returned from the database for convenience so we don't try to set the custom name  or type fields
+	if existing.AddedFromSolutionLibrary() {
```
ClayBenson94, line 36: Same here, if you wanna add the `logger.Warn()` from my suggestion above!

ClayBenson94 (approved):
Approving! Had some suggestions about potentially improved logging, and maybe a suggestion for the SQL constraint, but nothing that should stop this from merging as-is if you want!

migrations/V186__Add_MTO_Solution.sql
ClayBenson94: Basically a way for us to say "you can't have neither, you can't have both" with regards to these fields

ClayBenson94: Wait -- this might not actually work the way I thought, since if they're all set to values (but those values are different) then they're still distinct...

I'll keep this around for posterity, but I think we should just ignore this suggestion 😂

StevenWadeOddball: Sounds good! Yeah I think you are right that this wouldn't work as it currently stands because each area has separate values. It would be nice though!

pkg/graph/resolvers/mto_milestone.go
StevenWadeOddball: Thanks @ClayBenson94! We actually do expect this to happen every time there is a milestone sourced from a common milestone being updated, so I don't think we want to log a warning. (Because we return the common milestones name from SQL). We got around this previously because there were two distinct fields for operational needs.

pkg/graph/resolvers/mto_solution.go
StevenWadeOddball: Thanks @ClayBenson94, this is the same scenario, we expect this every time there is a solution sourced from a common solution.

pkg/graph/resolvers/mto_milestone.go
```diff
@@ -101,6 +101,10 @@ func MTOMilestoneUpdate(ctx context.Context, logger *zap.Logger, principal authe
 	if err != nil {
 		return nil, fmt.Errorf("unable to update MTO Milestone. Err %w", err)
 	}
+	// Clear out the name returned from the database for convenience so we don't try to set the custom name field
```
ClayBenson94, line 104:
```suggestion
	// Since storage.MTOMilestoneGetByID will return a `Name` property when fetching milestones sourced from the common milestone library, we need to clear out that field or else storage.MTOMilestoneUpdate will attempt to update the name (which won't be allowed, since this is a Milestone sourced from the common milestone library
```

migrations/V186__Add_MTO_Solution.sql
OddTomBrooks: We might be able to simplify this with XOR-like mutual exclusion. This is untested though. Non-blocker.

```suggestion
ADD CONSTRAINT check_custom_solution_fields CHECK (
    (mto_common_solution_key IS NULL) = (name IS NOT NULL AND type IS NOT NULL AND poc_email IS NOT NULL AND poc_name IS NOT NULL)
);
```

OddTomBrooks (approved):
I have no significant feedback. The changes are succinct and self-evident.

PR #1553, opened 2024-12-05
[MINT-3238] Standard Categories

ClayBenson94:
# MINT-3238

## Description

- Adds a new mutation that creates a standard set of categories and subcategories when called.
- Adds `Resolver` suite tests to cover new behavior
- Adds new Postman query

> [!NOTE]
> This PR adds a `sqlutils.WithTransactionNoReturn` method, which is equivalent to `sqlutils.WithTransaction`, except that it doesn't return any values -- useful for if you need to call a bunch of storage methods as part of a transaction but don't care about a specific return value (just if an error occurred)

> [!CAUTION]
> I took the approach of sequentially calling a storage method as part of a transaction rather than trying to create a complex SQL/Storage method that would bulk insert these (since allowing conflicts and ordering seemed like a difficult solve)
> There's currently a `TODO` in the codebase here to consider refactoring this, but it seems to work well enough as-is

## How to test this change

1.  Create a new model plan (or use an existing one from `scripts/dev db:seed`)
2. In postman, call the new `createStandardCategories` mutation that was added with the appropriate model plan ID
3. In postman, call `GetMTOCategories` to fetch the list of categories and subcategories to make sure everything works as expected!

> [!TIP]
> You can also try renaming categories and ensuring the `createStandardCategories` mutation re-creates them! (They will appear at the end of the categories list when re-created)

> [!NOTE]
> At time of writing, the UI won't properly handle rendering empty categories in the MTO Matrix page, so please test this feature using Postman!
> This is addressed in https://github.com/CMS-Enterprise/mint-app/pull/1556

## PR Author Checklist

## PR Reviewer Guidelines

### DIFF
--- a/MINT.postman_collection.json
+++ b/MINT.postman_collection.json
@@ -2818,6 +2818,39 @@
 					"name": "Categories",
 					"item": [
 						{
+							"name": "Create Standard Categories",
+							"event": [
+								{
+									"listen": "test",
+									"script": {
+										"exec": [
+											""
+										],
+										"type": "text/javascript",
+										"packages": {}
+									}
+								}
+							],
+							"request": {
+								"method": "POST",
+								"header": [],
+								"body": {
+									"mode": "graphql",
+									"graphql": {
+										"query": "mutation CreateStandardCategories {\n    createStandardCategories(modelPlanID: \"{{modelPlanID}}\")\n}",
+										"variables": ""
+									}
+								},
+								"url": {
+									"raw": "{{url}}",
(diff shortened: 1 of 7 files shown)


StevenWadeOddball (approved):
Thanks Clay, this looks good. It's worth calling out that this uses relative order instead of an explicit order. EG, if a category already exists, these will be appended to the end of the list in order. I think that is fine, but I want to call it out in case it is an issue. We can revisit later if so!

PR #1554, opened 2024-12-05
[MINT-3227] API Support for deleting an MTO Milestone

ClayBenson94:
# MINT-3227

## Description

- Adds API Support (GraphQL mutation) for deleting an MTO milestone
- Adds Postman Query
- Adds resolver test (honestly pretty minimal at this point, maybe could bolster this up)

> [!NOTE]
> This PR also introduces a `sqlutils.ExecProcedure` method, useful for calling delete statements which do not need to return any values

## How to test this change

1. Use `scripts/dev db:seed` to populate an MTO with some milestones
2. Optionally create some milestones from the common milestone library and link some solutions
3. Use the postman query to delete those milestones
  a. Ensure that the milestones are deleted properly
  b. Ensure that any milestone solution links are also deleted (but _not_ the solutions themselves

## PR Author Checklist

## PR Reviewer Guidelines

### DIFF
--- a/pkg/graph/schema/types/mto_milestone.graphql
+++ b/pkg/graph/schema/types/mto_milestone.graphql
@@ -65,5 +65,6 @@ extend type Mutation {
   @hasRole(role: MINT_USER)
   updateMTOMilestone(id: UUID!, changes: MTOMilestoneChanges!): MTOMilestone!
   @hasRole(role: MINT_USER)
-  #TODO (mto) will we need any other mutations here? Can we ever delete?
+  deleteMTOMilestone(id: UUID!): Boolean!
+  @hasRole(role: MINT_USER)
 }
--- a/pkg/sqlqueries/SQL/mto/milestone/delete.sql
+++ b/pkg/sqlqueries/SQL/mto/milestone/delete.sql
@@ -1,0 +1,1 @@
+DELETE FROM mto_milestone WHERE id = :id;
--- a/pkg/storage/mto_milestone.go
+++ b/pkg/storage/mto_milestone.go
@@ -76,6 +76,16 @@ func MTOMilestoneUpdate(np sqlutils.NamedPreparer, _ *zap.Logger, MTOMilestone *
 	return returned, nil
 }
 
+// MTOMilestoneDelete deletes an MTOMilestone in the database
+func MTOMilestoneDelete(np sqlutils.NamedPreparer, _ *zap.Logger, milestoneID uuid.UUID) error {
+	arg := map[string]interface{}{"id": milestoneID}
+	procErr := sqlutils.ExecProcedure(np, sqlqueries.MTOMilestone.Delete, arg)
+	if procErr != nil {
+		return fmt.Errorf("issue deleting MTOMilestone object: %w", procErr)
+	}
+	return nil
+}
+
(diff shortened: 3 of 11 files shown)


pkg/sqlqueries/SQL/mto/milestone/delete.sql
```diff
@@ -0,0 +1 @@
+DELETE FROM mto_milestone WHERE id = :id;
```
StevenWadeOddball, line 1: Q: Should this have a returning statement? I think we have it both ways in different parts of the app. I think there is some benefit in returning the record, but also fine to leave like this. I'm not sure if the FE ever uses the returned result

pkg/storage/mto_milestone.go
```diff
@@ -90,6 +90,16 @@ func MTOMilestoneUpdate(np sqlutils.NamedPreparer, _ *zap.Logger, MTOMilestone *
 	return returned, nil
 }
 
+// MTOMilestoneDelete deletes an MTOMilestone in the database
+func MTOMilestoneDelete(np sqlutils.NamedPreparer, _ *zap.Logger, milestoneID uuid.UUID) error {
```
StevenWadeOddball, line 94: To work with change history, we need to set a  Session Variable for the current user.  This comes from the Plan Collaborator delete method
	err := setCurrentSessionUserVariable(tx, userID)
	if err != nil {
		return nil, err
	}

pkg/graph/schema/types/mto_milestone.graphql
```diff
@@ -59,7 +59,8 @@ extend type Mutation {
   @hasRole(role: MINT_USER)
   updateMTOMilestone(id: UUID!, changes: MTOMilestoneChanges!): MTOMilestone!
   @hasRole(role: MINT_USER)
-  #TODO (mto) will we need any other mutations here? Can we ever delete?
+  deleteMTOMilestone(id: UUID!): Boolean!
```
StevenWadeOddball, line 62: Question: Should we return a milestone here instead of Boolean? It seems slightly more versatile to me.

StevenWadeOddball (changes requested):
Thanks Clay! This looks good functionally. I didn't validate the cascade behavior. I have one slight change request regarding setting a session variable for the delete statement. That's needed for change history.

pkg/graph/schema/types/mto_milestone.graphql
ClayBenson94: Added some comments in 53a587fe7e2caeacf386dd75fb0c92cc6cece6a0 -- but leaving this as-is for the sake of time 😁

pkg/storage/mto_milestone.go
ClayBenson94: This has been updated in 53a587fe7e2caeacf386dd75fb0c92cc6cece6a0 -- lemme know if that looks right to you, but I think it's all working as expected 👀

pkg/sqlqueries/SQL/mto/milestone/delete.sql
ClayBenson94: Same as below, tracked a comment to revisit this!

53a587fe7e2caeacf386dd75fb0c92cc6cece6a0

StevenWadeOddball (approved):
Thanks Clay! This looks good

PR #1565, opened 2024-12-11
[MINT-3241] Delete categories

OddTomBrooks:
# MINT-3241

## How to test this change

1.  

## PR Author Checklist

## PR Reviewer Guidelines


migrations/V193__Add_MTO_Category_Delete_Trigger.sql
```diff
@@ -0,0 +1,89 @@
...
+DROP FUNCTION IF EXISTS rebalance_milestones_before_category_delete();
+
+-- Create the trigger function that implements the custom cascading logic
+CREATE OR REPLACE FUNCTION rebalance_milestones_before_category_delete()
+  RETURNS TRIGGER AS $$
+DECLARE
+  cat_model_plan_id UUID;
+  is_top_level BOOLEAN;
+BEGIN
+  -- Determine if the category is top-level or a subcategory
```
StevenWadeOddball, line 47: Questions, does it matter if it is a parent category or not? Couldn't you always use the parent category id field? (for a top level field, that is null, so it would still correctly set it to null). That might simplify the logic. EG you always just set the milestone_id to the category parent_id

migrations/V193__Add_MTO_Category_Delete_Trigger.sql
StevenWadeOddball, line 1: This migration name needs to be updated, I think the next available one at the moment is 200

pkg/sqlqueries/SQL/mto/category/delete.sql
```diff
@@ -0,0 +1,12 @@
+DELETE FROM mto_category
```
StevenWadeOddball, line 1: Are you planning on making a cascade statement so subcategories automatically delete when their parent delete?

migrations/V193__Add_MTO_Category_Delete_Trigger.sql
StevenWadeOddball: eg, something like
```sql
BEGIN
    -- Update the milestones to reference the parent category or null
    UPDATE mto_milestone
    SET mto_category_id = OLD.parent_id
    WHERE mto_category_id = OLD.id;

    RETURN OLD;
END;
```

StevenWadeOddball (commented):
thanks Tom, this is looking good. I left a couple of questions and suggestions. Please let me know if you have any specific questions you want feedback on. Otherwise, I'll plan on doing a full review when you promote the PR.

migrations/V193__Add_MTO_Category_Delete_Trigger.sql
StevenWadeOddball: @OddTomBrooks , can you update this please? DB migrate is failing

pkg/storage/mto_category.go
```diff
@@ -82,6 +82,17 @@ func MTOCategoryCreate(np sqlutils.NamedPreparer, _ *zap.Logger, MTOCategory *mo
 	return returned, nil
 }
 
+// MTOCategoryDelete deletes an existing MTOCategory from the database
+func MTOCategoryDelete(np sqlutils.NamedPreparer, _ *zap.Logger, id uuid.UUID) error {
+	arg := map[string]interface{}{"id": id}
```
StevenWadeOddball, line 87: See MTOMilestoneDelete,

You need to set a transaction level variable of the user_id when deleting objects from the database in order to track  auditing information
```go
	err := setCurrentSessionUserVariable(tx, actorUserID)
	if err != nil {
		return err
	}
```

StevenWadeOddball (changes requested):
Thanks @OddTomBrooks , this is looking good. The app won't run currently because of the migration numbering. I provided some feedback, I will functionally test this when you have a chance to implement it

migrations/V193__Add_MTO_Category_Delete_Trigger.sql
```diff
@@ -0,0 +1,43 @@
...
+  RETURNS TRIGGER AS $$
+DECLARE
+  cat_ids UUID[];
+BEGIN
+  IF pg_trigger_depth() > 1 THEN
+    RETURN OLD;
+  END IF;
+
+  -- Recursively gather all descendant categories, including OLD.id
+  WITH RECURSIVE cat_tree AS (
```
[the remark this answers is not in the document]
OddTomBrooks: This has not been a simple insertion, I'll revisit it early tomorrow.

StevenWadeOddball (changes requested):
it looks like the unit tests for the delete are failing now. Could you take a look please and re-request review when it's corrected?

PR #1574, opened 2024-12-17
[Noref] collation migration fix

StevenWadeOddball:
## Description

Migrations were failiing in the deploy pipeline. Likely because a collation isn't dropped with DB clean. To address this, we just add a DROP if exists method in the migration. 

Future enhancements might look at dropping collations as part of DB clean, but this is the simplest solution for now

## How to test this change

1.  Verify tests pass. The true test is if the deploy pipeline functions as expected.

## PR Author Checklist

## PR Reviewer Guidelines


ClayBenson94 (approved):
Thanks @StevenWadeOddball!

PR #1577, opened 2024-12-18
[MINT-3226] Edit Solutions Linked to Milestones

OddTomBrooks:
# MINT-3226

## Description

Added support for optionally editing milestones to set linked solutions and an endpoint for directly setting linked solutions

## How to test this change

- Postman
  - Create a new MTO Milestone
  - Create a new MTO Solution
  - Edit the milestone, passing one or more linked solution IDs and common solution keys
  - Validate expected results
  - Experiment with removing linked solution IDs / common solution keys and validate the results
  - Use the set milestone linked solutions endpoint
  - Experiment with inputs and validate results

## PR Author Checklist

## PR Reviewer Guidelines


pkg/sqlqueries/SQL/mto/milestone/set_solution_links.sql
StevenWadeOddball, line 1: This is the SQL that I was suggesting you use a merge statement for (v195 is a good example). I think you are attempting the same sort of idea here, but this functionality does most of that for you in a succinct manner

pkg/storage/mto_milestone.go
```diff
@@ -132,3 +132,26 @@ func MTOMilestoneGetBySolutionIDLoader(np sqlutils.NamedPreparer, _ *zap.Logger,
...
+	commonSolutionKeys []models.MTOCommonSolutionKey,
+	createdBy uuid.UUID,
+) (*models.MTOMilestone, error) {
+	arg := map[string]interface{}{
+		"milestone_id":         milestoneID,
+		"solution_ids":         pq.Array(solutionIDs),
+		"common_solution_keys": pq.Array(commonSolutionKeys),
+		"created_by":           createdBy,
+	}
+	returned, procErr := sqlutils.GetProcedure[models.MTOMilestone](tx, sqlqueries.MTOMilestone.SetSolutionLinks, arg)
```
StevenWadeOddball, line 151: I suspect that it would make more sense to return the solutions instead of the milestone 🤔

StevenWadeOddball (commented):
just a couple high level thoughts! I haven't reviewed in great detail yet

pkg/graph/schema/types/mto_milestone.graphql
```diff
@@ -52,22 +52,26 @@ input MTOMilestoneChanges @goModel(model: "map[string]interface{}") {
...
   createMTOMilestoneCustom(modelPlanID: UUID!, name: String!, mtoCategoryID: UUID): MTOMilestone!
   @hasRole(role: MINT_USER)
   createMTOMilestoneCommon(modelPlanID: UUID!, commonMilestoneKey: MTOCommonMilestoneKey!, commonSolutions: [MTOCommonSolutionKey!]): MTOMilestone!
   @hasRole(role: MINT_USER)
-  updateMTOMilestone(id: UUID!, changes: MTOMilestoneChanges!): MTOMilestone!
+  updateMTOMilestone(id: UUID!, changes: MTOMilestoneChanges!, solutionLinks: MTOSolutionLinks): MTOMilestone!
   @hasRole(role: MINT_USER)
   deleteMTOMilestone(id: UUID!): Boolean!
   @hasRole(role: MINT_USER)
+  mtoMilestoneUpdateLinkedSolutions(id: UUID!, solutionLinks: MTOSolutionLinks!): MTOMilestone!
```
StevenWadeOddball, line 69: Maybe this should return an array of Solutions instead?

ClayBenson94: I'd be fine either way, honestly. Returning the milestone could even be preferred since the `id` we're passing in is for the milestone, so in a sense, it's the primary object being operated on!

This is also, more likely than not, another situation where the FE isn't even using the returned type, so I wouldn't worry too much about this in all honesty 😅

pkg/storage/mto_milestone.go
```diff
@@ -83,8 +83,8 @@ func MTOMilestoneCreate(np sqlutils.NamedPreparer, _ *zap.Logger, MTOMilestone *
 // NOTE: This method is used for updating both Custom and Library-sourced (Common) milestones. It WILL return an error
 // if you try to update one that has both custom info (like a `name`) and a CommonMilestoneKey
 // This is controlled not by application code, but by constraints initially added in this migration: migrations/V185__Add_MTO_Milestone.sql
-func MTOMilestoneUpdate(np sqlutils.NamedPreparer, _ *zap.Logger, MTOMilestone *models.MTOMilestone) (*models.MTOMilestone, error) {
```
StevenWadeOddball, line 86: I think that it's still appropriate to leave this a named preparer (eg either a tx or a store / db). Functionally, this is still fine for the PR, but I don't think we need to force a TX.

pkg/sqlqueries/SQL/mto/milestone_solution_link/merge_solutions_to_milestone.sql
```diff
@@ -0,0 +1,41 @@
...
+        milestone_id = :milestone_id
+        AND solution_id NOT IN (SELECT solution_id FROM source_links)
+    RETURNING solution_id
+)
+
+SELECT
+    s.id,
+    s.model_plan_id,
+    s.mto_common_solution_key,
+    s.name,
```
StevenWadeOddball, line 28: I noticed that this should join the common solution table, I'll push up a quick fix to ensure we get proper data.

StevenWadeOddball (commented):
Thanks Tom, this looks good, I noticed a couple small things, I'm going to push up a quick fix. I forgot to join the common solution table when returning solutions. I'll fix and push it up

pkg/sqlqueries/SQL/mto/milestone_solution_link/merge_solutions_to_milestone.sql
StevenWadeOddball: updated!

pkg/storage/mto_milestone.go
StevenWadeOddball: reverted to np

StevenWadeOddball (approved):
Thanks @OddTomBrooks , I think this is good to go!

PR #1579, opened 2024-12-19
[MINT-3240] FE work for Deleting Category from Action Menu

garyjzhao:
# MINT-3240

## Description

* Created `RemoveCategoryForm` component, test, copy, and mutations 
* Add `removeCategory` and `removeSubcategory` props to MTOContext and MTOModal 
* Add functions to ActionMenu to pop up remove category modal
* 👋🏼 

## How to test this change

1.  Navigate to MTO Matrix and hit the `...`
2. Click `Remove category` or `Remove sub-category`
3. Verify that modal matches figma 
4. hit the red button
5. verify that the category/subcategory selected is now gone

## PR Author Checklist

## PR Reviewer Guidelines


StevenWadeOddball (changes requested):
Hi Gary,

I did some functional testing instead of code review. I found that the option to remove the "uncategorized" subcategory exists even though that is an invalid operation. The same is true for the parent "uncategorized category.

image

I also found some behavior I'd love you to double check. Sometimes when I add a subcategory, then a milestone to that subcategory, and try to delete the parent category, it only deletes the subcategory instead. I got it to happen a few times, but it still is possible it is user error. I believe I clicked the parent category, but perhaps I clicked a sub on accident?
image

src/features/ModelPlan/ModelToOperations/_components/FormModal/index.tsx
```diff
@@ -69,16 +76,18 @@ const MTOModal = ({
...
+          modalType !== 'removeSubcategory' && (
+            <p className="margin-y-0 text-base">
+              <Trans
+                i18nKey={t('modal.allFieldsRequired')}
+                components={{
+                  s: <span className="text-secondary-dark" />
+                }}
+              />
+            </p>
+          )}
```
patrickseguraoddball, line 90: Not in love with having a prop for `isRequired` yet still needed a string literal conditional.  I understand that this modal is being used in the context provider and not explicitly passing down those props, but I'm tempted to request a better solution - as this type of conditional can get out of hand if we add more.  Maybe either remove the `isRequired` prop completely and have some type of mapped config for each of the modal forms?  We should also try to homogenize all the modals to use the context.  Right now we seem to be using the same component in two different ways (context provider and no context provider).  Maybe this can be tech debt but would rather them behave and implemented in a similar, expected fashion.  Let me know your thoughts if you this we should try to tackle this now.  As of now even for me its a bit confusing looking at the modal and it props, then at the context.  State is attempting to be handled in two different ways.

patrickseguraoddball (changes requested):

patrickseguraoddball (approved):

patrickseguraoddball (approved):

PR #1584, opened 2024-12-24
[MINT-3246] Suggested Milestone Banner on Home Page

garyjzhao:
# MINT-3246

## Description
* add `isSuggested` to modelToOperatinosMatrix gql
* Add `suggestedMilestone` to filter `isSuggested` milestones in modelToOperatinosMatrix data
* Conditional banner rendering 
* Shuffled around `GridContainer` in MTO home page to allow the suggested milestone to be full width (matching figma)

* Made changes to the header to make the header match figma more closely

## How to test this change

1.  Navigate to MTO Matrix
2. Verify that the banner's empty state matches figma
3. comment out line 49 and paste the following code in its place:
```
// TEMP only for demo
// UPDATED JANUARY 7th
  const demoData = useMemo(() => {
    const rawData = modelToOperationsMatrix?.commonMilestones || [];

    const testData = JSON.parse(JSON.stringify(rawData));

    if (testData.length > 0) {
      testData[0].isSuggested = true;
      testData[1].isSuggested = true;
      testData[0].isAdded = true;
    }

    return testData;
  }, [modelToOperationsMatrix?.commonMilestones]);

  const suggestedMilestones =
    demoData.filter(obj => obj.isSuggested && !obj.isAdded) || [];
```
4. reload page and verify that banner's not empty state matches figma
5. try to break it

## PR Author Checklist

## PR Reviewer Guidelines


patrickseguraoddball (changes requested):
This doesn't seem to do anything.
Screenshot 2025-01-03 at 9 44 36 AM

The pluralization on this is incorrect.  It should say "There is 1".  You can you the key modifier in the translation file `_other` to automatically handle pluralization.
Screenshot 2025-01-03 at 9 45 13 AM

It seems that nothing is clickable.  Maybe there's some zindex affecting it.  The `view suggested milestones` is not clickable
![Screenshot 2025-01-03 at 9 47 04 AM](https://github.com/user-attachments/assets/d72ace4a-f40

I would maybe consider extracting this banner work into it's own component, I can see us using this in a different place.  Not necessary but just a suggestion.  It could also use a test to ensure it's functionality.  Up to you

patrickseguraoddball (changes requested):
Screenshot 2025-01-07 at 11 55 02 AM

I added the suggested milestone, and it's still showing the banner with that some 1 milestone being suggested.  

I think you need to check for both if it's suggested, and if it's already added to the matirx `isAdded`

patrickseguraoddball (approved):
LG thanks

PR #1585, opened 2024-12-26
[MINT-3259] Allow linking milestone on add common solution

OddTomBrooks:
# MINT-3259

## Description

This PR allows for adding milestone links when adding common solutions

## How to test this change

1. Run scripts/dev test:go and validate unit test success
2. Run Postman endpoint to add common solution and validate that milestone links are added as expected

## PR Author Checklist

## PR Reviewer Guidelines


StevenWadeOddball (changes requested):
Thanks Tom. This functions as expected for adding a common solution, but ideally this functionality should be extensible for the update solution modal as well. ( similarly to the recent milestones work).

We should also double check if the work to link solutions when editing a solution is captured anywhere else. I thought it might be included as part of this work, if not here, we should make sure it is captured and uses whatever shared logic is developed here.

StevenWadeOddball:
Note, [MINT-3142](https://jiraent.cms.gov/browse/MINT-3142) will cover editing a solution. This work should still enable that with shared functionality, but we can make sure to write a task to update the solution update mutation.

pkg/graph/resolvers/mto_solution.go
```diff
@@ -110,7 +115,32 @@ func MTOSolutionCreateCommon(
 		return nil, err
 	}
 
-	return storage.MTOSolutionCreate(store, logger, mtoSolution)
+	return sqlutils.WithTransaction[models.MTOSolution](store, func(tx *sqlx.Tx) (*models.MTOSolution, error) {
+		solution, err := storage.MTOSolutionCreate(tx, logger, mtoSolution)
```
StevenWadeOddball, line 119: This is the section of code I'm referring to. You are currently looping through and creating links, and calling a store method for each link.  The solution for linking solutions to a milestone that takes place in `MTOMilestoneUpdateLinkedSolutionsWithTX` is reusable. It also reduces the amount of store calls. 

Currently, this code would not be useable for the update methods. By having a single resolver which can update all the milestones linked to a solution, we could reuse the logic for update. Otherwise, we'd need to refactor this when we implement the functionality for editing a solution.

pkg/sqlqueries/mto_milestone_solution_link.go
```diff
@@ -11,17 +11,23 @@ var mtoMilestoneGetMilestoneSolutionLinksByMilestoneIDSQL string
...
+
 type mtoMilestoneSolutionLinkScripts struct {
 	// creates a link between a milestone and a solution and returns the link ID
 	Create string
 	// returns all milestone-solution links for a given milestone by ID
 	GetByMilestoneID string
 	// Takes an array of solution_ids and a milestone_id and performs a merge to ensure only these solutions are listed to that milestone
 	MergeSolutionsToMilestones string
+	// Takes an array of milestone_ids and a solution_id and performs a merge to ensure only these milestones are listed to that solution
+	LinkMilestonesToSolution string
```
StevenWadeOddball, line 25: nit: maybe these should use the same verb as it does for the Milestone centric query? (It might be clearer when we come back to it later that they are doing similar things

pkg/sqlqueries/SQL/mto/milestone_solution_link/link_milestones_to_solution.sql
StevenWadeOddball, line 1: query looks good!
nit: just around naming the file consistent with the other related query.

StevenWadeOddball (approved):
🚀  Looks great! Thanks for making that change @OddTomBrooks , I think this will put us in a great place for the edit solution ticket. I put a couple of nits around file / variable naming. Up to you to if you want to change or not, it functions great.

pkg/sqlqueries/mto_milestone_solution_link.go
OddTomBrooks: This was a conscious decision in that I feel the previous naming convention has some issues. I resisted the urge to rename old methods, perhaps as a 1 point tech lift later.

PR #1590, opened 2024-12-31
[MINT-3266] Get all milestones by model plan ID with no solution link

OddTomBrooks:
# MINT-3266

## Description

1.  Added an endpoint to retrieve all MTO milestones with no linked solution

## How to test this change

- Use Postman
  - Create a model plan
  - Create an MTO milestone
  - Fetch all MTO milestones with no linked solution, validate that your milestone is retrieved
  - Link your MTO milestone with a custom solution (may require manual winding in the DB)
  - Fetch all MTO milestones with no linked solution, validate that your milestone is not retrieved
  - Unlink your MTO milestone from the custom solution (by deleting the corresponding entry in the custom linking table)
  - Fetch all MTO milestones with no linked solution, validate that your milestone is retrieved
  - Link your MTO milestone with a common solution
  - Fetch all MTO milestones with no linked solution, validate that your milestone is not retrieved
  - Unlink your MTO milestone from the common solution
  - Fetch all MTO milestones with no linked solution, validate that your milestone is retrieved

## PR Author Checklist

## PR Reviewer Guidelines


MINT.postman_collection.json
```diff
@@ -3418,6 +3418,39 @@
...
+									}
+								}
+							],
+							"request": {
+								"method": "POST",
+								"header": [],
+								"body": {
+									"mode": "graphql",
+									"graphql": {
+										"query": "query GetMTOMilestone {\n    mtoMilestonesWithNoLinkedSolutions(ids: [\"{{mtoMilestoneID}}\"]) {\n    id\n    name\n    key\n    addedFromMilestoneLibrary\n    facilitatedBy\n    needBy\n    status\n    riskIndicator\n    # isDraftMilestone\n    mtoCategoryID\n    categories {\n        category{\n            name\n            position\n        }\n        subCategory{\n            name\n            position\n        }\n    }    \n\n    # commonMilestone\n    # solutions\n    # category\n    # subCategory\n    }\n}",
```
StevenWadeOddball, line 3441: This should have this signature, it's invalid curently

```
    mtoMilestonesWithNoLinkedSolutions(modelPlanID:"{{modelPlanID}}") {
```

pkg/graph/schema/types/mto_milestone.graphql
```diff
@@ -66,5 +66,7 @@ extend type Mutation {
 extend type Query {
   mtoMilestone(id: UUID!): MTOMilestone!
   @hasAnyRole(roles: [MINT_USER, MINT_MAC]) #TODO (mto) is there any reason a MAC user shouldn't query this?
+  mtoMilestonesWithNoLinkedSolutions(modelPlanID: UUID!): [MTOMilestone!]!
```
StevenWadeOddball, line 69: I suspect this should live on the mtoMatrix under model plan, instead of directly on the top level query.

pkg/sqlqueries/SQL/mto/milestone/get_by_model_plan_id_no_linked_solution_LOADER.sql
```diff
@@ -0,0 +1,40 @@
...
+LEFT JOIN mto_common_milestone AS cm
+    ON m.mto_common_milestone_key = cm.key
+
+/* LEFT JOIN to see if this milestone is directly linked to any solution */
+LEFT JOIN mto_milestone_solution_link AS link
+    ON m.id = link.milestone_id
+
+/* LEFT JOIN to see if the milestone's mto_common_milestone_key is
+     found in the common link table */
+LEFT JOIN mto_common_milestone_solution_link AS cLink
```
StevenWadeOddball, line 31: Hmm, I'm not sure that this is a desired behavior. From the AC, I think we are only looking for milestones that aren't directly linked to this specific model plan. `mto_common_milestone_solution_link` is a configuration table to identify solutions that are commonly used for a milestone.

pkg/sqlqueries/SQL/mto/milestone/get_by_model_plan_id_no_linked_solution_LOADER.sql
```diff
@@ -0,0 +1,40 @@
...
+    ON m.mto_common_milestone_key = cLink.mto_common_milestone_key
+
+WHERE
+    /* Only milestones whose model_plan_id is in the QIDS set */
+    m.model_plan_id IN (SELECT model_plan_id FROM QIDS)
```
StevenWadeOddball, line 36: I think a join would be more appropriate here. We do it most places with a line like this (note you will need to adapt the names to conform since your CTE already has a name. 

`INNER JOIN QUERIED_IDS AS qIDs ON mto_milestone.model_plan_id = qIDs.model_plan_id`

pkg/sqlqueries/SQL/mto/milestone/get_by_model_plan_id_no_linked_solution_LOADER.sql
```diff
@@ -0,0 +1,40 @@
+WITH QIDS AS (
```
StevenWadeOddball, line 1: Perhaps a verbose name is useful here? we often call this CTE Queried_IDS, then alias it in a join. QIDS doesn't have much intrinsic meaning by itself.

StevenWadeOddball (changes requested):
Thanks Tom, this is looking pretty good, I think it's almost there. I have some suggestions for the query. Only one piece is a business logic concern ( linking to the common_milestone_solution_link table).

pkg/graph/schema/types/mto_milestone.graphql
ClayBenson94: And, presumably, in that change we could drop the `modelPlanID` param, yeah?

OddTomBrooks: Yes, this is done locally and will be pushed up shortly. We can source from the mto matrix instead.

pkg/graph/resolvers/mto_milestone_test.go
```diff
@@ -516,3 +522,65 @@ func (suite *ResolverSuite) TestMTOMilestoneUpdateLinkedSolutions_UnlinkByCommon
...
+	nonLinkedMilestones := suite.getMilestonesWithNoLinkedSolution(plan.ID)
+	milestoneIDs := lo.Map(nonLinkedMilestones, func(m *models.MTOMilestone, _ int) uuid.UUID {
+		return m.ID
+	})
+
+	suite.Contains(milestoneIDs, unlinkedMilestone.ID, "Unlinked milestone should be present")
+	suite.NotContains(milestoneIDs, linkedMilestone.ID, "Linked milestone should be excluded from the results")
+}
+
+func (suite *ResolverSuite) TestMTOMilestoneNoLinkedSolutions_MultiplePlans() {
```
StevenWadeOddball, line 551: @OddTomBrooks , I'm not sure that this is actually verifying the loader.  If you take a look at `TestMTOMilestoneGetBySolutionIDLoader`, the helper function takes a list of keys, and expected results. It then uses go routines to ensure that test cases hit the data loader at the same time, could you  try to use that helper method here please?

pkg/graph/resolvers/mto_milestone_test.go
```diff
@@ -516,3 +522,65 @@ func (suite *ResolverSuite) TestMTOMilestoneUpdateLinkedSolutions_UnlinkByCommon
 	suite.NoError(err)
 	suite.Len(linkedSolutions, 0)
 }
+
+func (suite *ResolverSuite) TestMTOMilestoneNoLinkedSolutions_SinglePlan() {
```
StevenWadeOddball, line 526: This probably doesn't hurt, but I'm not sure you need it if you have multiple test cases in the other loader. Perhaps it's worthwhile to test a case when no results will be returned by SQL at all?

pkg/graph/resolvers/resolver_test.go
```diff
@@ -408,6 +408,32 @@ func (suite *ResolverSuite) createMilestoneCommon(
 	return milestone
 }
 
+func (suite *ResolverSuite) getMilestonesWithNoLinkedSolution(
```
StevenWadeOddball, line 411: I don't think this function is needed. These functions here are generally are setting up test data. This is abstracting a fetch. I think reworking the unit test to use `VerifyLoaders` would negate the use of this as well.

StevenWadeOddball (changes requested):
Thanks @OddTomBrooks, functional testing looks good. Can you please revisit the unit test? We have a a utility function to standardize data loader unit testing. It should be quick change, let me know if you have any questions about how the utility function works `TestMTOMilestoneGetBySolutionIDLoader` is a good example of how to use it

StevenWadeOddball (approved):
Thanks @OddTomBrooks

PR #1593, opened 2025-01-06
[MINT-3267] Preliminary MTO Change History Work

StevenWadeOddball:
# MINT-3267

## Description

This PR introduces serves to enable change history from the backend.
1. Enable the base auditing trigger on all relevant tables
2. Add missing translations
3. Correct locations in MTO that were not correctly setting modified properties (and causing translations to fail)
4. Correct any translations that were causing validation failures

## How to test this change

As always, you can translate audits locally by running the db:seed:interactive command and selecting translate. Translated audits are fetch-able through Postman queries.

1.  Ensure you can create update delete mto entities as previously.
2. Looking at the translated audits, verify if they look to you like they will be sufficient for rendering on the front end. 
    a. If something comes up later when developing the frontend that was missed now, we can address it at that point. 

## PR Author Checklist

## PR Reviewer Guidelines


mappings/translation_utilities.go
```diff
@@ -49,6 +49,23 @@ func GetTranslation(tableName models.TableName) (Translation, error) {
...
+	case models.TNMTOMilestoneSolutionLink:
+		return MTOMilestoneSolutionLinkTranslation()
+	case models.TNMTOInfo:
+		return MTOInfoTranslation()
+	case models.TNMTOCommonSolutionContact:
+		return MTOCommonSolutionContactTranslation()
+	// case models.TNMTOCommonMilestone:
+	// 	return MTOCommonMilestoneTranslation()
+	// case models.TNMTOCommonSolution:
+	// 	return MTOCommonSolutionTranslation()
```
OddTomBrooks, line 68: Are these still relevant?

OddTomBrooks (commented):
I haven't done functional testing yet. The code appears fine, I just have a question around some commented out logic. After functional testing I anticipate approving the PR.

mappings/translation_utilities.go
StevenWadeOddball: Thanks! No, I deleted it. They were previously going to be used, but we are trying to use change history without this meta data. We will add back in if needed later.

OddTomBrooks (approved):
Functional testing seems to work

PR #1595, opened 2025-01-09
[MINT-3236] Solution milestone linking

patrickseguraoddball:
# MINT-3236

## Description

The work for this linking was a bit complicated.  The query to fetch solution links returns created records all with IDs.  The suggested solutions (which are all part of the common solutions) need to be separated and checked independently of the multiselect.  Cards that are not checked, do not have an id, and will populate the mutation with an `MTOCommonMilestoneKey`.  The multiselect is separated into the remaining common milestones (minus suggested) and custom milestones.  These selections are stored in app state, from the parent `EditMilestoneForm` props.  The mutation for linking solutions is separated into `solutionIDs` and `commonSolutionKeys`

- Added table component on edit milestone panel
- Added modal for editing solution links for a milestone
- Extracted `SolutionCard` to its own component to be used both with and without a form component/checkbox
- Added support for unsaved changes
- Unit tests

## How to test this change

1.  Seed DB
2. Navigate to the MTO plan
3. Add a custom solution
4. Add a milestone with solutions selected
5. Edit the milestone from the table
6. Verify the solution table is correct with previously added solutions
7. Click to edit the milestone, tinker around with added and removing
8. Go back to edit and verify all looks correct.
9. Try to exit without saving
10. Save
11. Try to break again

## PR Author Checklist

## PR Reviewer Guidelines


garyjzhao (approved):
this is good stuff. LGTM

PR #1607, opened 2025-01-17
[MINT-3260] Solution Library

garyjzhao:
# MINT-3260

## Description

* Added `SolutionLibrary` component, copy, test, snaps, query, mocks
* Added `MTOSolutionCard` component
* Added `modalCalledFrom?` to MTOModalContext
* Moved `MTOModalProvider` out of MTOHome to MTO Route switcher
* 

## How to test this change

1.  Seed and Nav to MTO
2. `Browse Solution Library`
3. Verify it matches figma
4. Verify functionality works, like search filter, tabbed nav, hide added solution, query params, add a custom solution (notably the text in the modal) 

## PR Author Checklist

## PR Reviewer Guidelines


src/features/ModelPlan/ModelToOperations/SolutionLibrary/index.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+              <SolutionLibrary />
+            </MessageProvider>
+          </Route>
+        </MemoryRouter>
+      </MockedProvider>
+    );
+
+    // Match the snapshot
+    expect(asFragment()).toMatchSnapshot();
+  });
```
patrickseguraoddball, line 35: We need to assert the async data has loaded here.  The snapshot just renders a loading component

patrickseguraoddball (changes requested):
Nice work Gary :) 

Moving to the last page, then choosing to show all solutions from the table footer does reset the pagination, and instead shows an empty page.

Screenshot 2025-01-17 at 12 37 49 PM

Same goes for when you are switching solution types, if you have moved to a subequent page.  Pagination needs to be reset for all these actions
Screenshot 2025-01-17 at 12 38 04 PM

On mobile this should switch to a dropdown select - per Figma - https://www.figma.com/design/BcI7pNwfkmwaDD2EGGTsYC/IT-Lead?node-id=5833-29952&t=bFPAdURrIu60Qyd2-0
Screenshot 2025-01-17 at 12 40 53 PM

The `about this solution` should render underneath the add button, not to the side on mobile
Screenshot 2025-01-17 at 12 41 33 PM

This alert references operational solutions.  Hopefully this wasnt the case when i implements milestone library as well.  If so could you update both
Screenshot 2025-01-17 at 12 43 57 PM

garyjzhao:
@patrickseguraoddball addressed all but one of your requests. the alert the refers to operational solutions matches the copy i see from figma: https://www.figma.com/design/BcI7pNwfkmwaDD2EGGTsYC/IT-Lead?node-id=4912-169052&t=sGsLLuSSBJwtzhID-0

image

are you referring to somewhere else?

patrickseguraoddball (changes requested):
This is still showing blank when latter page is clicked, then choosing to view any of the other segmented button tabs


I think in liew of "Rows" is should say something like "Show 6 solutions". Double check with UX but that's what they directed for the MTO matrix

patrickseguraoddball (approved):

PR #1608, opened 2025-01-17
[MINT-3270] Updating milestones linked to an MTO solution

OddTomBrooks:
# MINT-3270

## Description

This feature allows us to link milestones by ID to a solution. We do this intentionally in two places, one specific resolver for updating linked milestones and an integration of that resolver into the update solution mutation. Per discussion, unlike with milestones, we do not support linking by common key - only by milestone ID.

## How to test this change

1. Open Postman
2. Create a custom milestone and custom solution
3. Validate that execution of UpdateMTOSolutionLinkedMilestones returns an array of one linked milestone
4. Remove the "{{mtoMilestoneID}}" entry from the input array in the query
5. Validate that execution of the same query returns null (inferred as an empty slice internally)
6. Open the UpdateMTOSolution query
7. Validate the same behaviors again, checking against the newly added milestones field at the bottom of the result set
8. Validate execution of unit tests

## PR Author Checklist

## PR Reviewer Guidelines


StevenWadeOddball (changes requested):
@OddTomBrooks, can you please add testing instructions, update the postman collection and correct merge conflicts? 

I will review in detail when these are addressed.

OddTomBrooks:
> @OddTomBrooks, can you please add testing instructions, update the postman collection and correct merge conflicts?
> 
> I will review in detail when these are addressed.

Yes, I'm working on the merge conflicts at the moment. For whatever reason, draft wasn't showing merge issues until promoted. Once that's done I'll update testing spec. Postman collection just needs to be exported, good callout.

OddTomBrooks:
@StevenWadeOddball Ready for review! 🥳

pkg/graph/resolvers/mto_solution.go
```diff
@@ -214,3 +241,59 @@ func MTOSolutionDelete(ctx context.Context, logger *zap.Logger, principal authen
 		return nil
 	})
 }
+
+// MTOSolutionUpdateLinkedMilestonesWithTX updates the linked milestones for a solution, deleting ones that are not included in the list
+func MTOSolutionUpdateLinkedMilestonesWithTX(
```
StevenWadeOddball, line 246: @OddTomBrooks, I think these new methods aren't needed. The work you did for the create common milestone should suffice here. I don't think these should / are doing anything different. Can you unify these please? (`MTOSolutionLinkMilestonesWithTX` is the matching method)

pkg/storage/mto_solution.go
```diff
@@ -183,3 +183,24 @@ func MTOSolutionDelete(tx *sqlx.Tx, actorUserID uuid.UUID, _ *zap.Logger, milest
 	}
 	return nil
 }
+
+func MTOSolutionLinkMilestonesToSolutions(
```
StevenWadeOddball, line 187: Similar to the message in resolvers, this is duplicating functionality (`MTOMilestoneSolutionLinkMilestonesToSolution`).

pkg/graph/resolvers/mto_solution_test.go
```diff
@@ -157,3 +158,50 @@ func (suite *ResolverSuite) TestCreateCommonSolutionAndLinkMilestones() {
...
+		milestoneLinks,
+	)
+	suite.NoError(err)
+
+	// Verify the solution is linked now
+	linkedSolutions, err := storage.MTOMilestoneSolutionLinkGetByMilestoneID(suite.testConfigs.Store, suite.testConfigs.Logger, milestone.ID)
+	suite.NoError(err)
+	suite.Len(linkedSolutions, 1)
+	suite.Equal(sol.ID, linkedSolutions[0].SolutionID)
+}
```
StevenWadeOddball, line 207: Since this resolver also is capable of unlinking solutions, this should probably test the unlinking behavior, as well as the behavior when the links input is not provided.

pkg/graph/schema/types/mto_solution.graphql
```diff
@@ -46,12 +44,18 @@ input MTOSolutionChanges @goModel(model: "map[string]interface{}") {
...
+
 extend type Mutation {
   createMTOSolutionCustom(modelPlanID: UUID!, solutionType: MTOSolutionType!,  name: String!, pocName: String!, pocEmail: String!): MTOSolution!
   @hasRole(role: MINT_USER)
   createMTOSolutionCommon(modelPlanID: UUID!, key: MTOCommonSolutionKey!, milestonesToLink: [UUID!]): MTOSolution!
   @hasRole(role: MINT_USER)
-  updateMTOSolution(id: UUID!, changes: MTOSolutionChanges!): MTOSolution!
+  updateMTOSolution(id: UUID!, changes: MTOSolutionChanges!, milestoneLinks: MTOMilestoneLinks): MTOSolution!
+  @hasRole(role: MINT_USER)
+  mtoSolutionUpdateLinkedMilestones(id: UUID!, milestoneLinks: MTOMilestoneLinks!): [MTOMilestone!]
```
StevenWadeOddball, line 58: I don't think this endpoint is needed, just the update endpoint. I think it should be removed unless needed by the frontend. It can be added later if needed

StevenWadeOddball (changes requested):
Thanks Tom,  functionally this is working, but there are some issues with code duplication and incorrect user account referencing.

(The created links are marked as the unknown user account instead of the modified by user)
image

I checked, and it is common from the storage method `MTOSolutionLinkMilestonesToSolutions` not actually passing the created_by parameter to the sql.  Using the other created code should address this though as well.

pkg/graph/schema/types/mto_solution.graphql
OddTomBrooks: This is the same behavior as we added for linking solutions to milestones. In the interest of consistency I think it's best to keep the same structure for parity. Additionally, it allows for more atomic endpoint testing. Here you can validate specific behavior without concern for integration into the update mutation.

pkg/graph/resolvers/mto_solution_test.go
OddTomBrooks: Fair enough

StevenWadeOddball (approved):
I tested functionality and removed unused code

PR #1658, opened 2025-04-18
MINT-3311/remove operational need and solutions triggers

StevenWadeOddball:
# MINT-3311

## Description

This PR removes triggers and trigger functions used for operational needs and solutions

It also removes any mutation for operational needs, solutions or subtasks, and removes the resolver and test codes that rely on any input types defined in model. If there are concerns about this, we can hold off on those changes for now and just keep the trigger logic changes for the moment.

The Front End changes are not yet implemented, so this will break as previous used mutations were removed.

## How to test this change

1.  Make sure this deletions don't cause issues in the application

## PR Author Checklist

## PR Reviewer Guidelines


patrickseguraoddball (approved):
Changes looked good/ I removed everything from the FE aside from translation.  Not sure if you wanted to gut the queries from the BE as a part of this, but will approve anyway.

StevenWadeOddball:
@patrickseguraoddball , thanks for removing tagging a possible solution! This looks good!

PR #1662, opened 2025-04-24
[MINT-3152] Export/Share

patrickseguraoddball:
# MINT-3152

## Description

Screenshot 2025-04-24 at 9 20 38 AM

- Added ability to export  MTO sections in both csv and PDF
  - Extended translation type to be a ble to dynamically flatten certain data when exporting
  - `ex: milestone.solutions = [{name:  "Solution 1"}] => milestone.solutions = ["Solution 1"]`
- Tweaked a lot of existing export rendering on PDF for better UI
- Added no-print classes and print context to better conditionall show/hide content
- Added export labels for better UX/understanding when exporting
- Share should work for each section and emails should link to relevant read view sections

## How to test this change

1.  Log into MINT and fill out lots of varied data within MTO and reall all of model plan
2. Export as csv and PDF for each section in dropdown
3. Share each section, verify in mailcatcher, verify links back to MINT work

## PR Author Checklist

## PR Reviewer Guidelines


src/features/ModelPlan/ReadOnly/_components/FilterView/util.tsx
```diff
@@ -1,5 +1,17 @@
 import React from 'react';
 
+export const filterGroupParams = [
```
StevenWadeOddball, line 3: Can this section reference the GQL type instead? (Or is it not possible because of casing or some other reason?)

patrickseguraoddball: Thanks Steven, yeah they need to be converted to lowercase to key of the url routes.  I went ahead and did that mapping

patrickseguraoddball:
Not sure why this cypress test is failing in CI, I can't replicate locally.  Looking further

StevenWadeOddball:
> Not sure why this cypress test is failing in CI, I can't replicate locally. Looking further

I think that this might have appeared when removing operational solutions. Maybe the discussion reply was related to tagging a possible solution?

StevenWadeOddball (approved):
Thanks Patrick, 

This looks great! A very involved problem to solve! I noticed that if ALL MTO info is exported, it seems that the first Milestone is entered, and then the row is shared and reused for all solutions for the Matrix. Is this a requirement of the library? I guess that it is also because it is really trying to display multiple entities in one table. 

I think this is fine, you can always export the data individually if you want them separately.

Also, do you expect that a link to the read-view would also scroll to the expected section? I verified that the links seem to be working as expected from mailcatcher, and opening the correct url. I'm not sure if you'd expect the view to scroll down if you provided an app section.

patrickseguraoddball:
Oh yeah I had seen what you are talking about with the milestone issue.  I think that's default bahavior, I'll look to see if there is config for it, but I dont see it as a dealbreaker.  If it's a pain I'll just leave as is.  CSV is really just supposed to be a data dump to get the data out anyway

PR #1708, opened 2025-05-14
[NOREF] Remove Existing CMS data

StevenWadeOddball:
# MTO QA, remove exiting_cms Data

## Description

Existing cms data is not a valid option any longer for a solution. The data was already removed from the DB, but persisted in GQL. This PR removes that reference. (See finding in [slack](https://cmsgov.slack.com/archives/C01B9TEC65A/p1747242187387529))

## How to test this change

1.  Confirm that existing cms data is no longer an option as a solution in the milestone drop down.

## PR Author Checklist

## PR Reviewer Guidelines


garyjzhao (changes requested):
Existing CMS Data solution is no longer on the dropdown nor common solutions. 

There is a two more spots in the app that has `EXISTING_CMS_DATA_AND_PROCESS`:
1. src/i18n/en-US/modelPlan/solutions.ts:125
2. src/i18n/en-US/modelPlan/opSolutionsMisc.ts:230

StevenWadeOddball:
> Existing CMS Data solution is no longer on the dropdown nor common solutions.
> 
> There is a two more spots in the app that has `EXISTING_CMS_DATA_AND_PROCESS`:
> 
> 1. src/i18n/en-US/modelPlan/solutions.ts:125
> 2. src/i18n/en-US/modelPlan/opSolutionsMisc.ts:230

Thanks @garyjzhao , I think those references are for the historic operational solution data (Does that sound right to you?)

I'm leaving all that data so we can continue to translate for those historic references for operational solution data, and just removing from MTO. In that light, does this look correct to you?

garyjzhao (approved):
Thanks for the clarification. LGTM

PR #1716, opened 2025-05-19
[NOREF] - QA fixes # 76 and 78

yiranli624:
# MINT-0000

## Description

QA fixes # 76 and 78. 
FIx help article content

Screenshot 2025-05-19 at 8 42 06 AM Screenshot 2025-05-19 at 8 42 16 AM

Screenshot 2025-05-19 at 9 26 53 AM

Screenshot 2025-05-19 at 11 33 57 AM

## How to test this change

1.  Verify fixes address issues in QA doc

## PR Author Checklist

## PR Reviewer Guidelines

### DIFF
--- a/src/i18n/en-US/modelPlan/modelToOperations.ts
+++ b/src/i18n/en-US/modelPlan/modelToOperations.ts
@@ -549,7 +549,7 @@ export const modelToOperationsMisc: Record<string, any> = {
       milestone: 'Milestone',
       helpText: 'Select all that apply',
       status: 'Status',
-      selectedMilestoneCount_other: '{{count}} selected milestones',
+      selectedMilestoneCount_other: '{{count}} associated milestones',
       selectedMilestoneCount: '{{count}} associated milestone',
       relatedMilestoneCount_other: '{{count}} related milestones',
       relatedMilestoneCount: '{{count}} related milestone',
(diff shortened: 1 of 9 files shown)


src/i18n/en-US/modelPlan/modelToOperations.ts
```diff
@@ -547,9 +547,10 @@ export const modelToOperationsMisc: Record<string, any> = {
       noMilestones:
         'You haven’t identified any specific milestones that this solution will help implement.',
       milestone: 'Milestone',
+      helpText: 'Select all that apply',
       status: 'Status',
       selectedMilestoneCount_other: '{{count}} selected milestones',
-      selectedMilestoneCount: '{{count}} selected milestone',
+      associatedMilestoneCount: '{{count}} associated milestone',
```
patrickseguraoddball, line 553: I'd be hesitant to change keys of translations when we only need to change the text.  We dont always have a full grasp of where these keys are being used, and it can lead to unmapped text in the app.  When you need to update text, generally just update the strings, it leads to less human error. 

Also the way the translations work is the `_other` is very important in interpolating plurals of text.  This as is, is not accurately translating the plural form of the count.  I would recomment reverting the renaming the the key, and the pluralization should work.  

`selectedMilestoneCount_other` is counting on a `selectedMilestoneCount` to exist, but when you changed it, it does not exist.

patrickseguraoddball (changes requested):
Thanks Elle, just one thing

src/i18n/en-US/modelPlan/modelToOperations.ts
patrickseguraoddball: Totally gotcha, thanks

patrickseguraoddball (approved):
LG thanks Elle!

PR #1729, opened 2025-05-27
[NOREF] Fixed cypress past dates/status trigger, removed unused flags

patrickseguraoddball:
# NOREF

## Description

- Removed unused/legacy flags that were causing issues with cypress
- Removed echimp BE LD config
- Updated cypress dates

There were dates set in the `modelPlan` cypress test that were in the past.  This cause a status update modal to trigger and hide necessary components to test the suite.  This pushed those dates very far into the future.

## How to test this change

1.  Verify cypress tests now pass

## PR Author Checklist

## PR Reviewer Guidelines


yiranli624 (dismissed):
Thank you for fixing it!

PR #1739, opened 2025-05-30
[MINT-3388 ]Fix go back from change history display incorrect wording

yiranli624:
…and update display string]

# MINT-3388

## Description

When clicking on a model’s “View change history” link in the read view, the go back button incorrectly says “Back to the model collaboration area”. That link is fine for when navigating to change history from the collab area as a model team member, but I don’t think we want it saying that for our read-only users as they won’t know what the collab area is.
As read only user, I should see 'Back to model information' (previously as 'Back to the Read view") when viewing change history.

After fix:

https://github.com/user-attachments/assets/258cd76e-ccc5-42e1-9bc4-2eff8690112b

## How to test this change

1.  login as a read only user
2. nav to change history
3. verify back button should display 'Back to model information' 

## PR Author Checklist

## PR Reviewer Guidelines


Peyton232 (approved):
Looks good, I approve

PR #1769, opened 2025-06-16
[NOREF] Status banner fallback creation date

patrickseguraoddball:
# NOREF

## Description

The component was built with the assumption that there would always be a change history record, but models created before 7/30/24 may not have a record as they existed before CH 

- Adds a fallback `createdDts` for all dates rendering `mostRecentEdit`

## How to test this change

1.  - Can add `const mostRecentEdit: any = {};` here:
     - https://github.com/CMS-Enterprise/mint-app/blob/f826130cffdd602c8c62e19b7a8736b53a6f7bbe/src/features/ModelPlan/CollaborationArea/index.tsx#L76
2. Then comment out mostRecentEdit from the fetched data destructure. 
3. Verify the `createdDts` is used as fallback

## PR Author Checklist

## PR Reviewer Guidelines

### DIFF
--- a/src/components/ShareExport/pdfSummary.tsx
+++ b/src/components/ShareExport/pdfSummary.tsx
@@ -43,7 +43,7 @@ const PDFSummary = ({
     }
   });
 
-  const { abbreviation, modelName, mostRecentEdit, status } =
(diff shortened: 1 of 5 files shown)


yiranli624 (approved):
lgtm!

PR #1785, opened 2025-06-27
[MINT-3382] Contractor modal

yiranli624:
# MINT-3382

## Description

Add and edit contractor on points of contact page

## How to test this change

1.  Able to add, edit, and remove contractors
2. Match [figma](https://www.figma.com/design/dzzLdDf6UwkLA51BtoAt3T/MINT---Knowledge-Center?node-id=8860-12240&t=WYmGFry5qDT0jiAr-0)

## PR Author Checklist

## PR Reviewer Guidelines


src/features/HelpAndKnowledge/SolutionsHelp/SolutionDetails/Solutions/Generic/_components/AddContractorForm/index.tsx
```diff
@@ -0,0 +1,188 @@
...
+  const [hasMutationError, setHasMutationError] = useState(false);
+
+  if (!selectedSolution) {
+    return null;
+  }
+
+  const onSubmit = (formData: FormValues) => {
+    create({
+      variables: {
+        key: selectedSolution.key.toUpperCase() as MtoCommonSolutionKey,
```
patrickseguraoddball, line 64: This needs to be addressed like the previous PR

src/features/HelpAndKnowledge/SolutionsHelp/SolutionDetails/Solutions/Generic/_components/AddContractorForm/index.tsx
```diff
@@ -0,0 +1,188 @@
...
+        contractorTitle: formData.contractorTitle,
+        contractorName: formData.contractorName
+      }
+    })
```
patrickseguraoddball, line 68: There's a mutation error when not populating Contractor title (which is not required).  The paradigm is not pass the property at all in the payload if it has not been populated.  We use a helper util `dirtyInput` that returns only dirty input to the payload of the mutation.  

Here we do not want to include the property `contractorTitle` in the payload.  
Screenshot 2025-07-01 at 7 26 05 AM

If `contractorTitle` was once populated and we want to remove that value, instead of passing empty string, I believe we pass `null`

src/features/HelpAndKnowledge/SolutionsHelp/SolutionDetails/Solutions/Generic/_components/AddContractorForm/index.tsx
```diff
@@ -0,0 +1,188 @@
...
+        setHasMutationError(true);
+      });
+  };
+
+  return (
+    <FormProvider {...methods}>
+      <Form
+        className="maxw-none"
+        data-testid="mailbox-form"
+        id="mailbox-form"
```
patrickseguraoddball, line 95: This seems to be copied code and should reflect contractors

src/features/HelpAndKnowledge/SolutionsHelp/SolutionDetails/Solutions/Generic/_components/AddContractorForm/index.tsx
```diff
@@ -0,0 +1,188 @@
...
+          <Alert
+            type="error"
+            slim
+            headingLevel="h1"
+            className="margin-bottom-2"
+          >
+            {miscT('addContractor.error')}
+          </Alert>
+        )}
+        <Fieldset disabled={!selectedSolution}>
```
patrickseguraoddball, line 108: I saw this on the other forms and not sure how I feel about how we should implement this.  Above on line 57 you are returning null for the entire component if `selectedSolution` is falsy.  This disabled check will never actually hit. I think this is fine but just wanted to point this out in case we wanted to sync on a way to standardize/handle if an incorrect/not found solution is mounted

src/features/HelpAndKnowledge/SolutionsHelp/SolutionDetails/Solutions/Generic/_components/AddContractorForm/index.tsx
```diff
@@ -0,0 +1,188 @@
...
+          <Controller
+            name="contractorTitle"
+            control={control}
+            rules={{
+              required: false,
+              validate: value => value !== 'default'
```
patrickseguraoddball, line 114: This doesn't need any validation I believe

src/features/HelpAndKnowledge/SolutionsHelp/SolutionDetails/Solutions/Generic/_components/AddContractorForm/index.tsx
```diff
@@ -0,0 +1,188 @@
...
+              </FormGroup>
+            )}
+          />
+
+          <Controller
+            name="contractorName"
+            control={control}
+            rules={{
+              required: true,
+              validate: value => value !== 'default'
```
patrickseguraoddball, line 140: I think the validate rule should be empty string instead of `default`

patrickseguraoddball (changes requested):
Nice Elle :)

The alerts are also not reappearing when clicking to close them

The name of the contractor should be in bold
Screenshot 2025-07-01 at 7 23 01 AM

I was conflicted with the previous implementation as well as contractor with the decision to replicate the entire edit form, in favor of using the same form for add and edit.  I think 95% of the form is the same, just the mutation and a little bit of text is different.  I think we might want to just reuse the same component?  Open for discussion on this.
- All of the notes/requested changes on the `AddContractoForm` apply to the `EditContractotForm`

src/features/HelpAndKnowledge/SolutionsHelp/SolutionDetails/Solutions/Generic/_components/AddContractorForm/index.tsx
yiranli624: nice catch! I will update that!

src/features/HelpAndKnowledge/SolutionsHelp/SolutionDetails/Solutions/Generic/_components/AddContractorForm/index.tsx
yiranli624: Yeah I feel the same! I tested with if we hit url with none existing solution, it got redirected. So it's not the best look, but since it's not really being hit at all, I thought it's okay to leave it like this, until future changes. 
Do you have other ideas? im all ears!

PR #1796, opened 2025-07-03
[MINT-3425] Add error handling for duplication

yiranli624:
# MINT-3425

## Description

Add error handling(message display) when adding a poc that already exist.

## How to test this change

1.  Create a duplicated poc, same EUA user selected/same mailbox address/same contractor name, it should not be allowed and an info should be displayed accordingly.

Screenshot 2025-07-03 at 1 06 49 PM

## PR Author Checklist

## PR Reviewer Guidelines

### DIFF
--- a/src/features/HelpAndKnowledge/SolutionsHelp/SolutionDetails/Solutions/Generic/_components/ContractorForm/index.tsx
+++ b/src/features/HelpAndKnowledge/SolutionsHelp/SolutionDetails/Solutions/Generic/_components/ContractorForm/index.tsx
@@ -65,7 +65,9 @@ const ContractorForm = ({
   const { showMessage } = useMessage();
   const [create] = useCreateMtoCommonSolutionContractorMutation();
   const [update] = useUpdateMtoCommonSolutionContractorMutation();
-  const [hasMutationError, setHasMutationError] = useState(false);
+  const [mutationError, setMutationError] = useState<
+    'duplicate' | 'generic' | null
+  >(null);
 
   if (!selectedSolution) {
     return null;
@@ -125,8 +127,9 @@ const ContractorForm = ({
           closeModal();
         }
       })
-      .catch(() => {
-        setHasMutationError(true);
+      .catch(error => {
+        const duplicateError = error.message.includes('duplicate');
+        setMutationError(duplicateError ? 'duplicate' : 'generic');
       });
   };
 
@@ -138,14 +141,26 @@ const ContractorForm = ({
         id="contractor-form"
         onSubmit={handleSubmit(onSubmit)}
       >
-        {hasMutationError && (
+        {mutationError !== null && (
           <Alert
             type="error"
             slim
             headingLevel="h1"
             className="margin-bottom-2"
           >
-            {miscT(`${mode}.error`)}
+            {mutationError === 'generic' ? (
+              miscT(`${mode}.error`)
+            ) : (
+              <Trans
+                i18nKey="mtoCommonSolutionContractorMisc:duplicateError"
+                values={{
+                  contractor: methods.getValues('contractorName')
+                }}
+                components={{
+                  bold: <span className="text-bold" />
+                }}
+              />
+            )}
           </Alert>
         )}
         <Fieldset disabled={!selectedSolution}>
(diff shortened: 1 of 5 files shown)


src/features/HelpAndKnowledge/SolutionsHelp/SolutionDetails/Solutions/Generic/_components/ContractorForm/index.tsx
```diff
@@ -24,7 +24,7 @@ import dirtyInput from 'utils/formUtil';
 import { ModeType } from '../ContractorModal';
 
 type FormValues = {
-  contractorTitle: string | null;
+  contractorTitle: string;
   contractorName: string;
 };
```
patrickseguraoddball, line 29: Was there a specific reason to remove `null`? 

After looking more closely we actually should correlate this type directly to the query.  This is more true to the data and will prevent future bugs

```
type FormValues = {
  contractorTitle: SolutionContractorType['contractorTitle'];
  contractorName: SolutionContractorType['contractorName'];
};
```

src/features/HelpAndKnowledge/SolutionsHelp/SolutionDetails/Solutions/Generic/_components/ContractorForm/index.tsx
```diff
@@ -96,9 +98,7 @@ const ContractorForm = ({
               id: contractor.id,
               changes: {
                 contractorTitle:
-                  contractorTitle === undefined
-                    ? undefined
-                    : contractorTitle || null,
+                  contractorTitle === '' ? null : contractorTitle,
```
patrickseguraoddball, line 101: Nice for adding the `null` here.  We recently did some work to build this directly into `ditryInput` to not have to worry about this in the future.  It isn't in main yet, but also feel free to at it to that util function if you want


`src/utils/formUtil.ts`
```
export const dirtyInput = (initialValues: any, values: any) => {
  if (!initialValues || !values) return {};

  const onlyDirtyInput: DirtyInputType = {};

  Object.keys(initialValues).forEach(field => {
    // Added conditional for 0 value, as number inputs are always initialized with 0, and it needs to be persisted even if not touched
    if (
      typeof initialValues[field] === 'object' &&
      !isEqual(initialValues[field], values[field])
    ) {
      onlyDirtyInput[field] = values[field];
    } else if (
      !isEqual(initialValues[field], values[field]) ||
      initialValues[field] === 0
    ) {
      onlyDirtyInput[field] = values[field] === '' ? null : values[field];
    }
  });
  return onlyDirtyInput;
};
```

src/features/HelpAndKnowledge/SolutionsHelp/SolutionDetails/Solutions/Generic/_components/TeamMemberForm/index.tsx
```diff
@@ -30,6 +30,7 @@ import { TeamMemberModeType } from '../MailboxAndTeamMemberModal';
 
 type FormValues = {
   userName: string;
+  displayName: string;
   role: string;
   isPrimary: boolean;
   receiveEmails: boolean;
```
patrickseguraoddball, line 36: Again sorry I overlooked this earlier, but I think we want to tie to form values to the generated query type, rather than redefine the type

src/features/HelpAndKnowledge/SolutionsHelp/SolutionDetails/Solutions/Generic/_components/ContractorForm/index.tsx
```diff
@@ -125,8 +125,9 @@ const ContractorForm = ({
           closeModal();
         }
       })
-      .catch(() => {
-        setHasMutationError(true);
+      .catch(error => {
+        const duplicateError = error.message.includes('duplicate');
```
patrickseguraoddball, line 129: My only suggestion (definitely dont need to take it as this works) but a more accurate and specific identifier to key on would be `uniq_user_id_per_solution_key`, which is present in the gql error message

patrickseguraoddball (changes requested):
Nice Elle!  This works great just a few suggestions, requested edits.  Open to sync if you need anything!

src/features/HelpAndKnowledge/SolutionsHelp/SolutionDetails/Solutions/Generic/_components/ContractorForm/index.tsx
yiranli624: I was gonna say the null come from query lol. Perfect i will update it!

src/features/HelpAndKnowledge/SolutionsHelp/SolutionDetails/Solutions/Generic/_components/ContractorForm/index.tsx
yiranli624: I like that better! Ty!

patrickseguraoddball (approved):
Thanks!

PR #1805, opened 2025-07-09
[MINT-3430] Help Center solution map refactor

patrickseguraoddball:
# MINT-3430

## Description

The Help and Knowledge Center was first created before the concept of solution existed on the BE (no enums, gql types, etc).  Once solutions were introduced in our tables and schema, this expanded the scope of our mapping, and created extra complexity to format into our existing model.  This tech debt work adjust FE solution mapping to use generated GQL types across the board for all of our solution implementations.  This DOES leave in tact legacy routing for sent emails that may still be linked.  All of the work in the PR should have no visual or functional changes except for our default routing will now be in the format `solution-key="SOLUTION_ENUM"` rather than `solution="solution-route"`

- Removed unneeded mapping fields
- Reworked the mapping to be an object rather than an array
  - Add a until to convert the object to an array where it made sense for data manipulation
- Changed all keys to key off the generated enum values
- Tweak some business utils/no change in functionality
- Added small util to still use legacy routes to adhere to GA4/ send titles
- Updated unit test

## How to test this change

1.  Log into MINT and navigate all through the HaKC
2. Click on emails from the `scripts/dev email` command that link to solutions
3. Navigate all through MTO
4. Verify all functionality remains

## PR Author Checklist

## PR Reviewer Guidelines

### DIFF
--- a/src/features/HelpAndKnowledge/SolutionsHelp/SolutionDetails/Solutions/Generic/About/__snapshots__/index.test.tsx.snap
+++ b/src/features/HelpAndKnowledge/SolutionsHelp/SolutionDetails/Solutions/Generic/About/__snapshots__/index.test.tsx.snap
@@ -168,19 +168,16 @@ exports[`Generic About Components > matches the snapshot > 4innovation 1`] = `
         ?
       </h4>
       <span>
-        <button
-          class="usa-button usa-button--unstyled"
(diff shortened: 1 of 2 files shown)


yiranli624 (approved):
lgtm, great work!

PR #1829, opened 2025-07-24
[MINT-3435] Reenable Discussion Tag

patrickseguraoddball:
# MINT-3435
# MINT-3434
# MINT-3433

## Description

This PR covers the above listed tickets.  Decided it was easier to include everthing in a single branch, as MINT-3434 was not needed and FE changes from MINT-3433 was negligent

After discussion with Steven, the approach was to not migrate legacy data, rather just add a new tag type for MTO common solutions.  With this approach I added a UUID to the `mto_common_solution` table, now using the same `EntityUUID` as the `user_accout`

- Added migration new tag type and new UUID (id) column on `mto_common_solution` table
- Converted email func from possible solution to mto solutions for `sendPlanDiscussionTagEmails` - we will no longer need them moving forward.  There are other legacy util left in place, however - as those are still needed to parse legacy possible solution data in the future
- Fixes type handling for MTO Common Solution tags to ensure correct UUID assignment and constraint satisfaction.
- Updates the resolver logic to use the correct loader and argument types for fetching MTO Common Solution entities
- Refactors the processMTOCommonSolutionHTMLMention function to use the solution key loader and properly assign UUIDs.
- Reenabled FE solution tag - adjusted for new tag type

## How to test this change

1.  Add a PoC to one of the MTO common solutions in the help center
2. Tag a discussion or reply with that MTO Common Solution in the UI.
3. Verify that the tag is created without errors.
4. Confirm that tagged MTO Common Solutions are rendered in the subsequent query
5. Verify email gets properly generated

## PR Author Checklist

## PR Reviewer Guidelines


migrations/V217__Add MTO_Common_Solution_Tag_Type.sql
```diff
@@ -0,0 +1,22 @@
+-- Add new tag type for MTO common solutions
+ALTER TYPE tag_type ADD VALUE 'MTO_COMMON_SOLUTION';
+
+/* 
+  Add the new UUID/id column. This is used to allow MTO common solutions to use the entity_uuid on tagged content, same as user_account.
+  Previously operational solutions were using int as id, moving to UUID allows for better compatibility with other tables 
+*/
+ALTER TABLE mto_common_solution
+ADD COLUMN id UUID;
+
```
StevenWadeOddball, line 10: thanks Patrick, would you mind changing this to a sql column comment? That way it will persist beyond a flyway migration
```suggestion
COMMENT ON COLUMN mto_common_solution.id IS 'This is used to allow MTO common solutions to use the entity_uuid on tagged content, same as user_account.  Previously operational solutions were using int as id, moving to UUID allows for better compatibility with other tables ';
```

pkg/storage/loaders/mto_common_solution.go
```diff
@@ -23,13 +23,16 @@ type mtoCommonSolutionLoaders struct {
 
 	// ByCommonMilestoneKey returns a list of mto Common Solution records by it's keys. It does not currently have contextual data
 	ByCommonMilestoneKey LoaderWrapper[models.MTOCommonMilestoneKey, []*models.MTOCommonSolution]
+
+	ByID LoaderWrapper[uuid.UUID, *models.MTOCommonSolution]
```
StevenWadeOddball, line 27: Can you add a comment to this function please?

StevenWadeOddball (commented):
Thanks @patrickseguraoddball , this looks good! Could you please update the bruno collection? ( I think we are still trying to tag possible solutions instead of MTO ones there).

PR #1834, opened 2025-07-29
[MINT-3444] Read-view GA4 route updates

patrickseguraoddball:
…, unit test

# MINT-3444

## Description

- Updated any manual routes that were pointing to `read-only` to `read-view`
  - Left legacy redirects to account to sent emails
- Added `excludedPaths` to exclude from GA4 pagesends
- Added unit test to regex matching of excluded paths

## How to test this change

Can either: 
1. Log out excluded path variable here - https://github.com/CMS-Enterprise/mint-app/blob/06d2a164465157cbeeae7176802d2026238145f4/src/hooks/useRouteTitle.ts#L86-L87
2. Navigate to any of the direct match routes listed in the `excludedPaths` const
3. Verify `true` is logged first, indicating a redirect and a triggered conditional on the GA4 send

Or:
1. Log into GA4 and navigate to any of the `excludedPaths` and verify they do not come through the data stream.  Note this is/may not be instant.

## PR Author Checklist

## PR Reviewer Guidelines

### DIFF
--- a/pkg/email/templates/discussion_reply_created_originator_body.html
+++ b/pkg/email/templates/discussion_reply_created_originator_body.html
@@ -31,7 +31,7 @@
         {{end}}
 
         <p style="font-weight: normal">
-          <a href="{{.ClientAddress}}/models/{{.ModelID}}/read-only/discussions?discussionID={{.DiscussionID}}">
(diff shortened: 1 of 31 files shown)


Peyton232 (approved):
validated

PR #1839, opened 2025-07-31
[NOREF] Address routing issue on useHandleMutation hook

patrickseguraoddball:
# NOREF

## Description

I noticed that when a mutation error occurs on a form, and the modal appears to confirm leave - if the user chose to stay, it was still rerouting.  

- Added a method to clear the destination value when a user chooses to stay on the page from the modal

## How to test this change

1.  Navigate to form (Timeline for example)
2. Edit this value to ensure a mutation error:
    https://github.com/CMS-Enterprise/mint-app/blob/e885c940ddcda4a3d869abb49e2a223041ea2b02/src/features/ModelPlan/Timeline/index.tsx#L87
To `id: '123'`
3. Submit the form or try to navigate away and verify the modal appears
4. Choose to stay on the page and verify
5. Verify leaving the page as well

## PR Author Checklist

## PR Reviewer Guidelines

### DIFF
--- a/src/features/ModelPlan/DataExchangeApproach/CollectingAndSendingData/index.tsx
+++ b/src/features/ModelPlan/DataExchangeApproach/CollectingAndSendingData/index.tsx
@@ -125,7 +125,7 @@ const CollectingAndSendingData = () => {
     <>
       <MutationErrorModal
         isOpen={mutationError.isModalOpen}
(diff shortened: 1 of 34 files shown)


garyjzhao (approved):
LGTM. Went through the described instructions and it is working as expected

PR #1876, opened 2025-08-25
[MINT-2995] In-app notifications for model status

yiranli624:
# MINT-2995

## How to test this change

1.  

## PR Author Checklist

## PR Reviewer Guidelines


Peyton232:
I'm not seeing a notification in notification center if I have the notifications for in-app set to off when I'm a model lead, when I turn the notifications on I see them however. 


Screenshot 2025-08-26 at 11 55 44 AM
I'm also seeing this phantom notification after 
1. create model plan 1
2. update model plan so not archived 
3. update timeline so end date in past
4. change notification setting sin app to on
5. create model plan 2
6. update model plan so not archived 
7. update timeline so end date in past
8. then I see the empty notification in the notification center 
Screenshot 2025-08-26 at 11 57 29 AM

It also just keeps populating with these empty notifications every couple minutes, I notice there is more and more activities in the DB so I believe that is probably a BE logic error, especially since we are not getting multiple emails on this. I'll take a look into why we are getting repeat activities in the DB but they do have data in the DB but displaying empty in the notification center which is concerning

Peyton232 (changes requested):
comment above

PR #1878, opened 2025-08-26
Mint 3452/new discussion notification

Peyton232:
# [MINT-3452](https://jiraent.cms.gov/browse/MINT-3452)

## Description

Added backend activity for new discussion added as well as user notification preferences for the new notification. Made an update to the email for discussion added, and tied in routing logic for generating activities for notifications. 

## How to test this change

1.  run test emails and check the new discussion started email and compare to [figma](https://www.figma.com/design/8MfXzXzmKWv8vBjTasUIkD/MINT---Notifications?node-id=5980-15355&p=f&t=pHHhqS7LvMyGKIj1-0) 
2. create a model plan (query_examples/MINT/Model Plan/Create Model Plan.bru)
3. add a plan discussion to model plan (query_examples/MINT/Discussions & Replies/Create Plan Discussion.bru)
4. check that NEW_DISCUSSION_ADDED is in activity database table

## PR Author Checklist

## PR Reviewer Guidelines


pkg/graph/resolvers/plan_discussion.go
```diff
@@ -77,31 +77,76 @@ func CreatePlanDiscussion(
 
 		userRole := discussion.UserRole.Humanize(models.ValueOrEmpty(discussion.UserRoleDescription))
 
-		// Send email to MINT Dev Team
+		// Send email to MINT Dev Team and those subscribed to discussion created emails
 		go func() {
-			sendEmailErr := sendPlanDiscussionCreatedEmail(
+			emailRecipients, sendEmailErr := GetEmailRecipients(
```
StevenWadeOddball, line 82: should this err be named differently?

pkg/graph/resolvers/plan_discussion.go
```diff
@@ -317,13 +362,87 @@ func sendPlanDiscussionTaggedSolutionEmail(
 	return nil
 }
 
+func GetEmailRecipients(
```
StevenWadeOddball, line 365: Since this is scoped to the entire resolver package, I think that it should probably have a more specific name. Can you please also add a function description it to describe the use of the function?

pkg/graph/resolvers/plan_discussion.go
```diff
@@ -317,13 +362,87 @@ func sendPlanDiscussionTaggedSolutionEmail(
...
+	return emailRecipients, nil
+}
+
+func GetNotificationRecipients(
+	ctx context.Context,
+	logger *zap.Logger,
+	store *storage.Store,
+	modelPlanID uuid.UUID,
+) ([]uuid.UUID, error) {
+	planCollaborators, err := store.PlanCollaboratorGetByModelPlanID(logger, modelPlanID)
```
StevenWadeOddball, line 408: There is a data loader for these
PlanCollaboratorGetByModelPlanIDLOADER

pkg/graph/resolvers/plan_discussion.go
```diff
@@ -317,13 +362,87 @@ func sendPlanDiscussionTaggedSolutionEmail(
 	return nil
 }
 
+func GetEmailRecipients(
+	ctx context.Context,
+	logger *zap.Logger,
+	store *storage.Store,
+	modelPlanID uuid.UUID,
+) ([]string, error) {
+	planCollaborators, err := store.PlanCollaboratorGetByModelPlanID(logger, modelPlanID)
```
StevenWadeOddball, line 371: I might suggest you look at how users are returned for dates changed. This logic here results in a number of database calls. Even though these are technically dataloaders, I believe they are being called synchronously. I think a bespoke call might serve us better here.

pkg/graph/resolvers/plan_discussion.go
```diff
@@ -77,31 +77,76 @@ func CreatePlanDiscussion(
 
 		userRole := discussion.UserRole.Humanize(models.ValueOrEmpty(discussion.UserRoleDescription))
 
-		// Send email to MINT Dev Team
+		// Send email to MINT Dev Team and those subscribed to discussion created emails
 		go func() {
```
StevenWadeOddball, line 81: I noted elsewhere, but I think there should probably be some refactoring around fetching preferences so that it doesn't go to the db as much.

pkg/notifications/new_discussion_added_meta.go
```diff
@@ -0,0 +1,45 @@
...
+	"context"
+
+	"github.com/google/uuid"
+
+	"github.com/cms-enterprise/mint-app/pkg/models"
+	"github.com/cms-enterprise/mint-app/pkg/sqlutils"
+)
+
+// ActivityNewDiscussionAddedCreate creates an activity for when a Discussion is added.
+func ActivityNewDiscussionAddedCreate(
```
StevenWadeOddball, line 13: I think looking at the dates changed activity could be good here. I think also makes sense to potentially move the email logic to happen here instead of in resolvers.

StevenWadeOddball (changes requested):
Thanks Petyon! This is looking good. I have some suggestions around refactoring for performance.  Mind taking a look?

StevenWadeOddball:
@Peyton232 can you also make this target a feature branch since there is going to be FE changes for this as well?

pkg/email/templates/plan_discussion_created_body.html
```diff
@@ -30,6 +30,11 @@ <h2>Discussion</h2>
 <hr/>
 <br/>
 
-{{template "shared_mint_mailbox_footer.html"}}
+<p>
```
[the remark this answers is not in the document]
Peyton232: Yeah I was thinking about this as well, definitely could be done, just wasn't sure what best way to pass it in. It would be the human readable term for the footer such as "new discussion added" and the text that links to the notification page, which is based on activity type enum. We could probably do that by having an const for readable name in the models for templating themselves and then passing in the enum type string for what page to be directed to for unsubscribe. I couldn't come up with a really clean way to do this but it's good food for thought

pkg/email/plan_discussion_created.go
```diff
@@ -19,3 +23,22 @@ type PlanDiscussionCreatedBodyContent struct {
 	ModelName         string
 	Role              string
 }
+
+// constructor for PlanDiscussionCreatedBodyContent
+func NewPlanDiscussionCreatedBodyContent(
+	clientAddress string,
+	planDiscussion *models.PlanDiscussion,
```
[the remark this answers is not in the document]
StevenWadeOddball: Yeah, I think that is fine. And we are already importing models into this package.

pkg/email/templates/plan_discussion_created_body.html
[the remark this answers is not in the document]
StevenWadeOddball: Thanks Peyton! Yeah, maybe we can keep this on the back burner and note to refactor. I know for some of them, Zoë generalized the content, so they could just use the shared footer as well 🤷. Maybe we have an embedded struct for footer stuff in all these email content structs, so the fields can be shared when you use the footer?

pkg/graph/schema/types/activity.graphql
```diff
@@ -6,11 +6,11 @@ enum ActivityType {
   ADDED_AS_COLLABORATOR
   TAGGED_IN_DISCUSSION
   TAGGED_IN_DISCUSSION_REPLY
-  NEW_DISCUSSION_REPLY
```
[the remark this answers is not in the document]
Peyton232: nope, added back in now, good catch

yiranli624 (changes requested):
The trigger email was sent to mint team instead of user, is that intended?

Few ui things:
- title should be 'New discussion started by xxx `in` xxxl', currently it says `on`
- subheader should be `Model Innovation Tool`, without `The`
- footer's `unsubscribe` and `notification setting` are unbolded, maybe double check with ui, or update.

Peyton232:
Checked with Zoë on the sub header and changed the shared template to reflect what is in the figma. Also trigger email being sent to mint team is fine, this email gets sent to mint team and any who are following the notification

yiranli624 (approved):
pass functional test! good work!

PR #1894, opened 2025-09-05
[MINT-3487] Updated session variable for suggested phase

patrickseguraoddball:
# MINT-3487

## Description

We recently noticed an edge case where it's possible to have a suggested phase/modal trigger, and the session variable would be stored for that model plan to persist that the user has already view this suggestion.  However if a new/different suggestion was triggered, the modal would not render, as it was not checking for a different suggested phase.  This PR now stores the value of the suggested phase and renders the modal if it is a new phase or if there is no session variable stored

## How to test this change

1.  Create a new model plan (seeded data wont work)
2. Add a clearance date in the past
3. Go back to collab area
4. Suggested phase modal appears, close it
5. Navigate away and return.  Verify it does not reappear
6. Go back to timline and add the performance end date to also be in the past. 
7. Go back to collab area and verify the modal reappears with the new suggested phase

## PR Author Checklist

## PR Reviewer Guidelines

### DIFF
--- a/src/features/ModelPlan/CollaborationArea/index.tsx
+++ b/src/features/ModelPlan/CollaborationArea/index.tsx
@@ -77,7 +77,8 @@ const CollaborationArea = () => {
 
   // Gets the sessions storage variable for statusChecked of modelPlan
   const statusCheckedStorage =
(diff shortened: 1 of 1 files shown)


yiranli624 (approved):
lgtm!

PR #1903, opened 2025-09-08
[MINT-3451] Add new discussion in notification and settings

yiranli624:
# MINT-3451

## Description
 
- Add new discussion in notification and settings, user should be able to save pref in settings, and see new discussion notification accordingly.
- update snapshots

## How to test this change

1.  Functional: update notification settings to include in app, then create discussion in model plan, review notification
2. verify all ui matches [figma](https://www.figma.com/design/8MfXzXzmKWv8vBjTasUIkD/MINT---Notifications?node-id=5980-15355&p=f&t=jhRo5U3f6lu5ag1g-0)
3. verify unsubscription works accordingly, could get link from mailcatcher

## PR Author Checklist

## PR Reviewer Guidelines


patrickseguraoddball (changes requested):
Looks good Elle.  I think the only thing missing is a cypress test for this new notification.  I realized we may have also missed the last notification.  If you could add some coverage for this, much appreciated

patrickseguraoddball (changes requested):
I noticed that the email unsubscribe link does not work for newly created models, only seeded models.  None of the links on the notification email seem to point to the correct model plan

Also I noticed that there's a mismatch between enums in the email param, and the actual enum value, causing the unsubscribe to fail

https://github.com/CMS-Enterprise/mint-app/blob/433e0867eeba51132c06ece9c0b83f63d47ea5ff/pkg/email/templates/model_plan_suggested_phase_body.html#L36

https://github.com/CMS-Enterprise/mint-app/blob/433e0867eeba51132c06ece9c0b83f63d47ea5ff/src/gql/generated/graphql.ts#L67-L79

`INCORRECT_MODEL_STATUS`  - `"INCORRECT_MODEL_STATUS"_ACTIVITY` mismatch

patrickseguraoddball (approved):
LG

PR #1904, opened 2025-09-08
[NOREF] Simplify parse meta data logic and move to models

StevenWadeOddball:
## Description

@Peyton232 looked at how we are currently parsing meta data, and found it confusing to note where we have to match an enum to a type. We also determined that we could genericize the logic and move to a map. This makes the location of meta data and the enum in one close place, and only will require a new mapping when a new activity type is created.

## How to test this change

1.  Log into the app
2. Make new in app notifications
3.  Make sure to make multiples of the same type to validate that it shows a unique activity for each.

## PR Author Checklist

## PR Reviewer Guidelines


pkg/models/activity.go
```diff
@@ -42,3 +60,34 @@ func NewActivity(actorID uuid.UUID, entityID uuid.UUID, activityType ActivityTyp
 		ActivityType: activityType,
 	}
 }
+
+// ParseRawActivityMetaData conditionally parses the raw meta data JSON into the appropriate struct based on ActivityType
+// It returns a boolean indicating if parsing was successful and an error if any occurred during parsing
+func (a *Activity) ParseRawActivityMetaData() (bool, error) {
```
Peyton232, line 66: I'm not quite seeing why we need a bool and error return type here, anytime there is an error we return false, and vice versa, so they seem coupled together so that your only returns are 

(false, err) or (true, nil)

Peyton232 (changes requested):
Looks really good, passes functional test, also tested duplicates of notifications so all looks good. Only concern is why we are returning a success bool and err in the ParseRawActivityMetaData function, I may just be missing something though

pkg/models/activity.go
StevenWadeOddball: Yeah, I think that's fair! I'll work to adjust it

Peyton232:
New discussion is being merged in right now, let's wait on this until that is shipped cause will need to add that to this map also

StevenWadeOddball:
Thanks @Peyton232 ! For anyone taking a look, Peyton is going to wrap this up and merge once the other discussion notification is merged

PR #1926, opened 2025-09-23
[MINT-3496] Notification refactor

yiranli624:
# MINT-3469

## Description

<NotificationSettings />  and <IndividualNotification /> components could use some love for easier and better maintenance. Currently when adding a new notification, dev would take some time to relearn the unfriendly html structure, then manually add the new notification at a dozen scattered places. 
This pr 
- Restructure html for <NotificationSettings /> components
- Extract 'which notification' logic to reduce the long and chunky functions, so future adding new notifications should be almost effortless.
- Migrate from formik to r-h-f.
- added util functions and unit tests.

into a bit confusion due to how the html is mapped out, and functions are takes some 
. So it could be rework to be more logical and intuitive to read.
Additionally, the logic for each notification (unsubscribe and other spots) have too many ifs, it's chunky and not easy to maintain.
rework html.
extract 'which notification' logic to reduce the long and chunky functions.

## How to test this change

1.  trigger all kinds of notification, and verify it still works and display the same.

## PR Author Checklist

## PR Reviewer Guidelines


patrickseguraoddball (changes requested):
Nice work Elle, thanks for taking this on.

https://github.com/CMS-Enterprise/mint-app/blob/abaf5b900a5404b468f21971c3cf6a3787621ee2/src/features/Notifications/Settings/index.tsx#L310-L312

I think this was always here but the component should be `<NotFound />` rather than `<NotFoundPartial />` as is it not wrapped in the `<MainContent />` component

Also something that has always been there and if you have bandwidth - when expanding a really long notification it scrolls to the bottom.  Would it be possible to have it work where it does not behave this way?  Not a blocker just a pet peeve that was never fixed.

This highLevelTimeline is not translated
Screenshot 2025-09-24 at 9 23 05 AM

The wrong email is being unsubscribed for when dates changes. `When a new Model Plan is created` is being unsubscribed instead.  The link in the email seems to be incorrect - `<a href="http://localhost:3005/notifications/settings?unsubscribe_email=NEW_MODEL_PLAN" style="font-weight: normal" target="_blank">unsubscribe</a>`
Screenshot 2025-09-24 at 9 26 52 AM
 
Screenshot 2025-09-24 at 9 26 44 AM

The link to unsubscribe from the data exchange approach complete is not working.  Here the link seems to be correct, and the mutation appears successful, but the checkbox is unaffected

yiranli624:
> Nice work Elle, thanks for taking this on.
> 
> https://github.com/CMS-Enterprise/mint-app/blob/abaf5b900a5404b468f21971c3cf6a3787621ee2/src/features/Notifications/Settings/index.tsx#L310-L312
> 
> I think this was always here but the component should be `<NotFound />` rather than `<NotFoundPartial />` as is it not wrapped in the `<MainContent />` component
> 
> Also something that has always been there and if you have bandwidth - when expanding a really long notification it scrolls to the bottom. Would it be possible to have it work where it does not behave this way? Not a blocker just a pet peeve that was never fixed.
> 
> This highLevelTimeline is not translated Screenshot 2025-09-24 at 9 23 05 AM
> 
> The wrong email is being unsubscribed for when dates changes. `When a new Model Plan is created` is being unsubscribed instead. The link in the email seems to be incorrect - `<a href="http://localhost:3005/notifications/settings?unsubscribe_email=NEW_MODEL_PLAN" style="font-weight: normal" target="_blank">unsubscribe</a>` Screenshot 2025-09-24 at 9 26 52 AM
> 
> Screenshot 2025-09-24 at 9 26 44 AM
> The link to unsubscribe from the data exchange approach complete is not working. Here the link seems to be correct, and the mutation appears successful, but the checkbox is unaffected

Ty for the detailed review, noted on the comments!  About the last point, unsubscribe is actually working, it's just bit slow, if you wait until the toasty alert disappear, then the checkbox will be updated. It seems working like this from before.

patrickseguraoddball (approved):
LG thanks!

PR #1935, opened 2025-09-25
MINT-3497: added migration, schema, and resolver logic for assigned_to field

Peyton232:
# [MINT-3497](https://jiraent.cms.gov/browse/MINT-3497)

## Description

This PR adds the ability to assign MTO milestones to specific users by implementing the assignedTo field and adding functionality to resolvers.

## How to test this change

1. create model plan createModelPlan
2. create a new milestne NewMTOMilestone
3. run currentUser to get user id or planCollaboratorByID to see what options can be added
4. add solution to mto NewMTOSolution
5. add category to MTO NewMTOCategory
6. update MTO with asisgnedTo udpateMTOMilestone
7. run get MTO to check assignedTo field
8. check assignedTo in DB

## PR Author Checklist

## PR Reviewer Guidelines

### DIFF
--- a/mappings/translation/mto_milestone.json
+++ b/mappings/translation/mto_milestone.json
@@ -67,6 +67,18 @@
     "formType": "TEXT",
     "order": 1.035
   },
+  "assignedTo": {
+    "gqlField": "assignedTo",
+    "goField": "AssignedTo",
+    "dbField": "assigned_to",
+    "label": "Assigned to",
+    "exportLabel": "Milestone assigned to",
+    "sublabel": "Select the person responsible for completing this milestone.",
+    "dataType": "UUID",
+    "formType": "SELECT",
+    "order": 1.037,
+    "tableReference": "user_account"
+  },
   "needBy": {
     "gqlField": "needBy",
     "goField": "NeedBy",
--- a/pkg/graph/model/models_gen.go
+++ b/pkg/graph/model/models_gen.go
@@ -126,6 +126,7 @@ type MTOMilestoneTranslation struct {
 	Status             models.TranslationFieldWithOptions `json:"status" db:"status"`
 	Key                models.TranslationField            `json:"key" db:"mto_common_milestone_key"`
 	MtoCategoryID      models.TranslationField            `json:"mtoCategoryID" db:"mto_category_id"`
+	AssignedTo         models.TranslationField            `json:"assignedTo" db:"assigned_to"`
 }
 
 type MTOSolutionLinks struct {
--- a/pkg/graph/schema/types/mto_milestone_translation.graphql
+++ b/pkg/graph/schema/types/mto_milestone_translation.graphql
@@ -17,4 +17,5 @@ type MTOMilestoneTranslation {
   status: TranslationFieldWithOptions! @goTag(key: "db", value: "status")
   key: TranslationField! @goTag(key: "db", value: "mto_common_milestone_key")
   mtoCategoryID: TranslationField! @goTag(key: "db", value: "mto_category_id")
+  assignedTo: TranslationField! @goTag(key: "db", value: "assigned_to")
 }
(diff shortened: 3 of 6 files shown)


pkg/graph/schema/types/mto_milestone.graphql
```diff
@@ -24,6 +24,7 @@ type MTOMilestone {
   Note, this data will be cleared if other is removed as an option
   """
   facilitatedByOther: String
+  assignedTo: UserAccount
```
StevenWadeOddball, line 27: I think we should split this, we typically allow access to the raw data. Since assignedTo is the name of the DB field, I'd suggest we have `assignedTo` as well as `assignedToUserAccount`. So if someone just wants the uuid, they can get it without having to invoke the loader

pkg/graph/resolvers/mto_milestone_test.go
```diff
@@ -581,3 +581,31 @@ func (suite *ResolverSuite) TestMTOMilestoneNoLinkedSolutions_MultiplePlans() {
...
+func (suite *ResolverSuite) TestMTOMilestoneAssignedTo() {
+	// Create model plan
+	plan := suite.createModelPlan("testing milestone assigned_to")
+	assignedTo := uuid.MustParse("00000001-0001-0001-0001-000000000001") // Mint user ID
+
+	// Create a custom milestone with an assigned_to
+	milestone, err := MTOMilestoneCreateCustom(suite.testConfigs.Context, suite.testConfigs.Logger, suite.testConfigs.Principal, suite.testConfigs.Store, "Test Milestone", plan.ID, nil)
+	suite.NoError(err)
+	suite.NotNil(milestone)
+
```
StevenWadeOddball, line 595: maybe worth asserting that assigned is Nil here?

pkg/graph/resolvers/mto_milestone_test.go
```diff
@@ -581,3 +581,31 @@ func (suite *ResolverSuite) TestMTOMilestoneNoLinkedSolutions_MultiplePlans() {
...
+	changes := map[string]interface{}{
+		"assignedTo": assignedTo,
+	}
+	updatedMilestone, err := MTOMilestoneUpdate(suite.testConfigs.Context, suite.testConfigs.Logger, suite.testConfigs.Principal, suite.testConfigs.Store, nil, nil, email.AddressBook{}, milestone.ID, changes, nil)
+	suite.NoError(err)
+	suite.NotNil(updatedMilestone)
+	suite.NotNil(updatedMilestone.AssignedTo)
```
StevenWadeOddball, line 602: if you want to avoid the panic, you could do
```suggestion
	if suite.NotNil(updatedMilestone.AssignedTo) {
	   suite.Equal(assignedTo, *updatedMilestone.AssignedTo)
	}
```

One can also override the suite to force this behavior (happens in EASi), but we don't do it in MINT currently.

pkg/storage/truncate.go
StevenWadeOddball, line 1: Are these changes needed?  Was something breaking? The change table is a loose coupling without fk constraints, so it shouldn't need to be removed first.

StevenWadeOddball (changes requested):
Thanks Peyton! This looks really great, just a few minor things. I have not tested functionally, but I don't anticipate any issues from reviewing the code! 🎉

pkg/storage/truncate.go
Peyton232: adding the FK reference to the assignedTo made the ordering wrong, i.e. needed to reorder specifically just MTO milestone. And yeah actually change table since it's loose coupling doesn't actually need to be removed first. Since I was re ordering one of these I decided to just group them in a more understandable way. I will go ahead and move audit change back since not necessary

StevenWadeOddball (commented):
Thanks Peyton! This looks good, I think it's ready to go, just the failing unit test (need to add a translation value)

PR #1939, opened 2025-09-30
[NOREF] added retry logic and delay to linking new solutions

Peyton232:
# NOREF

## Description

[TraceID Search](https://splunk.cloud.cms.gov/en-US/app/cms_mint_landing_app/search?display.page.search.mode=verbose&dispatch.sample_ratio=1&q=search%20source%3D/aws/ecs/prod/mint-app%20AND%20traceID%3D5b9fcd61-55ee-4330-a02e-99e5b143b03a&earliest=-7d%40h&latest=now)

## How to test this change

1. To help reproduce functionality put this block of code "time.Sleep(200 * time.Millisecond) // TODO remove after done testing" in of the MTOMilestoneUpdateLinkedSolutionsWithTX function, just before the return. This will help ensure that the transaction isn't done yet and trigger the retry logic   
2. create a model plan
3. NewMTOMilestone
4. NewMTOSolution
5. NewMTOCategory
6. udpateMTOMilestone but add a solution not already linked like commonSolutionKeys: [CDX, IDR, CONNECT, LOI, RFA],

## PR Author Checklist

## PR Reviewer Guidelines

### DIFF
--- a/pkg/graph/resolvers/mto_milestone.go
+++ b/pkg/graph/resolvers/mto_milestone.go
@@ -3,6 +3,8 @@ package resolvers
 import (
 	"context"
 	"fmt"
+	"strings"
+	"time"
 
 	"github.com/samber/lo"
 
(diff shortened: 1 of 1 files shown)


yiranli624 (dismissed):
Passed functional test! good work!

StevenWadeOddball (commented):
Thanks Peyton, I think there might be a simpler fix here. the solution selected email can actually take a named preparer, so you could remove the store reference from the parent function and just use the transaction. Maybe it makes sense to remove the go routine when we use a transaction?

What do you think?

StevenWadeOddball:
> > Thanks Peyton, I think there might be a simpler fix here. the solution selected email can actually take a named preparer, so you could remove the store reference from the parent function and just use the transaction. Maybe it makes sense to remove the go routine when we use a transaction?
> > What do you think?
> 
> That's a fair point, we could just make it part of the transaction as well, could run into the scenario of someone getting an email even though the transaction gets rolled back later on, I think that would rarely happen though so should be fine. If we wanted something tighter we could get utilize a done channel and when the transaction is done we send to that channel and unblock the goroutines, but that feels like a large departure from our current design.

Those are some good points 🤔. It feels almost like the email should be later on, after the transaction completes, but it is starting to feel a bit convoluted. At this point, I think the email could fire if the transaction is rolled back now too 🤔

StevenWadeOddball (approved):
Thanks @Peyton232! This should be good. I think you are right that firing a message when transaction succeeds would be great too, but maybe something to refactor later. Maybe as part of the transaction helpers when a transaction is committed 🤔

PR #1942, opened 2025-10-02
[MINT-3531] Add validation for needby field in edit milestone 

yiranli624:
# MINT-3531

## Description
 Users can manually enter non-conforming dates in the "Need by" field but the system doesn't throw an error and appears to save when it actually doesn’t. This pr add field validation for needby to prevent user from adding invalid date.

## How to test this change

1.  Edit the details for a milestone that has no "Need by" date
2. Enter "1/5/26" or some other non-conforming date format
3. Save changes
4. Verify field validation appear, and should not allow user to save. 

## PR Author Checklist

## PR Reviewer Guidelines


Peyton232 (approved):
works as expected

PR #1948, opened 2025-10-07
Mint 3511 assigned collaborator update

Peyton232:
# [MINT-3511](https://jiraent.cms.gov/browse/MINT-3511)

## Description

Whenever a plan collaborator is removed then any milestones they are assigned to are changed to null. 

## How to test this change

Test 1
1.  create a model plan 
2. create another model plan collaborator 
3. delete a collaborator 

Test 2
1. create a model plan 
2. create another model plan collaborator 
3. create a milestone
4. assign someone to that milestone
5. delete a collaborator 
6. verify the assignedTo is nulled out and no longer assigned to the collaborator you just deleted 

## PR Author Checklist

## PR Reviewer Guidelines

### DIFF
--- a/pkg/graph/resolvers/plan_collaborator.go
+++ b/pkg/graph/resolvers/plan_collaborator.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/google/uuid"
+	"github.com/jmoiron/sqlx"
 	"go.uber.org/zap"
 
 	"github.com/cms-enterprise/mint-app/pkg/email"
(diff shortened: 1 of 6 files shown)


pkg/graph/resolvers/plan_collaborator_test.go
```diff
@@ -321,7 +321,7 @@ func (suite *ResolverSuite) TestPlanCollaboratorDeleteLastModelLead() {
 	collaborator := collaborators[0]
 	deletedPlanCollaborator, err := PlanCollaboratorDelete(suite.testConfigs.Logger, collaborator.ID, suite.testConfigs.Principal, suite.testConfigs.Store)
 	suite.Error(err)
-	suite.EqualValues("pq: There must be at least one MODEL_LEAD assigned to each model plan", err.Error())
+	suite.Contains(err.Error(), "There must be at least one MODEL_LEAD assigned to each model plan")
```
StevenWadeOddball, line 324: This looks like a more appropriate method! 🔥

pkg/storage/plan_collaboratorStore.go
```diff
@@ -106,37 +108,33 @@ func (s *Store) PlanCollaboratorUpdate(
...
-		return nil, err
+	args := map[string]interface{}{
+		"id":          id,
+		"modified_by": userID,
 	}
-	defer stmt.Close()
 
-	collaborator := &models.PlanCollaborator{}
-	err = stmt.Get(collaborator, utilitysql.CreateIDQueryMap(id))
+	retCollaborator, err := sqlutils.SelectProcedure[models.PlanCollaborator](tx, sqlqueries.PlanCollaborator.Delete, args)
```
StevenWadeOddball, line 128: I think this is supposed to be a get procedure. It's a delete by id, so we expect only one value to be returned, right?

pkg/storage/plan_collaboratorStore.go
```diff
@@ -106,37 +108,33 @@ func (s *Store) PlanCollaboratorUpdate(
 
 // PlanCollaboratorDelete deletes the plan collaborator for a given id
 func (s *Store) PlanCollaboratorDelete(
+	tx *sqlx.Tx,
```
StevenWadeOddball, line 111: If changing the signature, I think this is good opportunity to update the method to match the new paradigm (package level function vs receiver method)

pkg/graph/resolvers/plan_collaborator.go
```diff
@@ -161,8 +162,39 @@ func PlanCollaboratorDelete(logger *zap.Logger, id uuid.UUID, principal authenti
 	if err != nil {
 		return nil, err
 	}
-	retCollaborator, err := store.PlanCollaboratorDelete(logger, id, principal.Account().ID)
-	return retCollaborator, err
+
+	return sqlutils.WithTransaction(store, func(tx *sqlx.Tx) (*models.PlanCollaborator, error) {
+		// First, update any milestones assigned to this collaborator
+		updatedMilestones, err := storage.MTOMilestoneUpdateAssignedToNullByCollaborator(
```
StevenWadeOddball, line 168: Hmm is this necessary because the assignement points to a user not a collaborator? Does a milestone have to be a assigned to a collaborator? Or could anyone be assigned to the milestone? If only a collaborator could be assigned, this could just be cascade logic.

StevenWadeOddball (commented):
Thanks @Peyton232! Some minor comments and questions. The code is looking good however! I have not tested this functionally

PR #2232, opened 2026-04-13
Mint 3649/category and subcat

samoddball:
# MINT-3649

## Description

- return unfiltered list of categories and subcategories

## How to test this change

1.  there is a bruno query for this now in the codebase (`query_examples/MINT/Models to Operations Matrix/Templates/Get MTO Template Categories.bru`)
2. that query can be run and verified that it returns data
3. note - when `Uncategorized` would have been the only subcategory, we return an empty array

## PR Author Checklist

## PR Reviewer Guidelines


yiranli624 (commented):
Test with bruno, looks great!

pkg/graph/schema/types/mto_common_milestone.graphql
```diff
@@ -1,6 +1,11 @@
 """
 MTOCommonMilestone represents a Milestone from the "Common Milestones" library
 """
+type CommonCategory {
```
Peyton232, line 4: nitpick: The comment above this is for MTOCommonMilestone, should move that to be next to that type and then probably have a similar comment for CommonCategory

garyjzhao (approved):
Also tested with Bruno, and it looks good to me

pkg/sqlqueries/SQL/mto/common_milestone/get_common_categories.sql
```diff
@@ -0,0 +1,23 @@
...
+    COALESCE(
+        ARRAY_REMOVE(ARRAY_AGG(DISTINCT child.name ORDER BY child.name), NULL),
+        CAST(ARRAY[] AS TEXT[])
+    ) AS sub_categories
+FROM parent_categories parent
+LEFT JOIN mto_template_category child
+    ON
+        child.parent_id = parent.id
+        AND child.name <> ''
+GROUP BY parent.name
```
Peyton232, line 19: Are the parent's names here globally unique? Other wise may have an edge case of something like 

```
"Housing" (id 1)
"Housing" (id 2)
```

pkg/sqlqueries/SQL/mto/common_milestone/get_common_categories.sql
```diff
@@ -0,0 +1,23 @@
...
+        tc.name
+    FROM mto_template_category tc
+    WHERE
+        tc.parent_id IS NULL
+        AND tc.name <> ''
+)
+
+SELECT
+    parent.name,
+    COALESCE(
```
Peyton232: nitpick: Technically COALESCE here is redundant because ARRAY_AGG returns null if over no rows.

Peyton232 (approved):
Overall looks good to me, couple nitpicks but nothing worth holding up the pr over

pkg/graph/schema/types/mto_common_milestone.graphql
samoddball: oops - nice catch, thanks!

pkg/sqlqueries/SQL/mto/common_milestone/get_common_categories.sql
samoddball: they aren't globally unique, but that is the goal of the group - we only want one example of each parent. there are multiple instances of the same name, but that can happen if a parent is a subcategory of a different parent (the table itself serves as the bridge table, so the many-to-many relationship lives on the table alongside the actual data)

pkg/sqlqueries/SQL/mto/common_milestone/get_common_categories.sql
samoddball: nice catch! removed, thanks

PR #2233, opened 2026-04-13
`facilited_by_other` column and app code. bruno queries

samoddball:
# NOREF

## Description

- add migration for adding `facilitated_by_other` as nullable text column to `mto_common_milestone` table

## How to test this change

1.  i tested this manually - in DB, add some text into the new column, confirm it comes back when running the bruno query `GetMTO Common Milestones Without Model Plan ID`

## PR Author Checklist

## PR Reviewer Guidelines

### DIFF
--- a/migrations/V263__Add_Facilitated_By_Other_To_MTO_Common_Milestone.sql
+++ b/migrations/V263__Add_Facilitated_By_Other_To_MTO_Common_Milestone.sql
@@ -1,0 +1,5 @@
+ALTER TABLE mto_common_milestone
+ADD COLUMN facilitated_by_other TEXT;
+
+COMMENT ON COLUMN mto_common_milestone.facilitated_by_other IS
+'Optional free-text detail used when the facilitated_by_role array includes OTHER.';
--- a/migrations/V263__Add_Facilitated_By_Other_To_MTO_Common_Milestone.sql
+++ b/migrations/V263__Add_Facilitated_By_Other_To_MTO_Common_Milestone.sql
@@ -1,5 +1,16 @@
 ALTER TABLE mto_common_milestone
-ADD COLUMN facilitated_by_other TEXT;
+ADD COLUMN facilitated_by_other ZERO_STRING;
 
 COMMENT ON COLUMN mto_common_milestone.facilitated_by_other IS
 'Optional free-text detail used when the facilitated_by_role array includes OTHER.';
+
+ALTER TABLE mto_common_milestone
+ADD CONSTRAINT mto_common_milestone_check_facilitated_by_other_only_if_other
+CHECK (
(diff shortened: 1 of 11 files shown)


garyjzhao (approved):
Tested on Bruno and LGTM

migrations/V263__Add_Facilitated_By_Other_To_MTO_Common_Milestone.sql
```diff
@@ -0,0 +1,16 @@
+ALTER TABLE mto_common_milestone
+ADD COLUMN facilitated_by_other ZERO_STRING;
+
+COMMENT ON COLUMN mto_common_milestone.facilitated_by_other IS
+'Optional free-text detail used when the facilitated_by_role array includes OTHER.';
+
+ALTER TABLE mto_common_milestone
+ADD CONSTRAINT mto_common_milestone_check_facilitated_by_other_only_if_other
```
Peyton232, line 8: Would we want the inverse constraint here also? 

I.e 
Right now you enforce:
If text exists -> must include OTHER 

But you don’t enforce:
If OTHER is selected -> text must exist 

So currently this would be allowed 
facilitated_by_role = ['OTHER']
facilitated_by_other = NULL

This could also be handled at the application layer but for data integrity this may be something we want to consider

migrations/V263__Add_Facilitated_By_Other_To_MTO_Common_Milestone.sql
```diff
@@ -0,0 +1,16 @@
...
+CHECK (
+    facilitated_by_other IS NULL
+    OR facilitated_by_role @> ARRAY['OTHER']::MTO_FACILITATOR[]
```
Peyton232: facilitated_by_role @> ARRAY['OTHER'] 
If this evaluates to NULL then the constraint could fail unexpectedly. Might be a good place for a COALESCE

Peyton232 (changes requested):
Overall looks good, had a couple things on the sql and specifically I think there could be a potential NULL error on it so could be worth checking out.

migrations/V263__Add_Facilitated_By_Other_To_MTO_Common_Milestone.sql
samoddball: 🦅 👁️  - nice spot. fixing

migrations/V263__Add_Facilitated_By_Other_To_MTO_Common_Milestone.sql
samoddball: nice! updated, and confirmed UI will guard this too

Peyton232 (approved):
Looks good, great job

PR #2240, opened 2026-04-16
Mint 3646/create common milestones

samoddball:
# MINT-3646

## Description

- implements `CreateMTOCommonMilestone` BE infra
- adds bruno query
- allows `section` and `trigger_table` columns to be null on `mto_common_milestone` table

## How to test this change

1.  ensure tests pass
2. run `query_examples/MINT/Models to Operations Matrix/Common Milestones/CreateMTO Common Milestone.bru` and confirm success

## PR Author Checklist

## PR Reviewer Guidelines

### DIFF
--- a/pkg/storage/mto_common_milestone.go
+++ b/pkg/storage/mto_common_milestone.go
@@ -44,7 +44,6 @@ func MTOCommonMilestoneGetByIDLoader(np sqlutils.NamedPreparer, _ *zap.Logger, i
 // MTOCommonMilestoneCreate creates a common milestone library row and its common solution links.
 func MTOCommonMilestoneCreate(
 	np sqlutils.TransactionPreparer,
-	logger *zap.Logger,
 	name string,
 	description string,
 	categoryName string,
@@ -57,7 +56,6 @@ func MTOCommonMilestoneCreate(
 	return sqlutils.WithTransaction(np, func(tx *sqlx.Tx) (*models.MTOCommonMilestone, error) {
 		return createMTOCommonMilestone(
 			tx,
-			logger,
 			name,
 			description,
 			categoryName,
@@ -72,7 +70,6 @@ func MTOCommonMilestoneCreate(
 
 func createMTOCommonMilestone(
 	tx *sqlx.Tx,
-	_ *zap.Logger,
 	name string,
 	description string,
 	categoryName string,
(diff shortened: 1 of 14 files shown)


pkg/storage/mto_common_milestone.go
```diff
@@ -41,6 +41,85 @@ func MTOCommonMilestoneGetByIDLoader(np sqlutils.NamedPreparer, _ *zap.Logger, i
...
+			facilitatedByOther,
+			mtoCommonSolutionKeys,
+			actorUserID,
+		)
+	})
+}
+
+func createMTOCommonMilestone(
+	tx *sqlx.Tx,
+	_ *zap.Logger,
```
Peyton232: Is there a reason are we passing in logger if it is unused?

pkg/storage/mto_common_milestone.go
```diff
@@ -41,6 +41,85 @@ func MTOCommonMilestoneGetByIDLoader(np sqlutils.NamedPreparer, _ *zap.Logger, i
...
+			subCategoryName,
+			facilitatedByRole,
+			facilitatedByOther,
+			mtoCommonSolutionKeys,
+			actorUserID,
+		)
+	})
+}
+
+func createMTOCommonMilestone(
```
Peyton232, line 74: as of right now createMTOCommonMilestone is doing
session setup
ID generation
creation
linking

This isn't too bad but might be cleaner and easier to test if we split it up with something like insertMTOCommonMilestone
linkSolutionsToMilestone

Peyton232 (approved):
Overall looks good and functions as expected. Couple nitpicks but nothing here seems majorly concerning.

pkg/storage/mto_common_milestone.go
samoddball: i really just followed the pattern - i don't like it and will happily remove for this case. the `_ *zap.Logger` is everywhere

samoddball: this is removed for the vertical

pkg/storage/mto_common_milestone.go
samoddball: yep good call - updated, thanks

Peyton232 (approved):
Looks good now, great job

PR #2251, opened 2026-04-21
[MINT-3672] Disable model timeline application period

aterstriep:
# MINT-3672

[Figma](https://www.figma.com/design/qXhO12xSSkDF4LaToGKJoV/MINT-Blue-Sky-Ideas?node-id=5524-51516&t=8kPcaNKvRKNgf1wg-1)

## Description
- Timeline form
  - Disabled application period fields if model type is mandatory 
  - Added application period help text and model type
- Model basics form
  - Updated "Model type" field label capitalization
  - Added model type field help text

## How to test this change
1.  Go to seeded model "Plan with Timeline"
2. Go to timeline form
3. Info box with help text and model type should be displayed
    - Model type should display "No answer provided"
4. Application period fields should be enabled
5. Go to second page of model basics form
6. For model type, select "Mandatory national"
7. Save form
8. Go back to timeline form
9. Model type should be displayed in info box
10. Application period fields should be disabled

## PR Author Checklist

- [x] Updated the [BRUNO Collection](../query_examples/MINT) if necessary.

## PR Reviewer Guidelines


yiranli624 (changes requested):
Disable feature works great! Missing delete data behavior:
- If data exists in either Application period date fields, delete it whenever either of the mandatory model types is selected and saved. Attribute these changes to the current user in Change History

src/features/ModelPlan/TaskList/Basics/Overview/index.tsx
```diff
@@ -81,6 +87,42 @@ const Overview = () => {
...
 
+  const { data: timelineData } = useGetTimelineQuery({
+    variables: {
+      id: modelID
+    }
+  });
+
+  const { id: timelineId = '' } = (timelineData?.modelPlan?.timeline ||
+    {}) as GetTimelineQuery['modelPlan']['timeline'];
+
```
yiranli624: Instead of using one more query, could update GetOverviewQuery to include timeline id.

src/features/ModelPlan/TaskList/Basics/Overview/index.tsx
```diff
@@ -81,6 +87,42 @@ const Overview = () => {
...
+  const [updateTimeline] = useMutation(TypedUpdateTimelineDocument, {
+    refetchQueries: [
+      { query: GetTimelineDocument, variables: { id: timelineId } }
+    ]
+  });
+
```
yiranli624: missing error handling here for updateTimeline mutation.

yiranli624 (changes requested):
the deleting timeline data works if I click 'Save and start next Model Plan section', but not working if I click 'back' or 'save and return'. Could you add the same behavior to those buttons too please :)

yiranli624 (dismissed):
I'm so sorry for the late turn around! On a closer look and talked to @Peyton232, I think we should take a completely different approach. The delete timeline data should happen in the backend, in the resolver for where the user selects Mandatory model type, or a new trigger to take care of it. Either way, this logic should not be living in the frontend.

Additionally, need to confirm with Jeff about if we want to edit the existing data. IE, what about the models that user already selected mandatory type, do we delete the timeline data?

aterstriep:
Changes were made to move the logic for clearing application dates to the backend, and frontend logic was removed.

yiranli624:
Great work Ashley! FE looks great! Waiting for Peyton to review BE code.

Peyton232 (dismissed):
Works locally and code looks good, approved

Peyton232 (approved):
reapproval

PR #2266, opened 2026-05-05
added data loader for is collaborator

Peyton232:
# [MINT-3707](https://jiraent.cms.gov/browse/MINT-3707)

## Description

Replaces the per-request SQL call for ModelPlan.isCollaborator with a dataloader that batches all collaborator checks in a single query when multiple model plans are resolved in the same request.

## How to test this change

1. Create a bunch of model plans and add yourself as collaborator to some of them. 
2.  Log in and navigate to the model plans list (Home page) - this triggers GetModelPlans which resolves isCollaborator for every plan in the collection, so it's the highest-value batch scenario
3. Confirm the page loads correctly and plan cards show the correct edit/read-only access (collaborator plans should show edit controls, non-collaborator plans should not)
4. Open any individual model plan you're a collaborator on and navigate to a section (e.g. Basics) - GetAllBasics also fetches isCollaborator and should still gate access correctly

## PR Author Checklist

## PR Reviewer Guidelines

### DIFF
--- a/pkg/graph/resolvers/plan_collaborator.go
+++ b/pkg/graph/resolvers/plan_collaborator.go
@@ -14,7 +14,6 @@ import (
 	"github.com/cms-enterprise/mint-app/pkg/storage/loaders"
 	"github.com/cms-enterprise/mint-app/pkg/userhelpers"
 
-	"github.com/cms-enterprise/mint-app/pkg/appcontext"
 	"github.com/cms-enterprise/mint-app/pkg/authentication"
 	"github.com/cms-enterprise/mint-app/pkg/graph/model"
 	"github.com/cms-enterprise/mint-app/pkg/models"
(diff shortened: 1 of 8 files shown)


pkg/graph/resolvers/plan_collaborator.go
```diff
@@ -167,8 +168,12 @@ func PlanCollaboratorGetByID(ctx context.Context, id uuid.UUID) (*models.PlanCol
 	return loaders.PlanCollaborators.ByID.Load(ctx, id)
 }
 
-// IsPlanCollaborator checks if a user is a collaborator on model plan is a favorite.
-func IsPlanCollaborator(logger *zap.Logger, principal authentication.Principal, store *storage.Store, modelPlanID uuid.UUID) (bool, error) {
-	// Future Enhancement: Consider making this a dataloader.
-	return store.CheckIfCollaborator(logger, principal.Account().ID, modelPlanID)
+// IsPlanCollaborator checks if the current principal is a collaborator on a model plan using a DataLoader.
+func IsPlanCollaborator(ctx context.Context, modelPlanID uuid.UUID) (bool, error) {
```
StevenWadeOddball: Thoughts on leaving the principal in the arg? I think that makes this more versatile (eg check if user XYZ is a collaborator, not just the current principal.

Peyton232: I like this because it uncouples it from the current principal, pushing a change now

StevenWadeOddball (dismissed):
This looks good! Thanks Peyton! I have not pulled this locally, but validating based on the code change and the test cases looking to cover this

samoddball (dismissed):
code looks good, and so does local testing!

StevenWadeOddball (approved):

PR #2277, opened 2026-05-12
[MINT-3722] get new s3 creds if needed

samoddball:
# MINT-3722

## Description

- gets new s3 creds in the event they expire/fail

## How to test this change

1.  ensure tests pass (we mock cred expiration in tests)
2. not super feasible to test via manual use

## PR Author Checklist

## PR Reviewer Guidelines


pkg/echimpcache/echimp_cache.go
```diff
@@ -64,31 +67,40 @@ func (c *crAndTDLCache) IsOld(viperConfig *viper.Viper) bool {
...
+}
+
+func (c *crAndTDLCache) refreshCacheWithReader(client *s3.S3Client, viperConfig *viper.Viper, logger *zap.Logger, readFromS3 crAndTDLReader) error {
 	CRKey := viperConfig.GetString(appconfig.AWSS3ECHIMPCRFileName)
 	TDLKey := viperConfig.GetString(appconfig.AWSS3ECHIMPTDLFileName)
 
-	crsRaw, err := parquet.ReadFromS3[*models.EChimpCRRaw](client, CRKey)
+	crsRaw, tdlsRaw, err := readFromS3(client, CRKey, TDLKey)
+	if err != nil {
+		if s3.S3ErrorHasExpiredCredentials(err) {
```
StevenWadeOddball: @samoddball , maybe this isn't the right place to refresh the client? Perhaps it can be backed into the S3 abstractions instead? EG at the download phase?

pkg/echimpcache/echimp_cache.go
```diff
@@ -17,6 +18,8 @@ import (
 
 var CRAndTDLCache *crAndTDLCache
 
+type crAndTDLReader func(client *s3.S3Client, crKey string, tdlKey string) ([]*models.EChimpCRRaw, []*models.EChimpTDLRaw, error)
```
StevenWadeOddball: we have test data in the repo already so I don't think you need to introduce this extra abstraction, 
see echimptestdata.SeedEChimpTestData(config.EChimpS3Client, viperConf)

pkg/s3/download.go
```diff
@@ -10,13 +10,12 @@ import (
...
-func (c S3Client) GetS3ObjectReaderAt(key string) (io.ReaderAt, int64, error) {
+func (c *S3Client) GetS3ObjectReaderAt(key string) (io.ReaderAt, int64, error) {
 	// Get the S3 object
-	resp, err := c.client.GetObject(context.TODO(), &s3.GetObjectInput{
+	resp, err := c.currentClient().GetObject(context.TODO(), &s3.GetObjectInput{
 		Bucket: &c.config.Bucket,
 		Key:    &key,
 	})
-
 	if err != nil {
```
StevenWadeOddball, line 21: perhaps you could try to handle refresh credentials here instead? This would help in case the file upload bucket creds ever expire as well.

StevenWadeOddball (changes requested):
Thanks Sam, this is looking good. A couple of suggestions for moving the retry logic. I think Copilot has some worthwhile feedback here as well.

pkg/graph/resolvers/model_plan_test.go
StevenWadeOddball, line 1: We have a test context we use in MINT which has boilerplate  configs etc. That's preferrable than ctx.Background for test suite.

pkg/echimpcache/echimp_cache.go
```diff
@@ -110,6 +113,14 @@ func (c *crAndTDLCache) refreshCache(client *s3.S3Client, viperConfig *viper.Vip
 
 }
 
+func readCRsFromS3(ctx context.Context, client *s3.S3Client, key string) ([]*models.EChimpCRRaw, error) {
```
StevenWadeOddball: NIT: does it make sense to leave these parquet wrapper functions in line? These functions are a net-increase in code, and only called in one spot.

pkg/echimpcache/echimp_cache.go
```diff
@@ -101,6 +83,15 @@ func (c *crAndTDLCache) refreshCacheWithReader(client *s3.S3Client, viperConfig
 	}
 	c.CRs = sanitizedCRS
```
StevenWadeOddball: you didn't do this, but would you mind correcting the double assignment here while you are in the files? (also happens ~ line 99)

StevenWadeOddball (commented):
Thanks Sam, this is looking good. I think it will be great to have this re-auth baked into the client methods

I think it could help future usability to add some function comments to some of the new S3 code. Could you take a look at that please? I find myself a little lost why I'd call withCredentialRefresh / vs with Builder vs withRewind. Even though these aren't exported, I think some comments would go a long way to making this a lot easier to use these methods in the future. 


I haven't done functional testing either, commenting just so feedback is visible

pkg/s3/errors.go
```diff
@@ -8,12 +8,18 @@ import (
 
 const noSuchKeyErrCode = "NoSuchKey"
 
+var expiredCredentialErrCodes = map[string]struct{}{
+	"ExpiredToken":          {},
+	"ExpiredTokenException": {},
+	"InvalidToken":          {},
+	"RequestExpired":        {},
+	"TokenRefreshRequired":  {},
```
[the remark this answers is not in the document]
samoddball: added `InvalidClientTokenId` but not the others to avoid noise

pkg/echimpcache/echimp_cache.go
StevenWadeOddball, line 1: I don't think it is a problem for this PR, but it does seem that the a lot more is touched in this file than would need to be ( refactoring receiver methods to be functions etc). I think this is fine, maybe a better choice, but I think it would be good to consider how to limit the agents to be more targeted to the issue.

StevenWadeOddball (approved):
Thanks @samoddball! I think that this looks good! I think it should be good to merge. If you want more confidence before going to the deploy pipeline, I think this would be fine to kick off a manual deploy to dev

pkg/echimpcache/echimp_cache.go
samoddball: agree!

PR #2290, opened 2026-05-19
[MINT-3742][MINT-3743][MINT-3750][MINT-3749][MINT-3753] graphql and sql support for ctat ticket and creation

samoddball:
# MINT-3742, MINT-3743, MINT-3750, MINT-3749, MINT-3753

## Description

- introduces enum sets for new dropdown options
- introduces Graphql support (stubs) for the other ctat fields, including admin fields
- introduces sql support to match

## How to test this change

1.  no tests for this change, just ensure builds/lint/tests pass
2. the DB constraints will be heavily tested in the upcoming app code work
3. NOTE: known _potential_ issue. The sequential human readable ID number will not reset on a truncate. We can fix this in a handful of ways, but it is only an issue for tests. We can either RESET IDENTITY while truncating (which would reset any other IDENTITY-style columns), or we can manually RESET it in tests - tbd in a follow-up PR

## PR Author Checklist

## PR Reviewer Guidelines


pkg/graph/schema/types/ctat_request.graphql
```diff
@@ -1,10 +1,185 @@
...
+}
+
+"""
+A request for CTAT assistance
 """
 type CTATRequest {
   id: UUID!
   humanReadableID: String!
+  requester: UUID!
+
```
StevenWadeOddball, line 162: uuid is a good format for the data! We generally have a resolver for the account as well.

```suggestion
requesterUserAccount: UserAccount!
```

pkg/graph/schema/types/ctat_request.graphql
```diff
@@ -14,7 +189,52 @@ type CTATRequest {
...
+  status: CTATStatus
+  assignedAdmin: UUID
+  notes: String
+  resolution: String
+}
+
 extend type Query {
-  ctatRequest(id: UUID!): CTATRequest
+  ctatRequest(id: UUID!): CTATRequest!
   ctatRequests: [CTATRequest!]
```
StevenWadeOddball: I think requiring the array matches the paradigm
```suggestion
  ctatRequests: [CTATRequest!]!
```

pkg/graph/schema/types/ctat_request.graphql
```diff
@@ -14,7 +189,52 @@ type CTATRequest {
...
+  dateAssistanceNeededBy: Time!
+  supportingDocuments: [CTATRequestDocumentInput!]
+}
+
+"""
+Input for an admin update to a CTAT request.
+"""
+input CTATAdminUpdateInput {
+  status: CTATStatus
+  assignedAdmin: UUID
```
StevenWadeOddball: I think this should probably be a username instead of UUID (see how it's handled for collaborator create input https://github.com/CMS-Enterprise/mint-app/blob/06621738e885d8b5e2440ae05e62419cb5517798/pkg/graph/schema/types/plan_collaborator.graphql#L38

pkg/graph/schema/types/ctat_request.graphql
```diff
@@ -14,7 +189,52 @@ type CTATRequest {
...
+  describeHelpNeeded: String!
+  requestUrgency: CTATRequestUrgency!
+  dateAssistanceNeededBy: Time!
+  supportingDocuments: [CTATRequestDocumentInput!]
+}
+
+"""
+Input for an admin update to a CTAT request.
+"""
+input CTATAdminUpdateInput {
```
StevenWadeOddball: I wonder if it will make sense to use map[string]any here? Largely an implementation time question, but it would allow map changes to be called instead of manually handling all input.
```suggestion
input CTATAdminUpdateInput@goModel(model: "map[string]any") {

```

pkg/graph/schema/types/ctat_request.graphql
```diff
@@ -1,10 +1,185 @@
...
+"""
+A request for CTAT assistance
 """
 type CTATRequest {
   id: UUID!
   humanReadableID: String!
+  requester: UUID!
+
+  status: CTATStatus!
+  assignedAdmin: UserInfo
```
StevenWadeOddball: I think this should probably get the same approach as the requester etc. This is more in line with the other account paradigm in app

```suggestion
assignedAdmin: UUID
assignedAdminUserAccount: UserAccount
```

StevenWadeOddball (commented):
thanks Sam, haven't looked at the table in great detail yet, but wanted to give a little type feedback in the meantime!

migrations/V268__Add_CTAT_tables.sql
```diff
@@ -0,0 +1,299 @@
...
+        RETURNING last_number INTO NEW.human_readable_id_number;
+    END IF;
+
+    RETURN NEW;
+END;
+$body$ LANGUAGE plpgsql;
+
+CREATE TABLE ctat_request (
+    id UUID PRIMARY KEY NOT NULL,
+    human_readable_id_prefix ZERO_STRING NOT NULL DEFAULT 'CTAT',
```
StevenWadeOddball: I think we can discard this table, I think there can be handled all in the resolver code to append 'CTAT'. We only anticipate one type of request, so probably wont need this.

migrations/V268__Add_CTAT_tables.sql
```diff
@@ -0,0 +1,299 @@
...
+    human_readable_id_number INTEGER NOT NULL CHECK (human_readable_id_number > 0),
```
StevenWadeOddball: I'm not sure that a trigger is needed here, instead of native postgres functionality
Can you try some other options instead please? If a trigger can be avoided, that is a preferable option.

This below will auto-increment the column, but it doesn't allow overriding it (unless you explicitly tell it too). 
```suggestion
    human_readable_id_number INTEGER GENERATED ALWAYS AS IDENTITY,
```

I think simplifying would be ideal here.

migrations/V268__Add_CTAT_tables.sql
```diff
@@ -0,0 +1,299 @@
...
+SELECT audit.AUDIT_TABLE(
+    'public',
+    'ctat_request_document',
+    'id',
+    'ctat_request_id',
+    '{created_by,created_dts,modified_by,modified_dts}'::TEXT[],
+    '{*,id,ctat_request_id}'::TEXT[]
+);
+
+SELECT audit.AUDIT_TABLE(
```
StevenWadeOddball: Note, I'm not sure anyone updated this function, but the audit table does have the concept of a secondary foreign key for situations like this. (see mto_milestone_solution_link). This would allow model_plan_id to also be linked.

migrations/V268__Add_CTAT_Request_table.sql
StevenWadeOddball, line 1: Thanks Sam, `nit` I wonder if organizationally this might be better with a migration per table? Your call

migrations/V268__Add_CTAT_tables.sql
```diff
@@ -0,0 +1,299 @@
...
+COMMENT ON COLUMN ctat_request_model_plan_link.model_plan_id IS 'The related MINT model plan.';
+COMMENT ON COLUMN ctat_request_model_plan_link.created_by IS 'The user who created the CTAT request model plan link.';
+COMMENT ON COLUMN ctat_request_model_plan_link.created_dts IS 'The timestamp when the CTAT request model plan link was created.';
+COMMENT ON COLUMN ctat_request_model_plan_link.modified_by IS 'The user who last modified the CTAT request model plan link.';
+COMMENT ON COLUMN ctat_request_model_plan_link.modified_dts IS 'The timestamp when the CTAT request model plan link was last modified.';
+
+-- Register the CTAT tables with the existing audit pipeline so inserts/updates/deletes are
+-- captured in audit.change and translated audit views. We explicitly include IDs/foreign keys in
+-- insert_fields where helpful so initial create events retain enough context to be understandable.
+SELECT audit.AUDIT_TABLE(
```
StevenWadeOddball, line 168: Worth calling out, these will need to be left out of the translated audit flow in the ignored table list. The translated audit table assumes that this is linked to a model plan. Only the linking table really fits this paradigm, but I think it is left out of scope for this.

pkg/models/ctat_enums.go
StevenWadeOddball, line 1: nit, maybe this should be called ctat_enums? This sounds like there should be a model called ctat

pkg/models/ctat_request.go
```diff
@@ -0,0 +1,55 @@
...
+type CTATRequest struct {
+	baseStruct
+
+	Requester  uuid.UUID  `json:"requester" db:"requester"`
+	Status     CTATStatus `json:"status" db:"status"`
+	Notes      *string    `json:"notes,omitempty" db:"notes"`
+	Resolution *string    `json:"resolution,omitempty" db:"resolution"`
+
+	AssignedAdminID *uuid.UUID `json:"-" db:"assigned_admin"`
+	AssignedAdmin   *UserInfo  `json:"assignedAdmin,omitempty" db:"-"`
```
StevenWadeOddball: I'd remove this from the type and let it be in the resolvers (or you can make it a receiver method like we do with created_by_user account etc)

pkg/models/ctat_request.go
```diff
@@ -0,0 +1,55 @@
...
+
+	TypeOfHelpNeeded       EnumArray[CTATHelpNeededType] `json:"typeOfHelpNeeded" db:"type_of_help_needed"`
+	TypeOfHelpNeededOther  *string                       `json:"typeOfHelpNeededOther,omitempty" db:"type_of_help_needed_other"`
+	DescribeHelpNeeded     string                        `json:"describeHelpNeeded" db:"describe_help_needed"`
+	RequestUrgency         CTATRequestUrgency            `json:"requestUrgency" db:"request_urgency"`
+	DateAssistanceNeededBy time.Time                     `json:"dateAssistanceNeededBy" db:"date_assistance_needed_by"`
+
+	SupportingDocuments []*CTATRequestDocument `json:"supportingDocuments,omitempty" db:"-"`
+
+	HumanReadableIDPrefix string `json:"-" db:"human_readable_id_prefix"`
```
StevenWadeOddball: I think this can be simplified, prefix should always be CTAT. (or simple to add in the future if we need it, but we might make another table instead)

pkg/models/ctat_request.go
```diff
@@ -0,0 +1,55 @@
...
+	HumanReadableIDNumber int    `json:"-" db:"human_readable_id_number"`
```
StevenWadeOddball: Maybe worth exposing on the FE too? So no string parsing is needed if convenient

pkg/models/table_names.go
```diff
@@ -21,6 +21,10 @@ const (
 	TNActivity                         TableName = "activity"
 	TNAnalyzedAudit                    TableName = "analyzed_audit"
 	TNDiscussionReply                  TableName = "discussion_reply"
+	TNCTATRequest                      TableName = "ctat_request"
```
StevenWadeOddball, line 24: Could be good to add directly to untranslated table list now (otherwise there will be lots of error messages when this gets deployed)
https://github.com/CMS-Enterprise/mint-app/blob/a783110fbd03c69a6f9a6dc38fc51f51f83f44f0/pkg/translatedaudit/translated_audit.go#L23-L52

StevenWadeOddball (changes requested):
Thanks @samoddball , this is looking good. I have some high level comments about the structure, I didn't review all enum values or field, but focused on larger scale questions about details about auto-incrementing etc. 

Constraints, enum values, fields etc I will leave to @garyjzhao to review. These will also self reveal as development goes on if anything needs to be added.

pkg/graph/schema/types/ctat_request.graphql
```diff
@@ -14,7 +189,52 @@ type CTATRequest {
...
+
 extend type Query {
-  ctatRequest(id: UUID!): CTATRequest
+  ctatRequest(id: UUID!): CTATRequest!
   ctatRequests: [CTATRequest!]
 }
+
+extend type Mutation {
+  createCTATRequest(input: CTATRequestInput!): CTATRequest!
+  adminUpdateCTATRequest(input: CTATAdminUpdateInput!): CTATRequest!
```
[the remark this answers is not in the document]
samoddball: added `id` to the `input` type

src/i18n/en-US/ctatRequest.ts
```diff
@@ -0,0 +1,227 @@
...
+  [CtatContractActivityType.IMPLEMENTATION]: 'Implementation',
+  [CtatContractActivityType.LEARNING]: 'Learning',
+  [CtatContractActivityType.TECHNICAL_ASSISTANCE]: 'Technical Assistance',
+  [CtatContractActivityType.OTHER]: 'Other'
+};
+
+export const contractTypes: Record<CtatContractType, string> = {
+  [CtatContractType.COST_PLUS_FIXED_FEE]: 'Cost Plus Fixed Fee',
+  [CtatContractType.COST_REIMBURSEMENT]: 'Cost Reimbursement',
+  [CtatContractType.FIRM_FIXED_PRICE]: 'Firmed Fixed Price',
```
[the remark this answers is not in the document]
samoddball: this is how it is designated in designs, and i have fixed the const to match (firmed)

pkg/models/ctat_request.go
samoddball: updated and removed/simplified

pkg/models/ctat_request.go
samoddball: are you referring to exposing both the full ticket number (ABC-123) and also just the number suffix (123)? if so, can do!

pkg/graph/resolvers/ctat_request.resolvers.go
```diff
@@ -10,15 +10,37 @@ import (
...
 	"github.com/google/uuid"
 
+	"github.com/cms-enterprise/mint-app/pkg/graph/generated"
 	"github.com/cms-enterprise/mint-app/pkg/graph/model"
+	"github.com/cms-enterprise/mint-app/pkg/models"
 )
 
+// TypeOfHelpNeeded is the resolver for the typeOfHelpNeeded field.
+func (r *cTATRequestResolver) TypeOfHelpNeeded(ctx context.Context, obj *models.CTATRequest) ([]models.CTATHelpNeededType, error) {
+	return obj.TypeOfHelpNeeded, nil
```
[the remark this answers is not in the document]
samoddball: this does compile, i am not sure it is needed

migrations/V268__Add_CTAT_tables.sql
samoddball: yep agreed - removed

migrations/V268__Add_CTAT_tables.sql
samoddball: i'll just keep these out for now

pkg/models/ctat_request.go
samoddball: good call! moved to resolvers, thanks

pkg/graph/schema/types/ctat_request.graphql
samoddball: nice! thanks - updated

pkg/models/ctat_request.go
StevenWadeOddball: Thanks Sam! Yeah, that's what I was thinking :D

samoddball: done! thanks!

migrations/V268__Add_CTAT_Request_table.sql
samoddball: yep sounds like a nice cleanup - made the change! thanks

migrations/V267__Add_CTAT_Enums.sql
```diff
@@ -1,7 +1,7 @@
 CREATE TYPE CTAT_CONTRACT_TYPE AS ENUM (
     'COST_PLUS_FIXED_FEE',
     'COST_REIMBURSEMENT',
-    'FIRM_FIXED_PRICE',
+    'FIRMED_FIXED_PRICE',
```
StevenWadeOddball, line 4: I did a quick seach, I think you had it correct with `Firm` instead of `Firmed`

migrations/V268__Add_CTAT_Request_table.sql
```diff
@@ -0,0 +1,175 @@
...
+        (
+            cmmi_division != 'OTHER'
+            AND cmmi_division_other IS NULL
+        )
+    ),
+
+    -- Enforces the dependent-dropdown behavior in the database. Each named group only accepts
+    -- divisions from its own namespace (for example BSG can only store BSG_* or division OTHER).
+    -- This keeps invalid cross-group combinations out even if application validation is bypassed.
+    CONSTRAINT ctat_request_cmmi_group_division_check CHECK (
```
StevenWadeOddball, line 76: This might be simplifiable, so it's easier to expand in the future. Perhaps something like below? (I've not tested it)
```suggestion
cmmi_group = 'OTHER'
OR cmmi_division = 'OTHER'
OR cmmi_division::TEXT LIKE cmmi_group::TEXT || '_%'
```

pkg/graph/resolvers/ctat_request.resolvers.go
```diff
@@ -10,15 +10,37 @@ import (
 
 	"github.com/google/uuid"
 
+	"github.com/cms-enterprise/mint-app/pkg/graph/generated"
 	"github.com/cms-enterprise/mint-app/pkg/graph/model"
+	"github.com/cms-enterprise/mint-app/pkg/models"
 )
 
+// TypeOfHelpNeeded is the resolver for the typeOfHelpNeeded field.
+func (r *cTATRequestResolver) TypeOfHelpNeeded(ctx context.Context, obj *models.CTATRequest) ([]models.CTATHelpNeededType, error) {
```
StevenWadeOddball, line 19: Not critical, but do you have any idea why this isn't auto resolving from gqlgen? Not worth a lot of thought, but curious.

pkg/graph/schema/types/ctat_request.graphql
```diff
@@ -14,7 +192,53 @@ type CTATRequest {
...
+Input for uploading a supporting document for a CTAT request.
+"""
+input CTATRequestDocumentInput {
+  fileData: Upload!
+}
+
+"""
+Input for creating a CTAT request.
+"""
+input CTATRequestInput {
```
StevenWadeOddball, line 205: No change needed now, but wanted to note, this might be simpler as `@goModel(model: "map[string]any")` as well. No functional impact on the FE, so I'd say leave it, but think about it when it comes time to implement the resolver in case it simplifies manual mapping. BaseStructPre-Create should handle it well otherwise.

pkg/models/ctat_request.go
```diff
@@ -0,0 +1,88 @@
...
+	Resolution *string    `json:"resolution,omitempty" db:"resolution"`
+
+	AssignedAdmin *uuid.UUID `json:"assignedAdmin,omitempty" db:"assigned_admin"`
+
+	CmmiGroup         CTATCMMIGroupOption     `json:"cmmiGroup" db:"cmmi_group"`
+	CmmiGroupOther    *string                 `json:"cmmiGroupOther,omitempty" db:"cmmi_group_other"`
+	CmmiDivision      *CTATCMMIDivisionOption `json:"cmmiDivision,omitempty" db:"cmmi_division"`
+	CmmiDivisionOther *string                 `json:"cmmiDivisionOther,omitempty" db:"cmmi_division_other"`
+
+	RelatedMINTModels []uuid.UUID `json:"relatedMINTModels,omitempty" db:"-"`
```
StevenWadeOddball, line 30: This is should be a resolver right? I don't think it needs to be on the model. In MINT, fields like this aren't generally put on in the models pacakge themself, unless there are extenuating circumstances.

pkg/models/ctat_request.go
```diff
@@ -0,0 +1,88 @@
...
+	ContractType              *CTATContractType         `json:"contractType,omitempty" db:"contract_type"`
+	ContractTypeOther         *string                   `json:"contractTypeOther,omitempty" db:"contract_type_other"`
+
+	TypeOfHelpNeeded       EnumArray[CTATHelpNeededType] `json:"typeOfHelpNeeded" db:"type_of_help_needed"`
+	TypeOfHelpNeededOther  *string                       `json:"typeOfHelpNeededOther,omitempty" db:"type_of_help_needed_other"`
+	DescribeHelpNeeded     string                        `json:"describeHelpNeeded" db:"describe_help_needed"`
+	RequestUrgency         CTATRequestUrgency            `json:"requestUrgency" db:"request_urgency"`
+	DateAssistanceNeededBy time.Time                     `json:"dateAssistanceNeededBy" db:"date_assistance_needed_by"`
+
+	SupportingDocuments []*CTATRequestDocument `json:"supportingDocuments,omitempty" db:"-"`
```
StevenWadeOddball, line 45: Same as related models, generally we don't include on the model

StevenWadeOddball (approved):
Thanks Sam, this looks good!

There are a couple of things that probably should be updated here, but I'm approving since this is going to a feature branch. Feel free in quick follow up, or here if that makes sense to you!

pkg/models/ctat_request.go
samoddball: this has been addressed in follow ups!

pkg/models/ctat_request.go
samoddball: this has been addressed in follow ups!

pkg/graph/resolvers/ctat_request.resolvers.go
samoddball: it's something to do with the enum array type that comes out of the DB that can be returned (directly, somehow) as the proper Go-typed slice. i will look further into, but i found some examples elsewhere of the same thing

samoddball: here is an example of the same thing

```go
// DataMonitoringFileTypes is the resolver for the dataMonitoringFileTypes field.
func (r *iDDOCQuestionnaireResolver) DataMonitoringFileTypes(ctx context.Context, obj *models.IDDOCQuestionnaire) ([]models.IDDOCFileType, error) {
	return obj.DataMonitoringFileTypes, nil
}
```

where `DataMonitoringFileTypes` looks like this on the model 

```go
DataMonitoringFileTypes   EnumArray[IDDOCFileType] `json:"dataMonitoringFileTypes" db:"data_monitoring_file_types"`
```

migrations/V267__Add_CTAT_Enums.sql
samoddball: this is fixed in a follow up!

PR #2302, opened 2026-05-26
sql fix for template errors

Peyton232:
# [MINT-3659](https://jiraent.cms.gov/browse/MINT-3659)

## Description

ON CONFLICT DO NOTHING suppresses RETURNING, so when a milestone already existed, GetProcedure got sql: no rows in result set and errored. Added the same UNION ALL fallback the category SQL already uses to return the existing row instead.

## How to test this change

Apply the same MTO template twice to the same model plan — second apply should succeed silently with no error logged. (if testing through UI, can't apply template twice, so instead add milestones that are in a template and then apply the template.)

## PR Author Checklist

## PR Reviewer Guidelines

### DIFF
--- a/pkg/sqlqueries/SQL/mto/milestone/create_allow_conflicts.sql
+++ b/pkg/sqlqueries/SQL/mto/milestone/create_allow_conflicts.sql
@@ -29,7 +29,7 @@ WITH retVal AS (
         :created_by,
         CURRENT_TIMESTAMP
     )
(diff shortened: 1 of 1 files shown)


StevenWadeOddball (approved):
Thanks @Peyton232 , this works well. Copilots suggestions about including additional fields make sense, but at the same time, it is not important for the functionality in question. This successfully suppresses the errors

PR #2319, opened 2026-06-02
[MINT-3754] Mint 3754/get CTAT ticket

samoddball:
# MINT-3754

## Description

- implements "Get Ticket" slice for CTAT

## How to test this change

1.  ensure build/lint/tests pass
2. ensure you can query using the new "Get CTAT Request.bru" file _after_ seeding the db

## PR Author Checklist

## PR Reviewer Guidelines


pkg/graph/resolvers/ctat_request.go
```diff
@@ -30,6 +30,11 @@ func CTATRequestCollectionGetForAdmin(ctx context.Context, store *storage.Store)
 	return storage.CTATRequestCollectionGetForAdmin(store)
 }
 
+// CTATRequestGetByID implements resolver logic to get a CTAT request by ID.
+func CTATRequestGetByID(_ context.Context, id uuid.UUID, store *storage.Store) (*models.CTATRequest, error) {
```
StevenWadeOddball: This should probably just use the dataloader

pkg/storage/ctat_request.go
```diff
@@ -29,6 +29,18 @@ func CTATRequestCreate(np sqlutils.NamedPreparer, request *models.CTATRequest) (
 	return createdRequest, nil
 }
 
+// CTATRequestGetByID returns a CTAT request by ID.
+func CTATRequestGetByID(np sqlutils.NamedPreparer, id uuid.UUID) (*models.CTATRequest, error) {
```
StevenWadeOddball: Can this be removed in favor of the dataloader?

StevenWadeOddball (commented):
thanks Sam, looks good, I think at a high level we can rely on a data loader instead of a single store method for get. Thoughts?

pkg/graph/resolvers/ctat_request.go
samoddball: we can add a loader! we have one now for getting by requester ID, but we certainly can have one for by ID - adding now, thanks!

pkg/storage/ctat_request.go
samoddball: removed!

pkg/graph/resolvers/ctat_request.resolvers.go
```diff
@@ -92,7 +93,7 @@ func (r *mutationResolver) AdminUpdateCTATRequest(ctx context.Context, input map
 
 // CtatRequest is the resolver for the ctatRequest field.
 func (r *queryResolver) CtatRequest(ctx context.Context, id uuid.UUID) (*models.CTATRequest, error) {
-	return CTATRequestGetByID(ctx, id, r.store)
```
StevenWadeOddball: maybe good to leave this function and include the wrapper on the dataloader there?

StevenWadeOddball (approved):
Thanks Sam, looks good code wise. I'll leave functionality to @garyjzhao . Only one nit about wrapping the data loader in the resolver file. Non-blocking but could be nice.

StevenWadeOddball (approved):

pkg/graph/resolvers/ctat_request.resolvers.go
samoddball: fixed!

garyjzhao (approved):
amaze amaze amaze. tested locally and bruno is working as expected

PR #2331, opened 2026-06-10
Mint 3759/enum naming updates

samoddball:
# MINT-3759

## Description

- naming updates based on figma changes

## How to test this change

1.  ensure enums now match figma

## PR Author Checklist

## PR Reviewer Guidelines


garyjzhao (approved):
code looks good to me. will defer to BE team to properly review the `go` files and migrations

.github/workflows/check_migrations.yml
StevenWadeOddball, line 1: Thanks, this is a trade off. If you make a change that would break the check in a merge to a feature branch, we won't know until the feature branch is targeting main in a PR. (So we don't know as early). However, this stops the false positive failures we have been seeing, so I think that it is a reasonable move.

StevenWadeOddball (approved):

PR #2333, opened 2026-06-10
[MINT-3758] Mint 3758/enum findings cleanup

samoddball:
# MINT-3758

## Description

- clean up zero strings
- clean up unneeded storage tests
- clean up seed and tests to remove custom sql and redundant custom data type

## How to test this change

1.  ensure build/tests/lint pass
2. ensure bruno queries for creating/updating ctat tickets work

## PR Author Checklist

## PR Reviewer Guidelines


cmd/dbseed/ctat.go
```diff
@@ -469,21 +468,21 @@ func (s *Seeder) createCTATRequest(input ctatRequestSeedInput) *models.CTATReque
 	return &models.CTATRequest{
```
StevenWadeOddball: we normally rely on resolvers in DB seed. I was going to suggest using that newCTATRequest from the resolvers package, but it probably makes sense to refactor to be based on resolver code instead of embedding SQL here (CTATRequestCreate etc)

pkg/graph/resolvers/ctat_request.go
```diff
@@ -114,7 +113,7 @@ func CTATRequestAdminUpdate(
 			return nil, fmt.Errorf("resolution must be a string")
 		}
 
-		existing.Resolution = zero.StringFromPtr(resolution)
+		existing.Resolution = resolution
 	}
 
 	err = BaseStructPreUpdate(logger, existing, nil, principal, store, false, false)
```
StevenWadeOddball: Should this be `true`, `true`? EG --> Apply changes instead of mapping manually like above, and also verify the user has permission (in this case, that they are an admin.)

pkg/graph/resolvers/ctat_request_test.go
```diff
@@ -285,12 +284,12 @@ func (suite *ResolverSuite) TestAdminUpdateCTATRequestUpdatesResolution() {
 	})
 	suite.NoError(err)
 	suite.NotNil(resp)
-	suite.Equal(zero.StringFrom(resolution), resp.Resolution)
+	suite.Equal(resolution, *resp.Resolution)
```
StevenWadeOddball, line 272: potential nil pointer dereference here

pkg/graph/resolvers/ctat_request_test.go
```diff
@@ -285,12 +284,12 @@ func (suite *ResolverSuite) TestAdminUpdateCTATRequestUpdatesResolution() {
...
 
 	rows, err := storage.CTATRequestGetByIDLOADER(suite.testConfigs.Store, []uuid.UUID{request.ID})
 	suite.NoError(err)
 	suite.Len(rows, 1)
-	suite.Equal(zero.StringFrom(resolution), rows[0].Resolution)
+	suite.Equal(resolution, *rows[0].Resolution)
```
StevenWadeOddball: potential nil pointer dereference here

pkg/graph/resolvers/ctat_request_test.go
StevenWadeOddball, line 1: Can you review this file for nil pointer dereferences? MINT doesn't use `Assertions` = `Require()` like EASi does, so when dereferencing, you need to check that it's first not nil, or you can set up the test to use require if there is a check earlier. The test will still fail with a dereference, but it looses contextually useful information with a panic

pkg/storage/ctat_request_test.go
StevenWadeOddball, line 1: Typically, we just rely on the resovler test suite for this functionality and don't make a separate store test file. Unless there is distinct reason to assert this in test, I'd suggest it stays in the resolvers package.

pkg/graph/resolvers/ctat_request_test.go
StevenWadeOddball, line 1: This also might be a follow up, but in general we don't write bespoke SQL for the test suites. Common practice is to use the existing resolvers or helpers that are wrapping a resolver to set up the test as needed.  Unless there is an absolute need, there shouldn't be any bespoke inline SQL for test cases etc

StevenWadeOddball (approved):
Thanks @samoddball , I think that the enum cleanup looks good. I found a couple of other points that should be addressed, but they could be a follow up PR if you want to address the enums here alone.

pkg/storage/ctat_request_test.go
samoddball: made the change! thanks!

pkg/graph/resolvers/ctat_request_test.go
samoddball: checked! thanks!

cmd/dbseed/ctat.go
samoddball: done! thanks!

pkg/storage/ctat_request_test.go
```diff
@@ -1,524 +0,0 @@
-package storage
```
samoddball, line 1: remove unneeded test file

pkg/graph/resolvers/ctat_request.go
samoddball: we have this in graphql `@hasRole(role: MINT_ASSESSMENT)`, but i don't think we need the access check in the `BaseStructPreUpdate` call - i am not sure if it makes sense for there to be a model plan relation/discussion relation/milestone relation on a CTAT ticket?

i have cleaned up the ApplyChanges

pkg/graph/resolvers/ctat_request_test.go
samoddball: done! thanks!

cmd/dbseed/ctat.go
```diff
@@ -87,479 +117,184 @@ func (s *Seeder) seedCTATRequests(
...
-		CMMIDivision:           &bsgDivision,
-		RelatedModelPlanIDs:    []uuid.UUID{planWithBasics.ID, sampleModelPlan.ID},
+	s.createCTATRequest(requesterBTAL, &model.CTATRequestInput{
+		CmmiGroup:              models.CTATCMMIGroupOptionBSG,
+		CmmiDivision:           &bsgDivision,
+		RelatedMINTModels:      []uuid.UUID{planWithBasics.ID, sampleModelPlan.ID},
 		ContractActivityType:   &technicalAssistance,
-		ContractName:           new("CMMI Operational Support Services"),
-		ContractNumber:         new("75FCMC24F0001"),
+		ContractName:           models.StringPointer("CMMI Operational Support Services"),
```
StevenWadeOddball: probably better to keep the new() functionality throughout here

pkg/graph/resolvers/ctat_request.go
```diff
@@ -311,36 +283,40 @@ func buildCTATSubmittedBodyContent(
 		bodyContent.RequesterEmail = requesterAccount.Email
 	}
 
-	if ctatRequest.CmmiGroup == models.CTATCMMIGroupOptionOther && ctatRequest.CmmiGroupOther.String != "" {
-		bodyContent.CMMIGroup = fmt.Sprintf("%s (%s)", bodyContent.CMMIGroup, ctatRequest.CmmiGroupOther.String)
+	if ctatRequest.ContractName != nil {
+		bodyContent.ContractName = *ctatRequest.ContractName
```
StevenWadeOddball, line 287: Non blocking, can string pointers be used here in the email content as well? I think go templating can handle a nil case conditionally

pkg/graph/resolvers/ctat_request.go
```diff
@@ -415,8 +405,8 @@ func sendCTATUpdateEmail(
 	// note: we won't send an email for a whitespace-only change, but we will still save the change in the DB
 	statusUpdated := originalRequest.Status != updatedRequest.Status
```
StevenWadeOddball, line 406: I think this works good to keep it in the loop of the app. Have you considered hooking into change history at all for this? This is probably simpler, but worth noting change history checks if a value has changed as well

pkg/graph/resolvers/ctat_request_test.go
```diff
@@ -483,17 +455,25 @@ func (suite *ResolverSuite) TestBuildCTATSubmittedBodyContentFormatsOtherValues(
 	)
```
StevenWadeOddball, line 455: same as db seed for this file, we should prefer new() since that is now available

StevenWadeOddball (approved):
Thanks Sam, couple more nits, but looks good! Feel free to merge and address later if you like, or decline making changes if you prefer.

pkg/graph/resolvers/ctat_request.go
samoddball: may look into in a follow up, but for now will keep this path. thanks!

pkg/graph/resolvers/ctat_request.go
samoddball: interesting! didn't realize that. may look into in a follow up!

PR #2335, opened 2026-06-11
[NOREF] waiver suggested triggers and some cleanup

Peyton232:
# NOREF

## Description

Adds DB-level triggers to auto-manage suggested_waiver rows when a waiver_assessment_survey is created or updated, replacing the equivalent Go functions. Also changes updateWaiver to an upsert keyed on (modelPlanID, commonWaiverID) so waiver rows no longer need to exist before being updated.

## How to test this change

1.  Create model plan 
2. Run Get All Common Waivers with modelPlanID -> all waivers should have isSuggested: true
3. Run Get Waiver Assessment Survey -> confirm survey exists with waivers array
4. Run Update Waiver with modelPlanID + a commonWaiverID from step 2 -> confirm waiver upserted with willUseWaiver
5. Run Update Waiver Assessment Survey (change any boolean field) -> re-run Get All Common Waivers -> suggested waivers should still reflect correctly

test for the trigger specifically: go test ./pkg/graph/resolvers/... -run TestResolverSuite/TestWaiverSuggestionTrigger -v

## PR Author Checklist

## PR Reviewer Guidelines


pkg/graph/resolvers/waiver_suggestion_trigger_test.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/samber/lo"
+
+	"github.com/cms-enterprise/mint-app/pkg/email"
+	"github.com/cms-enterprise/mint-app/pkg/models"
+)
+
+// medicarePaymentWaiver1ID is the stable ID for "Medicare Payment Waiver 1", seeded by V269.
+var medicarePaymentWaiver1ID = uuid.MustParse("9f955945-7afd-481f-8558-e7e0fd465463")
+
+func (suite *ResolverSuite) TestWaiverSuggestionTrigger() {
```
StevenWadeOddball, line 14: Can you revisit the in-line SQL here? We should prefer native app functionality whenever possible, and only resort to in-line SQL as as last resort when no other option is feasible.

pkg/graph/resolvers/waiver_assessment_survey.go
```diff
@@ -202,24 +88,60 @@ func WaiverGetByID(logger *zap.Logger, store *storage.Store, id uuid.UUID) (*mod
 	return storage.WaiverGetByID(store, logger, id)
 }
 
-// WaiverUpdate applies changes to a waiver row and persists them
-func WaiverUpdate(
+// UpdateSelectedWaivers upserts a waiver row for each selection in a single transaction.
+// Creates the row if it does not exist, otherwise updates will_use_waiver and not_using_reason.
+func UpdateSelectedWaivers(
```
StevenWadeOddball: This should likely be in a file specific to waivers, not in the assessment survey file.

pkg/graph/resolvers/waiver_assessment_survey.go
```diff
@@ -202,24 +88,60 @@ func WaiverGetByID(logger *zap.Logger, store *storage.Store, id uuid.UUID) (*mod
...
+	modelPlanID uuid.UUID,
+	selections []*models.WaiverSelectionInput,
 	principal authentication.Principal,
 	store *storage.Store,
-) (*models.Waiver, error) {
-	existing, err := storage.WaiverGetByID(store, logger, id)
+) ([]*models.Waiver, error) {
+	actor := principal.Account().ID
+
+	tx, err := store.Beginx()
```
StevenWadeOddball: We should use the WithTransaction helpers whenever possible.

pkg/graph/resolvers/waiver_assessment_survey.go
```diff
@@ -202,24 +88,60 @@ func WaiverGetByID(logger *zap.Logger, store *storage.Store, id uuid.UUID) (*mod
...
+
+	if err := tx.Commit(); err != nil {
+		return nil, err
+	}
+	return results, nil
+}
+
+// WaiverUpdate upserts a waiver row for the given model plan and common waiver,
+// applying the provided changes. Creates the row if it does not yet exist.
+func WaiverUpdate(
```
StevenWadeOddball: Maybe this should say Upsert instead of update to be clear about what is happening?

pkg/graph/resolvers/waiver_assessment_survey.go
```diff
@@ -202,24 +88,60 @@ func WaiverGetByID(logger *zap.Logger, store *storage.Store, id uuid.UUID) (*mod
...
+	actor := principal.Account().ID
+
+	tx, err := store.Beginx()
 	if err != nil {
 		return nil, err
 	}
+	defer tx.Rollback() //nolint:errcheck
+
+	results := make([]*models.Waiver, 0, len(selections))
+	for _, sel := range selections {
```
StevenWadeOddball: Have you considered making one SQL statement to handle this instead of relying on a for loop? we do a collection insert / update for tags that you could reference.

pkg/graph/resolvers/waiver_suggestion_trigger_test.go
StevenWadeOddball: And I think it's totally fair to put a todo on this and note that it is something to revist when we have the chosen list of waivers to mention

pkg/models/waiver.go
```diff
@@ -12,6 +12,14 @@ type Waiver struct {
 	NotUsingReason *string   `json:"notUsingReason" db:"not_using_reason"`
 }
 
+// WaiverSelectionInput is the input for a single waiver decision in a bulk update.
+// It is autobiound by gqlgen via the pkg/models autobind path.
+type WaiverSelectionInput struct {
```
StevenWadeOddball, line 17: Perhaps worthwhie to include db struct tags for mapping to sql statements?

pkg/storage/suggested_waiverStore.go
```diff
@@ -43,3 +43,15 @@ func SuggestedWaiverDeleteAllByModelPlanID(np sqlutils.NamedPreparer, _ *zap.Log
 
 	return sqlutils.SelectProcedure[models.SuggestedWaiver](np, sqlqueries.SuggestedWaiver.DeleteByModelPlanID, args)
 }
+
+// SuggestedWaiverInsertForModelPlan inserts a suggested_waiver row for every common_waiver
+// that should be suggested based on the current survey answers (DB-level CASE logic).
+// Skips rows that already exist (ON CONFLICT DO NOTHING).
+func SuggestedWaiverInsertForModelPlan(np sqlutils.NamedPreparer, _ *zap.Logger, modelPlanID uuid.UUID, createdBy uuid.UUID) ([]*models.SuggestedWaiver, error) {
```
StevenWadeOddball: At this point suggested waivers are handled with Database triggers, should this and the related queries and functionalites be removed?

pkg/sqlqueries/waiver.go
```diff
@@ -28,4 +33,5 @@ var Waiver = waiverScripts{
 	GetByID:                waiverGetByIDSQL,
 	GetByModelPlanIDLoader: waiverGetByModelPlanIDLoaderSQL,
 	Update:                 waiverUpdateSQL,
```
StevenWadeOddball: Should revisit and remove fields we wont use (like update if we only do an upsert)

query_examples/MINT/Task List Sections/Get Common Waiver.bru
```diff
@@ -39,3 +39,11 @@ body:graphql:vars {
     "id": "9f955945-7afd-481f-8558-e7e0fd465463"
   }
 }
+
+tests {
+  let responseData = res.getBody().data
```
StevenWadeOddball: NIT: Does it make sense to set this here? Perhaps set in the get all, and just set the first waiver id?

migrations/V272__Backfill_Waiver_Assessment_Survey_And_Waivers.sql
```diff
@@ -53,3 +53,83 @@ WHERE NOT EXISTS (
...
+ALTER TABLE suggested_waiver
+ENABLE TRIGGER audit_trigger;
+
+-- Part 3: DB triggers to manage suggested_waiver automatically going forward.
+-- On INSERT into waiver_assessment_survey (new model plan): seed all common waivers as
+-- suggested because no survey answers exist yet (all fields NULL).
+-- On UPDATE: delete existing suggestions and re-insert based on current answers.
+-- survey_question_field is NULL on all common_waiver rows until real mappings arrive via
+-- a future migration, so the ELSE TRUE branch suggests every waiver unconditionally.
+CREATE OR REPLACE FUNCTION MANAGE_SUGGESTED_WAIVERS()
```
StevenWadeOddball: I think this can be made more efficient. I would recommend putting the trigger in it's own migration for clarity. I suggest looking at v196 and v195 for how we do this for suggested milestones using a merge statement. There are change history considerations to keep in mind here, so we shouldn't delete and re-insert them all, just ones that should be deleted, and insert others.

StevenWadeOddball (changes requested):
Thanks Peyton, looking good so far, I have some suggestions, please let me know if you have any questions about this.

pkg/graph/resolvers/waiver_suggestion_trigger_test.go
Peyton232: I'll just leave a TODO for now

migrations/V273__Suggested_Waiver_Trigger.sql
```diff
@@ -0,0 +1,56 @@
...
+DECLARE
+    actor UUID;
+BEGIN
+    actor = COALESCE(NEW.modified_by, NEW.created_by);
+
+    MERGE INTO suggested_waiver AS target
+    USING (
+        SELECT
+            cw.id AS common_waiver_id,
+            CASE
```
StevenWadeOddball, line 57: Can you look into generically handling this case? h_store could be a good option so you don't need to manually handle each field.

StevenWadeOddball (commented):
Thanks Peyton,

Changes are looking good! I did not do functional testing on this. Only one more suggestion regarding handling fields. It would be good to not hard code the fields into the trigger if it can be avoided. I suggest looking into an h_store to get around it. This sort of logic can also be useful to make sure you are only changing fields as they make sense to be updated, so change history is not triggered unless it's needed

samoddball (approved):
functionally looks good!

migrations/V273__Suggested_Waiver_Trigger.sql
StevenWadeOddball, line 1: Thanks @Peyton232 , can you leave a todo to revisit this when we get the final configuration? Depending on the config, we might want to change this up a bit, but it's looking good. We will also want to test this more thoroughly when each common waiver is defined by field. I did a little validating by updating the config myself for now.

Checks.
1. Are waivers defined 1:1 by survey question? Or is the suggestion based on multiple fields? (Refactor if so)
2. Is the logic for checking the waiver suggestion needed elsewhere? Consider separating the responsibility to a function for a view of the data, and a trigger that uses that data to do mutations
3. Would it make more sense to separate the insert and update logics?  The trigger might be simpler without needing to handle the separate insert case. (Maybe a separate trigger, or just rely on app code for that?)

StevenWadeOddball: Also it's worth noting, this currently assumes the only value that makes a waiver suggested is `true`. If that changes, we will need an extra column to determine if suggested or not

StevenWadeOddball (approved):
Thanks @Peyton232 ,

I think this is looking good for now and can be merged. Can you add some notes to the trigger migration please? I left some suggestions for when the config is given to us. Implementation might need to change depending on the final requirements, but this should be good for now!

PR #2388, opened 2026-07-09
[MINT-3780] custom date emails

samoddball:
# MINT-3780

## Description

- adds email for custom date creation
- updates existing plan timeline email to include "Additional model dates" section (and to conditionally render the sections only if they are needed)

## How to test this change

1.  ensure all tests pass
2. run `scripts/dev email` and ensure the new emails appear in mailcatcher and appear accurate. new emails will have subjects: `New date added to Test Model Plan` and `Dates updated for Test Model Plan`

## PR Author Checklist

## PR Reviewer Guidelines


pkg/email/templates/model_plan_date_changed_body.html
```diff
@@ -17,32 +10,80 @@ <h3>Anticipated high level timeline</h3>
...
+              {{template "date" .OldDate}}
+            {{end}}
+        {{end}}
+      <br/>
+    </p>
+  {{end}}
+{{end}}
+
+<!-- only render section if there are custom timeline date changes -->
+{{if .CustomTimelineDateChanges}}
```
StevenWadeOddball, line 51: @samoddball, am I remembering right that these will never fire from the application? Thinking about it, I guess they would update from the main page. Is that correct?

pkg/email/templates/model_plan_date_changed_body.html
```diff
@@ -58,4 +99,4 @@ <h3>Anticipated high level timeline</h3>
   </p>
```
StevenWadeOddball, line 99: @samoddball, to remember, is this for the difference between the email going to an admin and it going to somesone subscribed to the email? I notice the footer isn't there on the email script.

pkg/graph/resolvers/plan_timeline_helper.go
```diff
@@ -67,6 +68,121 @@ func copyPlanTimelineTime(t *time.Time) *time.Time {
 	return nil
 }
 
+func getCustomTimelineDateUpdateIDs(customTimelineUpdates []*model.CustomTimelineDateUpdateDatesInput) ([]uuid.UUID, error) {
```
StevenWadeOddball: NIT, I think this is fine, but note, could rely on samber.lo library for filtering etc

pkg/graph/resolvers/custom_timeline_date.go
```diff
@@ -45,7 +50,156 @@ func CustomTimelineDateCreate(
...
+		principal,
+		emailService,
+		addressBook,
+		createdCustomDate,
+	)
+
+	return createdCustomDate, nil
+}
+
+func processCustomTimelineDateCreatedEmails(
```
StevenWadeOddball, line 71: Not critical, but it would be ideal to use the logging interfaces when possible instead of the zap logger directly, especially for utility functions. 
EG `logger logging.ChainableLogger[T]`

pkg/graph/resolvers/custom_timeline_date.go
```diff
@@ -45,7 +50,156 @@ func CustomTimelineDateCreate(
...
+	customTimelineDate *models.CustomTimelineDate,
+) {
+	if emailService == nil || customTimelineDate == nil {
+		return
+	}
+
+	principalAccount := principal.Account()
+	if principalAccount == nil {
+		logger.Error("failed to send custom timeline date created email because principal account is nil",
+			zap.String("customTimelineDateID", customTimelineDate.ID.String()),
```
StevenWadeOddball, line 87: not critical here, but you could consider decorating the logger higher up since it is being decorated the same place in both error statements (though if you don't use it except on error, this implementation might be more efficient)

pkg/graph/resolvers/custom_timeline_date.go
```diff
@@ -45,7 +50,156 @@ func CustomTimelineDateCreate(
...
+	}
+
+	emailRecipientUserAccounts, _ := models.FilterNotificationPreferences(recipientUserAccounts)
+	recipientEmails := make([]string, 0, len(emailRecipientUserAccounts))
+	for _, recipient := range emailRecipientUserAccounts {
+		recipientEmails = append(recipientEmails, recipient.Email)
+	}
+
+	// send email notifications to appropriate users
+	if len(recipientEmails) > 0 {
```
StevenWadeOddball, line 172: This deviates from the other timeline date changed email, note the other email puts the  async part on the email send, instead of the top level. It would be nice to reduce differences from the two functions and consolidate if possible. Fine to keep approaches separate if that fits the use case the best.

pkg/graph/resolvers/plan_timeline_helper.go
```diff
@@ -67,6 +68,121 @@ func copyPlanTimelineTime(t *time.Time) *time.Time {
...
+		if customTimelineDateChange.IsChanged {
+			customTimelineDateChanges = append(customTimelineDateChanges, customTimelineDateChange)
+		}
+	}
+
+	return customTimelineDateChanges, nil
+}
+
+func customTimelineDatesByID(customTimelineDates []*models.CustomTimelineDate) (map[uuid.UUID]*models.CustomTimelineDate, error) {
+	customTimelineDatesByID := make(map[uuid.UUID]*models.CustomTimelineDate, len(customTimelineDates))
```
StevenWadeOddball, line 138: NIT: I think it's preferrable to use the samber.lo library like we do throughout the app for consistency, and ease of reading. eg, lo.associate(customTimelineDates....)

pkg/graph/resolvers/plan_timeline.go
```diff
@@ -129,14 +136,27 @@ func UpdatePlanTimeline(
 
 		// update custom dates separately
 		if len(customTimelineUpdates) > 0 {
-			_, err := storage.CustomTimelineDateUpdateDatesByIDs(tx, principal.Account().ID, customTimelineUpdates)
+			existingCustomTimelineDates, err := storage.CustomTimelineDateGetByIDLoader(tx, customTimelineUpdateIDs)
```
StevenWadeOddball: This should use the dataloader with a loadManyCall. It's even though using a TX, this will still be fine becuase the dates have not been updated in the scope of the transaction. I would suggest that dataloader store methods should not be used outside of the dataloader call themselves

StevenWadeOddball (commented):
Thanks Sam,

This is looking great functionally. Mostly feedback about some approach, but mostly NITS.

Also, worth calling out there is in app notification for this. I don't see it called out in FIGMA, but I'll raise it to Jeff and Zoe to verify

pkg/email/templates/model_plan_date_changed_body.html
samoddball: yep! that was the key that i was missing - they can update in bulk from the main timeline page (along with the standard dates)

pkg/graph/resolvers/plan_timeline_helper.go
samoddball: cleaned this up! did not use `lo` cause we are mapping, filtering, and Uniq'ing, so i kept it in one loop but removed the unneeded error return

pkg/graph/resolvers/custom_timeline_date.go
samoddball: only using when needed - thanks!

pkg/graph/resolvers/plan_timeline_helper.go
samoddball: cleaned this up! skipped `lo` here, we need a small bit of loop control, but removed unneeded err return, thanks!

pkg/graph/resolvers/custom_timeline_date.go
samoddball: fixed! thanks

pkg/email/templates/model_plan_date_changed_body.html
samoddball: this email is the pre-existing timeline date change email with the new custom date added as an optional section (thus making the initial timeline section optional too, should there only be custom dates changed)

not sure what exactly the diff is in the final line in the file!

samoddball: it might just be a newline

pkg/graph/resolvers/plan_timeline.go
```diff
@@ -135,25 +130,28 @@ func UpdatePlanTimeline(
...
+			if errs != nil {
+				if len(errs) > 0 {
+					return nil, errs[0]
+				}
+
+				return nil, errors.New("problem getting existing custom timeline dates when updating plan timeline")
 			}
 
-			updatedCustomTimelineDates, err := storage.CustomTimelineDateUpdateDatesByIDs(tx, principal.Account().ID, customTimelineUpdates)
+			updatedCustomTimelineDates, err := storage.CustomTimelineDateUpdateDatesByIDsAndDates(tx, principal.Account().ID, dedupedCustomTimelineUpdates)
```
StevenWadeOddball: I think this store function got renamed incorrectly, it looks like the method is only updating by ID

samoddball: oops yep you are right! thanks - i thought i named it incorrectly in the first place and fixed it cause i thought it was a GET 😂 fixed! thanks

StevenWadeOddball (approved):

PR #2389, opened 2026-07-09
[MINT-3761] Spike Okta redirect login behind feature flag

garyjzhao:
# MINT-3761

## Description
Spikes migrating MINT from the embedded Okta Sign-In Widget to CMS Enterprise Login Page (ELP) / redirect-based Okta login.

This is a **frontend-only** change. Authorization server, claims (including `mint-groups`), audience (`aud=mint`), and redirect URIs are unchanged. Behavior is gated by `VITE_OKTA_REDIRECT_LOGIN_ENABLED` (default **`false`**), so deployed envs keep the widget until the flag is explicitly enabled.

### Why an env flag instead of LaunchDarkly
MINT’s LaunchDarkly client only initializes **after** the user is authenticated (`FlagsWrapper` waits on `authState.isAuthenticated`, then loads a signed user key via GraphQL). Sign In / `/signin` / `signInWithRedirect` all run **before** login, so an LD flag would not be available in time without a separate anonymous LD client (not used today). `VITE_OKTA_REDIRECT_LOGIN_ENABLED` follows the same pre-auth pattern as `VITE_LOCAL_AUTH_ENABLED` and can gate the whole environment at build time. Once redirect login is permanent, this flag goes away (`TODO(MINT-3761)`); it is not intended to move to LD.

### When the flag is on
- Landing / Header **Sign In** calls `signInWithRedirect()` directly (no intermediate widget page)
- `/signin` becomes a brief spinner bounce (still used by `ProtectedRoute` and deep links)
- Post-callback restore always lands on `/pre-decisional-notice` with `nextState` for the intended return path
- Local auth remains available via `/signin?local=true` and a dual-gated **Local Auth** header button (`isOktaRedirectLoginEnabled() && isLocalAuthEnabled()`)

### Always (flag on or off)
- OktaAuth requests scopes `openid`, `profile`, `email` (matches prior widget request; Auth JS default was narrower)
- Failed Okta callbacks show a user-facing error inside MINT header/footer (`OktaLoginCallback`), instead of a blank `/implicit/callback` page
- `UserInfoWrapper` no longer returns `null` on the callback route (that was blanking the page before `LoginCallback` could mount)

### Explicitly deferred
- Embedded widget package/component kept as rollback; cleanup sites marked `TODO(MINT-3761)`
- Okta Cypress e2e (`cy.login()` / widget DOM) is **skipped** until rewritten for hosted ELP
- *to discuss*: to figure out how to toggle `VITE_OKTA_REDIRECT_LOGIN_ENABLED` in the test/dev/impl environments. Cannot use LD as mentioned above. 

### Commits
1. Flag plumbing
2. Scopes + gated NDA/`nextState` restore
3. Redirect Sign In UI
4. Callback error UX
5. Cypress local login + skip Okta e2e
6. Widget cleanup TODOs

## How to test this change
1. **Flag off (default / rollback)**  
   Ensure `VITE_OKTA_REDIRECT_LOGIN_ENABLED=false` (committed `.envrc` default). Restart Vite. Visit `/signin` and confirm the embedded Sign-In Widget still works.
2. **Flag on (redirect path)**  
   In `.envrc.local` (preferred) or temporarily in env: `export VITE_OKTA_REDIRECT_LOGIN_ENABLED=true`, then restart Vite.
3. From `http://localhost:3005/`, click **Sign In** — should go straight to Okta/IDM (no intermediate Sign In button page).
4. Complete login → `/implicit/callback` → `/pre-decisional-notice` → continue to home.
5. Log out. Visit a protected URL while logged out (e.g. http://localhost:3005/models/00000000-0000-0000-0000-000000000005/read-view/participants-and-providers). After login + NDA, confirm you return to that URL (`nextState`).
6. With local auth enabled (`VITE_LOCAL_AUTH_ENABLED=true`), confirm **Local Auth** appears next to Sign In and opens DevLogin via `/signin?local=true`.
7. **Callback error UX:** open  
   `http://localhost:3005/implicit/callback?error=access_denied&error_description=test`  
   Confirm an error alert appears with header/footer (not a blank page), plus Return home / Try signing in again.
8. Optional: run Cypress local-auth specs; confirm the Okta “logs in with okta” case is skipped.

To enable in a deployed env later: set GitHub Actions environment variable `VITE_OKTA_REDIRECT_LOGIN_ENABLED=true` for that environment (see `build_frontend_assets.yml` / `build_static.sh`). Leave unset/false until ready.

## PR Author Checklist

## PR Reviewer Guidelines


StevenWadeOddball (commented):
Thanks @garyjzhao !

I did some functional testing and it is looking good. http://localhost:3005/implicit/callback?error=access_denied&error_description=test did work as expected when not logged in, but it shows a blank page when logged in. That makes sense to me.

@yiranli624, can you take a look when you have some time!

yiranli624:
(The approach looks good to me! not sure if this is intentional or in progress, but with local auth, i could not pass the pre-decision page.)
Edit: Not able to recreate this bug, very possible just a glitch between env variable changes.

StevenWadeOddball:
> The approach looks good to me! not sure if this is intentional or in progress, but with local auth, i could not pass the pre-decision page

Thanks @yiranli624 ! Hmm, that should be working, it might be a bug

cypress/e2e/login.spec.js
```diff
@@ -1,7 +1,9 @@
 const maxAttempts = 3;
 
 describe('Logging in', () => {
-  it(
+  // TODO(MINT-3761): rewrite cy.login() for hosted Okta/ELP redirect login (widget DOM selectors
+  // break when VITE_OKTA_REDIRECT_LOGIN_ENABLED=true). See cypress/support/login.js.
+  it.skip(
```
yiranli624: is this going to be included in this pr?

yiranli624 (approved):
lgtm!

garyjzhao:
applying do not merge label for now. need to test in dev deployed environment before going through the pipeline

PR #2396, opened 2026-07-13
[NOREF] GQL types organization

StevenWadeOddball:
## Description

This PR seeks to group GQL types into related subfolders in order to make them more manageable. This should be strictly organizational with no functional impace.
> [!TIP]
> The files changed view might be more difficult to view the changes, than looking at the file structure in your favorite IDE

> [!IMPORTANT]
> We might want to further nest the separation for the core model collaboration area, but I left it like this for now
## How to test this change

1.  Smoke test the application is not impacted functionally
2. Review the organization and and see if you have any suggestions

## PR Author Checklist

## PR Reviewer Guidelines


yiranli624 (changes requested):
Data exchange related should be in questionnaires dir.
Also, should we move all plan_ to a model_plan dir? model_collab seems covering too much

StevenWadeOddball:
Thanks @yiranli624!

Good call. Yes, model collaboration is large. I do wonder about moving some of the layer deeper, so we have a model_plan folder, and that has sections folder, discusions folder, questionnaires folder etc underneath. I do want to avoid having too many layers, but maybe adding one more could make sense? That way we are also separating anything that isn't related to the model plan as well? What do you think?

yiranli624:
> Good call. Yes, model collaboration is large. I do wonder about moving some of the layer deeper, so we have a model_plan folder, and that has sections folder, discusions folder, questionnaires folder etc underneath. I do want to avoid having too many layers, but maybe adding one more could make sense? That way we are also separating anything that isn't related to the model plan as well? What do you think?

I'm with you on avoiding too many layers! maybe not moving layer deeper, but just one flat layer. IE model_plan is not inside of collaboration

yiranli624 (commented):
LGTM!

samoddball (approved):
looks good to me!

Peyton232 (approved):
This looks good to me, folder layout makes sense and should improve developer experience

PR #2399, opened 2026-07-13
[MINT-3799] emails to assigned member

samoddball:
# MINT-3799

## Description

- adds email send to assigned admin (if there is one) when there is a CTAT request update (email mimics that of the CTAT team's email)

## How to test this change

1.  ensure build/lint/tests pass
2. run `scripts/dev email` - confirm an email goes out to `audrey.abrams@mint.dev.cms.gov` with subject `Progress update on a contract assistance ticket (CTAT-021)`

## PR Author Checklist

## PR Reviewer Guidelines


StevenWadeOddball (commented):
Thanks Sam.

Did you consider sending the email once? The emailservice has the ability to cc email addresses etc. Might be better to try and send them all at once instead of two code blocks.

samoddball:
> Thanks Sam.
> 
> Did you consider sending the email once? The emailservice has the ability to cc email addresses etc. Might be better to try and send them all at once instead of two code blocks.

yes! per product team, two separate emails was preferred! had the same idea 😄 thanks Steven!

StevenWadeOddball (approved):
Thanks @samoddball , approving to unblock. I think this would probably be better if the email was fired at the same time in code even if it's two separate emails functionally. Not a huge deal though, but it would avoid some repetition

PR #2405, opened 2026-07-15
[NOREF] debounce echimp err

samoddball:
# NOREF

## Description

- coalesces concurrent echimp cache refresh failures and adds a cooldown period to prevent slews of logs from coming in all at once

## How to test this change

1.  ensure tests pass, and let's keep an eye out for the live echimp creds refresh (happens every few weeks)

## PR Author Checklist

## PR Reviewer Guidelines


pkg/echimpcache/echimp_cache.go
```diff
@@ -17,25 +18,33 @@ import (
 )
 
 var CRAndTDLCache *crAndTDLCache
+var crAndTDLCacheMu sync.Mutex
```
StevenWadeOddball: can you add a var comment to note why we need a global mutex as well as the mutex on the type please?

pkg/echimpcache/echimp_cache.go
```diff
@@ -17,25 +18,33 @@ import (
...
 // If the time since it was last updated has elapsed, it will fetch the data again
 func GetECHIMPCrAndTDLCache(ctx context.Context, client *s3.S3Client, viperConfig *viper.Viper, logger *zap.Logger) (*crAndTDLCache, error) {
-	if CRAndTDLCache == nil {
-		CRAndTDLCache = &crAndTDLCache{}
-	}
+	cache := getOrCreateECHIMPCache()
 	logger = logger.With(logfields.EchimpCacheAppSection)
-	if CRAndTDLCache.IsOld(viperConfig) {
-		err := CRAndTDLCache.refreshCache(ctx, client, viperConfig, logger)
+	if cache.IsOld(viperConfig) {
```
StevenWadeOddball: is this redundant now that it is in the ensurefresh function?

pkg/echimpcache/echimp_cache.go
```diff
@@ -64,6 +89,93 @@ func (c *crAndTDLCache) IsOld(viperConfig *viper.Viper) bool {
...
+	c.refreshInFlight = false
+	close(done)
+
+	return refreshAttemptResult{
+		err:         refreshErr,
+		hasSnapshot: hasSnapshot,
+	}
+}
+
 func (c *crAndTDLCache) refreshCache(ctx context.Context, client *s3.S3Client, viperConfig *viper.Viper, logger *zap.Logger) error {
```
StevenWadeOddball, line 219: should this method be updated to use the mutex now?

pkg/echimpcache/echimp_cache.go
```diff
@@ -64,6 +89,93 @@ func (c *crAndTDLCache) IsOld(viperConfig *viper.Viper) bool {
 
 }
 
+func (c *crAndTDLCache) ensureFresh(viperConfig *viper.Viper, logger *zap.Logger, refresh func() error) error {
```
StevenWadeOddball: Can you add some function comments to this and other helpers please? It will be helpful when we need to look at the code in the future

StevenWadeOddball (commented):
Thanks Sam,

This is looking good. Do you have any suggestions for local testing? Unit tests look good, we might need to rely on those for the debounce. I gave a couple of feedback points, but I think this will be great for silencing the repeated slack alerts!

pkg/echimpcache/echimp_cache.go
```diff
@@ -64,6 +89,93 @@ func (c *crAndTDLCache) IsOld(viperConfig *viper.Viper) bool {
...
+			return err
+		}
+
+		previousLastChecked := c.lastChecked
+		c.refreshInFlight = true
+		c.refreshDone = make(chan struct{})
+		c.mu.Unlock()
+
+		err := refresh()
+		result := c.completeRefreshAttempt(previousLastChecked, err)
```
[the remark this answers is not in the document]
samoddball: handled this

pkg/echimpcache/echimp_cache.go
samoddball: yes! good catch - thank you!

pkg/echimpcache/echimp_cache.go
samoddball: updated this to a sync.Once - i should have used that from the start! now only one mutex to manage in there, nice call out!

pkg/echimpcache/echimp_cache.go
samoddball: yes! updated. also removed any direct reads from other spots in the codebase - those are protected by the mutex as well. good shout!

StevenWadeOddball (commented):
Thanks @samoddball , I think copilots feedback about model plan makes sense. 

It's also worth considering, if you need to wrap the reading of the individual data now because of the mutex, it also suggests you might not even need to fetch the cache outside the package. You could just have the methods fetch the singleton and return the data. I'd also vote to leave it as is for now to avoid adding more scope to this

StevenWadeOddball (approved):

PR #2430, opened 2026-07-30
[MINT-3812] discussion topic infra

samoddball:
# MINT-3812

## Description

- adds enums to DB/gql/BE/UI
- adds `topic` column to DB (backfills with `OTHER`)
- adds `topic` to DB queries and Go types
- adds `topic` to bruno/UI queries and tests
- adds `topic` to change history and translation
- moves older query embeds from the `storage` package to the `sqlqueries` package (we had two spots for both `plan_discussion` and `audit_change` queries - both now live with their newer version in `sqlqueries`)

## How to test this change

1.  ensure migrations/build/lint/tests pass
2. for the `topic` infra, use bruno to create a new model plan, grab the new ID
3. use that ID in both the `PlanDiscussionCreate TAG CURRENT USER ONLY.bru` and `Create Plan Discussion.bru` queries and ensure you get success

## PR Author Checklist

## PR Reviewer Guidelines


cmd/test_email/main.go
```diff
@@ -148,6 +148,7 @@ func sendPlanDiscussionCreatedTestEmail(
 		uuid.Nil,
 		false,
 		uuid.Nil,
+		models.DiscussionTopicTypeOther,
```
samoddball, line 151: email is not yet wired up, this is just in anticipation of that

pkg/sqlqueries/SQL/audit_change/collection_by_id_and_table.sql
samoddball, line 1: the audit change folder was the final one that needed to move to the newer `sqlqueries` package. i was in here moving the outdated `plan_discussion` folder, so i moved this one too

golangci-lint-full
samoddball, line 1: don't know what this file was, but it shouldn't be included in the repo. looks like an accidental add a while back

samoddball:
e2e tests will fail as the UI won't have any topic dropdown, etc yet, but the BE will require it - it's ok for it to fail on this branch and up until the UI is done

pkg/storage/audit_changeStore.go
```diff
@@ -65,13 +56,13 @@ func (s *Store) AuditChangeCollectionByIDAndTableAndField(
...
+
 	orderClause := "" //default to ASCENDING
 	if sortDir == models.SortDesc {
 		orderClause = " ORDER BY 1 DESC"
 	}
 
-	orderedQuery = orderedQuery + orderClause
+	orderedQuery := sqlqueries.AuditChange.CollectionByIDAndTableAndField + orderClause
 
 	stmt, err := s.db.PrepareNamed(orderedQuery)
```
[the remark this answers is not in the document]
samoddball: this does not change any behavior, just uses a sql query that has been moved to the newer paradigm

src/features/ModelPlan/Discussions/QuestionAndReply.tsx
```diff
@@ -146,6 +146,7 @@ const QuestionAndReply = ({
       <Formik
         initialValues={{
           content: '',
+          topic: undefined,
           userRole: mostRecentUserRole || ('' as DiscussionUserRole),
           userRoleDescription: mostRecentUserRoleDescription || ''
         }}
```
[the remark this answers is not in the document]
samoddball: this will be required/set in a future UI branch - this is ok for now, we want the dropdown to default to the `- Select -` choice when the page loads

src/features/ModelPlan/Discussions/index.tsx
```diff
@@ -38,8 +39,10 @@ export type DiscussionsProps = {
 
 export type DiscussionFormPropTypes = Omit<
   PlanDiscussionCreateInput,
-  'modelPlanID'
->;
+  'modelPlanID' | 'topic'
+> & {
+  topic?: DiscussionTopicType;
+};
```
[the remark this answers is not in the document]
samoddball: this has been addressed

migrations/V269__Add_Discussion_Topic.sql
```diff
@@ -0,0 +1,20 @@
...
+    'WAIVER_ASSESSMENT_SURVEY',
+    '_4I_ACO_OS_QUESTIONNAIRE',
+    'MODEL_TO_OPERATIONS_MATRIX_MTO',
+    'DOCUMENTS',
+    'CONTRACTS',
+    'FFS_CRS_AND_TDLS',
+    'OTHER'
+);
+
+ALTER TABLE plan_discussion ADD COLUMN IF NOT EXISTS topic DISCUSSION_TOPIC_TYPE NOT NULL DEFAULT 'OTHER';
```
StevenWadeOddball: I wonder if it would be better to not set a default to OTHER. It's true we want historic data to go to OTHER, but we do want to force the user to select a choice otherwise.

migrations/V269__Add_Discussion_Topic.sql
```diff
@@ -0,0 +1,20 @@
...
+    'MODEL_PLAN_MODEL_BASICS',
+    'MODEL_PLAN_GENERAL_CHARACTERISTICS',
+    'MODEL_PLAN_PARTICIPANTS_AND_PROVIDERS',
+    'MODEL_PLAN_BENEFICIARIES',
+    'MODEL_PLAN_OPERATIONS_EVALUATION_AND_LEARNING',
+    'MODEL_PLAN_PAYMENT',
+    'MODEL_TIMELINE',
+    'DATA_EXCHANGE_APPROACH',
+    'WAIVER_ASSESSMENT_SURVEY',
+    '_4I_ACO_OS_QUESTIONNAIRE',
```
StevenWadeOddball:
```suggestion
'4I_ACO_OS_QUESTIONNAIRE'
```

pkg/graph/schema/types/plan_discussion.graphql
```diff
@@ -1,9 +1,29 @@
...
+  MODEL_PLAN_MODEL_BASICS
+  MODEL_PLAN_GENERAL_CHARACTERISTICS
+  MODEL_PLAN_PARTICIPANTS_AND_PROVIDERS
+  MODEL_PLAN_BENEFICIARIES
+  MODEL_PLAN_OPERATIONS_EVALUATION_AND_LEARNING
+  MODEL_PLAN_PAYMENT
+  MODEL_TIMELINE
+  DATA_EXCHANGE_APPROACH
+  WAIVER_ASSESSMENT_SURVEY
+  _4I_ACO_OS_QUESTIONNAIRE
```
StevenWadeOddball:
```suggestion
4I_ACO_OS_QUESTIONNAIRE
```

pkg/models/plan_discussion.go
```diff
@@ -4,10 +4,32 @@ import (
...
+	DiscussionTopicTypeModelPlanModelBasics                     DiscussionTopicType = "MODEL_PLAN_MODEL_BASICS"
+	DiscussionTopicTypeModelPlanGeneralCharacteristics          DiscussionTopicType = "MODEL_PLAN_GENERAL_CHARACTERISTICS"
+	DiscussionTopicTypeModelPlanParticipantsAndProviders        DiscussionTopicType = "MODEL_PLAN_PARTICIPANTS_AND_PROVIDERS"
+	DiscussionTopicTypeModelPlanBeneficiaries                   DiscussionTopicType = "MODEL_PLAN_BENEFICIARIES"
+	DiscussionTopicTypeModelPlanOperationsEvaluationAndLearning DiscussionTopicType = "MODEL_PLAN_OPERATIONS_EVALUATION_AND_LEARNING"
+	DiscussionTopicTypeModelPlanPayment                         DiscussionTopicType = "MODEL_PLAN_PAYMENT"
+	DiscussionTopicTypeModelTimeline                            DiscussionTopicType = "MODEL_TIMELINE"
+	DiscussionTopicTypeDataExchangeApproach                     DiscussionTopicType = "DATA_EXCHANGE_APPROACH"
+	DiscussionTopicTypeWaiverAssessmentSurvey                   DiscussionTopicType = "WAIVER_ASSESSMENT_SURVEY"
+	DiscussionTopicType4iAcoOsQuestionnaire                     DiscussionTopicType = "_4I_ACO_OS_QUESTIONNAIRE"
```
StevenWadeOddball:
```suggestion
DiscussionTopicType4iAcoOsQuestionnaire                     DiscussionTopicType = "4I_ACO_OS_QUESTIONNAIRE"
```

src/i18n/en-US/modelPlan/discussions.ts
```diff
@@ -6,6 +6,34 @@ import {
...
 
 export const discussions: TranslationDiscussions = {
+  topic: {
+    gqlField: 'topic',
+    goField: 'Topic',
+    dbField: 'topic',
+    label: 'Discussion topic',
+    dataType: TranslationDataType.ENUM,
+    formType: TranslationFormType.SELECT,
+    order: 1,
```
StevenWadeOddball: The paradigm for order is the first digit is the page, then there is a decimal number to show the relative order in that page. Note 1.01 already exists. 

This should also have a decimal representation

StevenWadeOddball (approved):
Thanks Sam,

This is looking good, there are just a couple of minor changes I'd suggest re default values, variable names, and translation settings. 

No need now that it's done and reviewed, but I'd also encourage this to be more targeted. 

Things like moving the queries, updating the FE etc are all great changes to consider in the scope of this feature, but a separate follow up PR to the feature branch would keep this focused.

pkg/graph/schema/types/plan_discussion.graphql
samoddball: the reason for the underscore is gql does not allow enums that begin with a number - the underscore is kept throughout for parity. thanks!

migrations/V269__Add_Discussion_Topic.sql
samoddball: updated! thanks

src/i18n/en-US/modelPlan/discussions.ts
samoddball: thanks Steven! i have bumped this to 1.001

StevenWadeOddball (dismissed):
Thanks Sam! That makes sense about it not starting with 4! Can you work to make an updated enum reference that matches the other references to the questionnaire in the app?

in models, we call it `IDDOCQuestionnaire`,

In common solutions we call 4I `INNOVATION`. 

I'd be inclined to call it `IDDOCQuestionnaire` in snake case form to match how we refer to it in the application. I'm concerned it might be confusing to have so many names to talk about the same concept

StevenWadeOddball (approved):
Thanks @samoddball ! This looks good!

